PR #6, opened 2015-05-05
Debian Importer

barnabycourt:


common/pulp_deb/common/constants.py
```diff
@@ -25,6 +25,7 @@
 DISTRIBUTOR_CONFIG_VALUE_PUBLISH_DIRECTORY = '/var/lib/pulp/published/deb'
 DISTRIBUTOR_CONFIG_FILE_PATH = 'server/plugins.conf.d/deb_distributor.json'
 
+UNIT_KEY_FIELDS = ["name", "version", "architecture", "filename"]
```
rbarlow: I noticed that you excluded the filename from unit key fields elsewhere - might want to drop that here as well. Also, it would be ideal to follow DRY and read this from the unit definition file instead of duplicating the information here as well.

rbarlow:
Unfortunately, I seem to have put all of my comments on a commit rather than on the pull request itself. Silly me. Most of my comments are here:

https://github.com/barnabycourt/pulp_deb/commit/7076e7e3482b655752c98e2f22f765f70060e55c

Watch out for the unit key comment, and for the potential issue with urljoin. Once you address those, LGTM.

PR #10, opened 2015-05-18
Add test requirements.

rbarlow:
Also, configure pulp-dev.py to install the packages.


rbarlow:
This one fails because python-debian is not being installed in our Jenkins environment.

beav:
LGTM

PR #15, opened 2015-12-03
Convert shebang to python2

ipanova:


seandst:
ok test

seandst:
woops, the failed was in travis, which appears to have an overzealous linting config

PR #16, opened 2016-02-23
 Block attempts to load server.conf when running tests

seandst:
https://pulp.plan.io/issues/413
re #413


rbarlow:
The test failures seem to be Jenkies installation problems and not problems introduced by this change.

PR #21, opened 2017-05-29
Add option to specify distribution suite

mdellweg:


mibanescu:
Thank you! Could you look into adding some tests? For the glaring bugs I don't mind adding tests and asking to re-submit, but for features I think I need help in that area.

Also, it doesn't look like the distributor will publish with the original suite, so I am a bit hesitant and would like to understand better if that's what you intended.

mdellweg:
I'm happy to add some tests. I will do the work on #23, which contains this as a first commit anyway.

PR #23, opened 2017-05-30
[WIP] Feature/select comparch

mdellweg:
This Patch adds the possibility to select several architectures and components from a debian repository.
All selected combinations will be synced and published.
When publishing, the architectures will be separate, but the components will be merged for now.


mibanescu:
Same comment as for #21 - I could use some help with tests, and I am concerned debpkgr is not smart enough to separate arches correctly at publish time.

mdellweg:
I'll try to give some pointers:
- Yes, the suite used to download is not used to publish again. By now, it will always be published as 'stable', which is not nice. I would suggest to just copy the name unless a corresponding parameter ('publish_suite' or similar) overwrites it. Maybe, this parameter accepts a list of names, so you can access the 'Release.gz' by either codename or suite (testing / stretch).
- For the components it is rather similar. I could think of a parameter stating, that all components should be merged (as it is now), otherwise they are mirrored.
- Concerning the architectures, debpkgr does not need to be smart there, as the patch already sorts by arch and merges the 'all'-packages to every other architecture.

mibanescu:
To set expectations properly:

We are in the process of trying to get pulp_deb released as part of pulp 2.14. I am also in the process of taking a couple of weeks of vacation, and that limits my ability to review your patches.

It would be great if you can join the pulp-dev mailing list (if you are on it already, you may have seen my email requesting it).

So I will need to decide whether we should pull these patches in before the freeze. I really wish I could get to review these today - it is so awesome to have help.

mdellweg:
I have subscribed now.
At the moment, my goal in this PR is to be able to do a sync of some debian archive including several releases by codename (properly linked by the corresponding suite), that can share common packages.
It will be configurable to include only a subset of components and architectures.
I will do my best to keep up on the tests (at the very least, keep the current ones passing).
Help and comments are appreciated.

mibanescu:
https://pulp.plan.io/issues/2764 lists my thoughts (or lack thereof) around this problem. I am not sure you saw it.

Essentially, I believe the sync codepath needs to save the original component of the unit, so that the distributor can do the right thing when it publishes the unit. I am not sure if that should be done as:
* additional metadata on the unit itself - it feels dirty, because you may have the same package in two different pulp repos (synced from two completely different upstream apt repos)
* an extra unit type the sync code will populate, that the distributor knows to look for
* additional configuration in the importer, that the syncer would populate. This seems like a strange side-effect and would prevent one from importing a deb unit in a different component than the default one, unless they go and edit the importer config, and that seems counter-intuitive to me.

So, high-level, if you are syncing comp1 and comp2, and: 
* deb package foo is both in comp1 and comp2, you'd have a dictionary-like unit that says "foo belongs to comp1 and comp2".
* deb package bar exists only in comp1, the dictionary-like unit would only say "bar belongs to comp1"

Adding extra unit types was not on my priority list :-) so I only created the ticket in the issue tracker.

mdellweg:
I'm just scratching the surface, but if you think an extra unit-type used like a join table model is the right way, then why hesitate. I just thought, a unit was always associated with a file to be published eventually.
I will look into it.

PR #24, opened 2017-06-19
Feature/sync release structure

mdellweg:
This branch is a rewrite from ATIX-AG:feature/select_comparch (https://github.com/pulp/pulp_deb/pull/23).
It follows the ideas by @mibanescu laid out in the discussion there and https://pulp.plan.io/issues/2764.
Sync options have been added to specify the releases to be synchronized and to filter packages by the components and architectures.


mibanescu:
Thank you for the PRs. I was on vacation till July 13th and couldn't look at the patches, and the first weeks back were too busy. I will try to read them this week, and will post back comments. It was not my intention to incorporate your work in what will get released with pulp 2.14, since they were riskier. I apologize for the lack of communication on my part.

mdellweg:
Looking forward to hearing from you.

plugins/pulp_deb/plugins/distributors/distributor.py
```diff
@@ -228,46 +233,95 @@ def _get_total(self):
...
-                                repo_name=self.get_repo().id,
-                                metadata=repometa,
-                                gpg_sign_options=sign_options)
-
-        arepo.create(filenames,
-                     component=self.Component,
-                     architecture=arch,
-                     with_symlinks=True)
+
+        for release_unit in self.parent.publish_releases.units:
```
mibanescu: I think we may have a backwards-incompatible problem here.

If I created a repo with a previous version of the plugin, there will be no release and component units present. Then self.parent.publish_components.units will be empty (and so will be self.parent.publish_releases.units). So the publish won't do anything.

I can see two possible solutions (although there may be more):
* Add a migration script that creates the missing release and component units
* When self.parent.publish_{components,releases}.units are empty, create a non-persisted placeholder unit that matches the old repos.

I would favor the latter. I imagine, for the simple deb repos that we were producing prior to this patch, there would be no need to specify a component-to-package relationship. I may have to revisit this comment once I read some more.

plugins/pulp_deb/plugins/distributors/distributor.py
```diff
@@ -228,46 +233,95 @@ def _get_total(self):
...
+            codename = release_unit.codename
+            rel_components = [comp for comp in comp_units
+                              if comp.id in release_unit.components]
+            architectures = set()
+
+            comp_arch_units = {}
+            for component_unit in rel_components:
+                # group units by architecture (all, amd64, armeb, ...)
+                arch_units = defaultdict(list)
+                for unit in [unit for unit in units if unit.id in component_unit.packages]:
```
mibanescu: This codepath will work fine if your major source of units is via repo syncing.
However, if you are uploading your own few packages into a repo with only one release and one component, I think maintaining a component-to-package list is cumbersome. I would like to suggest a "default component" where all packages not matching an existing component would go. The default component would have a null list of package names, and would be the placeholder in case no component units are present, that I mentioned above.

In other words, there are several usecases that I think should be treated somewhat independently (and optimized independently):

* create your own repo, populate it with a few packages. This is what we currently have, and is single-release single-component.
* sync a simple repo (you created yourself like in the previous use case, or single-release "stable", single-component "main"). There would be no need to create release and component units in this case. One can continue to upload the repo if they so choose.
* sync a complex repo (more than one release, or component, or with different names than the defaults mentioned above). Release and component units will need to be created in order to publish the repo correctly. Uploading your own units will mean either that there is a component with the name in its package list, or you'd also have to upload a new component in order to include the new package. The next sync will most likely blow away the component definition, so I don't like where this is going.

I would love to hear your opinion on this. I don't have the answers. I would like to keep the simple cases simple (the first two above), and clearly define what is supported and what is not for case number 3. Maybe we define complex syncing as a "full mirror" operation (i.e. you may not upload your own package since the next sync will overwrite that). There was a similar concept in pulp's parent/child node which is being deprecated - you could define a repo relationship as mirror or additive.

mdellweg: Thanks for sharing your opinion. You are right, syncing full existing repositories was my first goal here.
If i understand your concerns right, we can solve the problem by introducing a virtual `default` (and maybe an `all`) component, to include all otherwise non-published (respectively all) packages, being the non-persistent layer you mentioned. That way we should always have your "simple scenario" alongside the complex full release structure.
The "full mirror" sounds like the `--remove-missing` option (which honestly is not yet implemented).
I will continue by experimentation with the virtual components.

mdellweg:
I started with the easy task. I added a `default` release with component `all` to contain just every package of the repo. What do you think about the names? Despite being incompatible to prior versions, i dislike calling it stable/main (which would anyway be occupated most of the time).

plugins/pulp_deb/plugins/distributors/distributor.py
```diff
@@ -228,46 +233,123 @@ def _get_total(self):
...
+
+            for component in comp_arch_units:
+                for architecture, ca_units in comp_arch_units[component].iteritems():
+                    filenames = [unit.storage_path for unit in ca_units]
+                    arepo.create(filenames,
+                                 component=component,
+                                 architecture=architecture,
+                                 with_symlinks=True)
+
+        # create a special release with one component to include all packets
```
mibanescu: Don't you have to do the "default" component only if no other components are present? I think that's what confuses me about the patch.

plugins/pulp_deb/plugins/distributors/distributor.py
```diff
@@ -228,46 +233,123 @@ def _get_total(self):
...
+            arch_units[unit.architecture].append(unit)
+        # architecture 'all' is special; append it to all other architectures
+        all_units = arch_units.pop('all', [])
+        for arch in arch_units:
+            arch_units[arch].extend(all_units)
+            architectures.add(arch)
+
+        repo_meta = aptrepo.AptRepoMeta(
+            codename='default',
+            components=['all'],
```
mibanescu: In an effort to not have to write a data migration, I'd have to insist on keeping the existing naming convention (stable/main) even though I didn't come up with it.

plugins/pulp_deb/plugins/distributors/distributor.py
mdellweg: That would be the beauty of it. It does not hurt to be there.
However, if it must really be `stable/main` (see below) it is more complicated.

plugins/pulp_deb/plugins/distributors/distributor.py
mdellweg: Do you mean something like that?

mdellweg:
@mibanescu Any progress on this?

plugins/pulp_deb/plugins/distributors/distributor.py
mibanescu: Apologies, daily job gets in the way. I will look at it today/tomorrow.

plugins/pulp_deb/plugins/distributors/distributor.py
mibanescu: I believe this is really the part I am having problems understanding.

If you don't have any releases (and any components), then I would totally expect the existing behavior.
If you do have releases, I would NOT expect an additional "default" release to be present, unless that's expected for Debian repositories (I imagine not, but I am very inexperienced).

That's why I was asking why not return if len(release_units) == 0 (below)

In my previous comment I said "component" when I should have said "release" - I apologize.

As I am reading the code and running the tests, I experimented with that idea. Unfortunately, the tests continued to pass, which means we're missing some coverage for this case.

Please consider my questions while I get myself familiar with the code (again :-)

mdellweg: Well it would not be 'expected'. And i am happy to change that.
On the other hand, while discussing this here, i thought it might be beneficial to always have all content available in a consistent (across repos) location (hence: default/all), while this does not affect the already available locations.
What would you think about doing the stable/main thing if there is nothing, and having the default/all combination completely apart from that depending on a boolean configuration option.

As for the tests, i think it would be better to design them once we know, where the answer to this is going.

mibanescu: Putting my "backwards compatible hat" on (if there is such a thing).
The first thing that I had to implement was ability to publish my own content and ability to sync content published by me.
If an "old" pulp server tried to sync content from a "new" pulp server (i.e. prior to your changes vs. with your changes), I would expect the above to still hold true, at least for a while.
If we had an additional default/all release, it would break this behavior, because the old pulp server would be confused by the extra release.
So, it *seems* like the safe thing to do is: if you find value in it, do continue to have a default/all, with a boolean config option, and be off by default.
Agreed on the tests, I happened to notice because I was trying to understand the code and I was playing with the tests at the same time.

mdellweg: updated

mdellweg: @mibanescu Is your "backwards compatible hat" turning green, or is it still dark red?

mibanescu: :+1: much better! I have a stylistic comment, if there is nothing after an if statement I prefer to reverse the check and return early, to avoid an extra level of indentation. I will put a comment there.

plugins/pulp_deb/plugins/distributors/distributor.py
```diff
@@ -322,42 +322,48 @@ def process_main(self, item=None):
...
-                         with_symlinks=True)
+            generic_release_names.append(('stable', 'main'))
+
+        # create a special release with one component to include all packets
+        # group units by architecture (all, amd64, armeb, ...)
+        if self.get_config().get(constants.PUBLISH_DEFAULT_RELEASE_KEYWORD, False):
+            generic_release_names.append(('default', 'all'))
+
+        # only do this, iff necessary
+        if len(generic_release_names) > 0:
```
mibanescu: I personally would do:
if not generic_release_names:
    return
... everything else, one level of indentation less

mdellweg: Actually, i like it better without the shortcut, because if some time in the future, someone thinks, there needs to be one last thing we do in this routine, he might overlook this if-statement. I'm sure he will notice the indentation.
Also the block is intentionally meant to be conditional, so why try to hide it?

mdellweg:
@mibanescu now with tests...

jlsherrill:
@mibanescu any thoughts on when we might be able to get this in?

mibanescu:
Let me try to get to it this week. I was't sure if you're done with unit tests, and I would like to have some time to test the functionality we (my employer) are using; however, since we are unlikely to move to it any time soon (since we are in a lock-down period), I wouldn't want to block other people from using the plugin either. This open-source thing is hard to balance :-)

PR #29, opened 2017-08-18
Fix http publishing

mdellweg:
Due to a misconfiguration (config now stolen from pulp_rpm) the options serve-http(s) were not really useful.


mdellweg:
Well, i experienced the problem and tested my solution in the pulp_devel vagrant environment.
I honestly do not know, how it is installed in production environment. Let's hope for the best and thanks for your confidence!
BTW there is another copy of the http config at `etc/httpd/conf.d/pulp_deb.conf` is this on purpose or should it be removed?

mibanescu:
Sorry for not responding earlier - I should remove the other one, since it's not in use.

PR #32, opened 2017-10-17
Speedup distributor

mdellweg:


mdellweg:
This should reduce some algorithms from `O(n^2)` to `O(n log n)`.
Tell me, if you spot more.

PR #37, opened 2018-02-05
Add cryptographic verification to Release

mdellweg:


plugins/pulp_deb/plugins/importers/sync.py
```diff
@@ -129,6 +135,20 @@ def process_main(self, item=None):
         architectures = self.parent.architectures
         dl_reqs = []
         for release in releases:
+            if not os.path.isfile(self.parent.release_files[release]):
```
mibanescu: Can you separate this block into its own function, so it's easier to test?

pulp-dev.py
```diff
@@ -40,7 +40,7 @@
      '/etc/pulp/server/plugins.conf.d/deb_importer.json'),
 )
 
-PIP_INSTALLABLE = set(['debpkgr'])
+PIP_INSTALLABLE = set(['debpkgr', 'gnupg'])
```
mibanescu: I think the spec file needs to also be edited to pull in python2-gnupg (an EPEL package, as far as I can tell). I am checking with pulp's release engineering team to see if they have to build it into the pulp repos.

plugins/pulp_deb/plugins/importers/sync.py
mdellweg: you know, i am a bit lazy on the tests until i know, how the structure of the feature is going to settle. Tests will follow.

pulp-dev.py
mdellweg: Will do.

mdellweg:
Thanks for the detailed review.
For the signature checking policy i was inspired by apt's way, where you add a bunch of keys with 'apt-key' and i really does not matter who signed what.
Are you suggesting to specify the complete key fingerprint in the 'verify-release-with', or even better a list of viable signing keys fprs?

mibanescu:
Yes, I was thinking a list of viable signing key fingerprints. While it may not be exactly what apt does, I think there is a reason. pulp is inherently a multi-user system. More than one group may be using it. They may manage (and mirror) different sets of apt repositories, and having shared keyrings does not seem appropriate.

mdellweg:
Fair enough! And if we only accept whole fingerprints, it is as secure as it can get. ACK.

mdellweg:
@mibanescu Is this, what you thought about?

mibanescu:
Before I look at the PR, I wanted to share what the Pulp developers pointed me to: https://github.com/pulp/pulp_rpm/commit/01b21befdd719884ca0044253fbbe60c2f65f68c - the signature checking implemented in yum.

pulp-deb.spec
```diff
@@ -91,6 +91,7 @@ Requires: python-%{name}-common = %{version}-%{release}
 Requires: pulp-server >= %{pulp_version}
 Requires: python-debian
 Requires: python-debpkgr >= 1.0.0
+Requires: python-gnupg
```
mibanescu: I think it's python2-gnupg in EPEL repos.

mdellweg: Yes, i keep mixing this with the debian package names...

mdellweg:
@mibanescu I think, that should be it (with a test ^_^)

plugins/pulp_deb/plugins/importers/sync.py
```diff
@@ -123,12 +128,40 @@ class ParseReleaseStep(publish_step.PluginStep):
     def __init__(self, *args, **kwargs):
         super(ParseReleaseStep, self).__init__(*args, **kwargs)
         self.description = _('Parse Release Files')
+
+    def verify_release(self, release):
+        # check if Release file exists
+        if not os.path.isfile(self.parent.release_files[release]):
+            raise Exception("Release file not found. Check the feed option.")
+        # check signature
+        if self.get_config().get_boolean(constants.CONFIG_REQUIRE_SIGNATURE, False):
```
mibanescu: As a matter of personal preference, I would test the negative and return. That way the rest of the code is one level of indentation less, and seems easier to follow. But everything else looks good, and if you prefer not to change that part, let me know, and I will merge it.

mdellweg:
`Requires: python2-gnupg`
Has been lost while rebasing, due to the spec-file being deleted.

PR #39, opened 2018-02-28
Always publish architecture 'all'

m-bucher:
pulp_deb does not publish anything, if no architectural package is in the repo.
If we would always publish the all architecture, we should be safe.


mibanescu:
Markus, can you please add your name to AUTHORS? Thank you! The patch looks good, sorry it took me so long to look over it.

m-bucher:
@mibanescu Thanks for reviewing, I just updated the changes with the modified AUTHORS file.

PR #43, opened 2018-03-16
gpg_cmd is not allowed as plugin or override configuration

mibanescu:
Since the command configured with gpg_cmd executes remotely as user apache,
a user should not be allowed to change it via a distributor config or
an override at publish time.

Fixes #3498
https://pulp.plan.io/issues/3498

Change-Id: I88cdb4f51c237b1157e7424863df7049269939ca

### DIFF
--- a/plugins/pulp_deb/plugins/distributors/configuration.py
+++ b/plugins/pulp_deb/plugins/distributors/configuration.py
@@ -18,6 +18,8 @@ OPTIONAL_CONFIG_KEYS = (HTTP_PUBLISH_DIR_KEYWORD, HTTPS_PUBLISH_DIR_KEYWORD,
                         PUBLISH_DEFAULT_RELEASE_KEYWORD,
                         GPG_CMD, GPG_KEY_ID)
 
+LOCAL_CONFIG_KEYS = [GPG_CMD]
+
 ROOT_PUBLISH_DIR = '/var/lib/pulp/published/deb'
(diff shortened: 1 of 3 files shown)


mdellweg:
At a first glance this looks very reasonable (especially the tests).
Trying to take a closer look next week. :+1:

dkliban:
ok test

dkliban:
ok test

dkliban:
ok test

dkliban:
ok test

dkliban:
ok test

dkliban:
ok test

dkliban:
ok test

plugins/pulp_deb/plugins/distributors/configuration.py
```diff
@@ -40,10 +42,19 @@ def validate_config(repo, config, config_conduit):
     or not and why
     :rtype:  tuple of (bool, str or None)
     """
+    error_messages = []
+    msg = _('Configuration key [%(k)s] is not allowed in %(config)s configuration')
```
mdellweg: Maybe you can add a comment here, stating, that this needs to happen before the sqishing of the options into the flst dict below, because we need to check what option came from where.
This should prevent future us from accidantially 'cleaning up the order of execution'.

mibanescu: Added comment, thanks for the suggestion!

pcreech:
ok test

dkliban:
ok test

dkliban:
ok test

PR #44, opened 2018-06-06
Adds CCRC COMMITMENT file

bmbouter:
As passed with PUP-5, the CCRC shall be displayed on all GPLv2 Pulp
repositories.

https://github.com/pulp/pups/blob/master/pup-0005.md

https://pulp.plan.io/issues/3734
re #3734


bmbouter:
@mibanescu what are your thoughts on merging this?

PR #45, opened 2018-06-26
Fix installing the pip dependencies

mdellweg:
Since the spec file is no longer available, we cannot parse it any more when installing the pip dependencies.
On another thought, should they be available as rpm's already?

### DIFF
--- a/pulp-dev.py
+++ b/pulp-dev.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python2
-# -*- coding: utf-8 -*-
 
 import optparse
 import os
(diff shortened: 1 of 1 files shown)


mibanescu:
I have no easy way to verify the changes. I will keep an eye to when you remove the WIP (which I assume means work in progress) and just merge.

mdellweg:
Well my test case is no more than `vagrant up` in the `pulp/devel` repo with this one next to it.
"WIP:" I thought i had it and then realized it did not communicate its failure.

mdellweg:
I think, now i have it.
It must, however, go together with:
https://github.com/pulp/devel/pull/171

mdellweg:
Thanks @mibanescu !
Let's hope, this will not break again any time soon.

PR #46, opened 2018-06-28
Verify fingerprints compatibility

m-bucher:
Add backwards-compatibility for:
* older pulp-versions (`get_boolean()` without default-value)
* older gnupg-versions (`homedir` vs. `gnupghome`, order of parameters in `verify()`)

Moved determining keyserver out of loop.

### DIFF
--- a/plugins/pulp_deb/plugins/importers/sync.py
+++ b/plugins/pulp_deb/plugins/importers/sync.py
@@ -156,13 +156,14 @@ class ParseReleaseStep(publish_step.PluginStep):
         gpg = gnupg.GPG(homedir=os.path.join(self.get_working_dir(), 'gpg-home'))
         shared_gpg = gnupg.GPG(homedir=os.path.join('/', 'var', 'lib', 'pulp', 'gpg-home'))
 
+        keyserver = self.get_config().get(constants.CONFIG_KEYSERVER,
+                                          constants.CONFIG_KEYSERVER_DEFAULT)
+
         fingerprints = self.get_config().get(constants.CONFIG_ALLOWED_KEYS).split(',')
         # TODO check if full fingerprints are provided
         for fingerprint in fingerprints:
             if fingerprint not in [
                     key['fingerprint'] for key in shared_gpg.list_keys()]:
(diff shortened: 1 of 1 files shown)


plugins/pulp_deb/plugins/importers/sync.py
```diff
@@ -317,5 +335,5 @@ def generate_internal_storage_path(filename):
 
 def split_or_none(data):
     if data:
-        return data.split(',')
+        return map(lambda x: x.strip(), data.split(','))
```
mibanescu: For readability I typically prefer list comprehensions over map:

return [x.strip() for x in data.split(',')]

If you return None, wouldn't that break line 157 (for fingerprint in fingerprints)?

plugins/pulp_deb/plugins/importers/sync.py
```diff
@@ -129,31 +130,48 @@ def __init__(self, *args, **kwargs):
...
                 shared_gpg.recv_keys(keyserver, fingerprint)
         gpg.import_keys(shared_gpg.export_keys(fingerprints))
         if not os.path.isfile(self.parent.release_files[release] + '.gpg'):
             raise Exception("Release.gpg not found. Could not verify release integrity.")
-        with open(self.parent.release_files[release]) as f:
-            verified = gpg.verify_file(f, self.parent.release_files[release] + '.gpg')
-            if not verified.valid:
-                raise Exception("Verification of Release failed! {}".format(verified.stderr))
+
+        if LooseVersion(gnupg.__version__) < LooseVersion('1.0.0'):
```
mibanescu: Also for readability I would create a local variable rel_file = self.parent.release_files[release] and use it.

I don't think I understand how gpg.verify_file works (in old or new gpg). Is the order of the arguments reversed?

m-bucher: Yes, for some reason (probably to make it consistent with the gpg-command syntax) they reversed the order of arguments:
https://github.com/isislovecruft/python-gnupg/commit/2fb88ad6b4abd04239b1544ab775b49e63aab88c

PR #49, opened 2018-08-14
Fix repo metadata information

rabadin:
Fix the 'Label' and the 'Description' in the repo metadata (Release file).
This enables Apt Pinning because pinning uses this metadata to distinguish between repos.

Fixes #3917
https://pulp.plan.io/issues/3917


rabadin:
Thanks @mibanescu, I've rebased onto current master and integrated your patch (with a minor change to the import part and the test).

PR #52, opened 2018-09-24
Account for components with an extra path prefix

quba42:
The components in the repo metadata object ('Release' file) may be
prefixed. This prefix must be stripped from certain comparisons.
See also: https://wiki.debian.org/DebianRepository/Format#Components

~~Also sets the 'suite' correctly~~. (EDIT: This has already been fixed in master).

Fixes #4008
https://pulp.plan.io/issues/4008

~~WARNING: These changes are dependent on changes in python-debpkgr to work (as well as to fully resolve the issue).
Do not merge/release before the following pull request has made it into python-debpkgr:
https://github.com/sassoftware/python-debpkgr/pull/13~~

EDIT: With the merge of https://github.com/pulp/pulp_deb/pull/57 into master, the dependency to python-debpkgr https://github.com/sassoftware/python-debpkgr/pull/13 no longer exists!

### DIFF
--- a/plugins/pulp_deb/plugins/importers/sync.py
+++ b/plugins/pulp_deb/plugins/importers/sync.py
@@ -226,7 +226,7 @@ class ParseReleaseStep(publish_step.PluginStep):
                 pass
             # get release component units
             for component in repometa.components:
-                if components is None or component in components:
+                if components is None or component.split('/')[-1] in components:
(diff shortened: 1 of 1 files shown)


quba42:
Are the pulp_deb tests broken on GitHub? This should not be passing all tests in it's current form... (At least it is failing several tests on my local `pulp/devel` box.)

PR #53, opened 2018-09-26
Add missing gnupg dependency + correct indenting in setup.py

ekohl:
bc9c24693cb9c6db8cbcd565d11ed6ecb6ad178f added gnupg as a dependency but it wasn't declared as such.


ekohl:
@pcreech this also needs a packaging change. This is reproducible in nightlies with pulp-deb-plugins-1.8.0-0.1.beta.el7.noarch. Installing python2-gnupg from EPEL at least ensures I can start up, haven't verified the actual functionality.

PR #54, opened 2018-09-27
Beta 2

zjhuntin:


pcreech:
@mibanescu This is for the controlled 1.8-release branch, where we are cherry-picking the change back from master to get ready for release

common/setup.py
```diff
@@ -2,7 +2,7 @@
 
 setup(
     name='pulp_deb_common',
-    version='1.8b1',
+    version='2.7.1b2',
```
ekohl: This looks like an odd version jump

plugins/setup.py
```diff
@@ -2,7 +2,7 @@
 
 setup(
     name='pulp_deb_plugins',
-    version='1.8b1',
+    version='2.7.1b2',
```
pcreech: This should be 1.8b2

PR #57, opened 2018-11-05
MetadataStep performance

quba42:
This pull request aims to improve the performance of the MetadataStep when dealing with large Debian repositories (e.g. stretch/main). This is achieved by completely removing the debpkgr dependency from this step. The needed functionality is instead implemented directly in pulp_deb.

This pull request is intended as a pure code refactor that does not add or remove any features. (Though I have several ideas for follow on improvements including bug fixes and new features that would be easy to implement on top of these changes). The only "feature changes" I am aware of is that the relative order of fields in both 'Packages' and 'Release' files will have changed (I used the same order I found in upstream Debian Stretch mirrors), and that I no longer write redundant information into 'Release' files (debpkgr would write the codename into various non-codename fields).

I do not (yet) have any systematic benchmarks, but my initial tests suggest that the performance improvement is quite substantial. (I can reliably sync repositories with more than 50000 packages, on a system with 12 GiB of RAM. In the past such repositories have routinely lead to memory allocation failures on systems with 32 GiB of RAM.)

I am happy to answer any questions and would greatly appreciate any feedback so far.

### DIFF
--- a/plugins/pulp_deb/plugins/db/models.py
+++ b/plugins/pulp_deb/plugins/db/models.py
@@ -21,6 +21,9 @@ class DebPackage(FileContentUnit):
     name = mongoengine.StringField(required=True)
     version = mongoengine.StringField(required=True)
     architecture = mongoengine.StringField(required=True)
+    # Note: checksumtype and checksum should only be used for pulp internals.
+    # Use md5sum, sha1, and sha256 for publishing and similar.
+    # Currently, checksumtype should always be 'sha256'
     checksumtype = mongoengine.StringField(required=True)
     checksum = mongoengine.StringField(required=True)
 
@@ -30,6 +33,11 @@ class DebPackage(FileContentUnit):
     # List of required fields:
     REQUIRED_FIELDS = ['name', 'version', 'architecture', 'checksumtype',
                        'checksum', 'filename']
+
+    # Named checksum fields:
+    md5sum = mongoengine.StringField()
+    sha1 = mongoengine.StringField()
+    sha256 = mongoengine.StringField()
 
     # Other non control file fields:
     # Note: Relativepath does not appear to be meaningfully in use.
@@ -121,13 +129,15 @@ class DebPackage(FileContentUnit):
         initialization_params.update(cls._to_internal_dict_style(control_fields))
         initialization_params = cls._parse_rel_fields(initialization_params)
 
-        with open(filename) as input_file:
-            checksum = cls._compute_checksum(input_file)
+        checksums = cls.calculate_deb_checksums(filename)
 
         initialization_params.update(
             size=getsize(filename),
             checksumtype=util.TYPE_SHA256,
-            checksum=checksum,
+            checksum=checksums['sha256'],
+            md5sum=checksums['md5sum'],
+            sha1=checksums['sha1'],
+            sha256=checksums['sha256'],
             control_fields=control_fields,
         )
 
@@ -164,13 +174,18 @@ class DebPackage(FileContentUnit):
         initialization_params = cls._to_internal_dict_style(packages_paragraph)
         initialization_params = cls._parse_rel_fields(initialization_params)
 
-        # Since 'Size' is not a control file field we need to handle it separately:
+        # Handle non control file fields separately:
         if 'Size' in packages_paragraph:
             initialization_params['size'] = packages_paragraph['Size']
+        if 'MD5sum' in packages_paragraph:
+            initialization_params['md5sum'] = packages_paragraph['MD5sum']
+        if 'SHA1' in packages_paragraph:
+            initialization_params['sha1'] = packages_paragraph['SHA1']
 
         initialization_params.update(
             checksumtype=util.TYPE_SHA256,
             checksum=checksum,
+            sha256=checksum,
             control_fields=control_fields,
         )
 
@@ -220,11 +235,6 @@ class DebPackage(FileContentUnit):
                 missing_fields.append(field)
         if missing_fields:
             raise Error('Required fields are missing: {}'.format(missing_fields))
-
-    @classmethod
-    def _compute_checksum(cls, fobj):
-        cstype = util.TYPE_SHA256
-        return util.calculate_checksums(fobj, [cstype])[cstype]
 
     @staticmethod
     def calculate_deb_checksums(input_file_path):
--- a/plugins/pulp_deb/plugins/db/models.py
+++ b/plugins/pulp_deb/plugins/db/models.py
@@ -226,6 +226,25 @@ class DebPackage(FileContentUnit):
         cstype = util.TYPE_SHA256
         return util.calculate_checksums(fobj, [cstype])[cstype]
 
+    @staticmethod
+    def calculate_deb_checksums(input_file_path):
+        """
+        Uses util.calculate_checksums() to calculate the md5sum, sha1, and sha256
+        of a file. The return dict is guaranteed to use the deb style keys
+        'md5sum', 'sha1', and 'sha256'.
+
+        :param file_path: the path to the file for which checksums are needed
+        :returns: a dict containing the checksums
+        """
+        CHECKSUM_TYPES = {
+            'md5sum': util.TYPE_MD5,
+            'sha1': util.TYPE_SHA1,
+            'sha256': util.TYPE_SHA256,
+        }
+        with open(input_file_path) as input_file:
+            checksums = util.calculate_checksums(input_file, CHECKSUM_TYPES.values())
+        return {deb_key: checksums[util_key] for deb_key, util_key in CHECKSUM_TYPES.items()}
+
     @classmethod
     def filename_from_unit_key(cls, unit_key):
         return "{0}_{1}_{2}.{3}".format(
--- a/plugins/pulp_deb/plugins/distributors/configuration.py
+++ b/plugins/pulp_deb/plugins/distributors/configuration.py
@@ -107,7 +107,7 @@ def validate_config(repo, config, config_conduit):
                                        error_messages)
 
     try:
-        get_gpg_sign_options(repo, config)
+        get_gpg_signer(repo, config)
     except signer.SignerError as e:
         error_messages.append(str(e))
 
@@ -201,18 +201,20 @@ def get_repo_relative_path(repo, config=None):
     return relative_path.lstrip('/')
 
 
-def get_gpg_sign_options(repo=None, config=None):
+def get_gpg_signer(repo=None, config=None):
     cfg = config or {}
     cmd = cfg.get(GPG_CMD)
     if not cmd:
         return None
     key_id = cfg.get(GPG_KEY_ID)
     repository_name = None if repo is None else repo.id
-    return signer.SignOptions(cmd, repository_name=repository_name,
-                              key_id=key_id)
+    sign_options = signer.SignOptions(cmd, repository_name=repository_name,
+                                      key_id=key_id)
+    return signer.Signer(sign_options)
 
 
 # -- required config validation -----------------------------------------------
+
 
 def _validate_http(http, error_messages):
     _validate_boolean(HTTP_PUBLISH_DIR_KEYWORD, http, error_messages)
@@ -238,6 +240,7 @@ def _validate_relative_url(relative_url, error_messages):
 
 
 # -- optional config validation -----------------------------------------------
+
 
 def _validate_http_publish_dir(http_publish_dir, error_messages):
     _validate_usable_directory(HTTP_PUBLISH_DIR_KEYWORD, http_publish_dir,
--- a/plugins/pulp_deb/plugins/distributors/distributor.py
+++ b/plugins/pulp_deb/plugins/distributors/distributor.py
@@ -290,14 +290,6 @@ class MetadataStep(PluginStep):
         config = self.get_config()
         base_path = self.get_working_dir()
 
-        # Add missing checksum fields (this is bad for performance):
-        # This is a temporary fix!
-        for package in unit_dict.itervalues():
-            checksums = models.DebPackage.calculate_deb_checksums(package.storage_path)
-            package.sha1 = checksums['sha1']
-            package.sha256 = checksums['sha256']
-            package.md5sum = checksums['md5sum']
-
         # Do nothing, if the repository is empty (review this behaviour):
         if len(unit_dict) == 0:
             return
--- a/plugins/pulp_deb/plugins/distributors/distributor.py
+++ b/plugins/pulp_deb/plugins/distributors/distributor.py
@@ -13,8 +13,12 @@ from pulp.plugins.distributor import Distributor
 
 from pulp_deb.common import ids, constants
 from pulp_deb.plugins.db import models
+from metadata_files import (write_packages_file,
+                            write_release_file,
+                            gzip_compress_file,
+                            bz2_compress_file,)
+
 from . import configuration, yum_plugin_util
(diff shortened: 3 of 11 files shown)


quba42:
I was thinking that perhaps I should explain my rational behind this pull request a bit more. The basic idea is to improve performance by not parsing the same package files multiple times.

In the current implementation of the MetadataStep, the metadata associated with a repository is loaded from the db into memory, and is then processed to gain a list of all packages associated with each architecture for each component for each release in the repository. Particularly packages with architecture = "all" will be appended to every other architectures list (as well as a list of their own), meaning these packages will appear in multiple lists.

Each list is then processed by a call to python-debpkgr, which proceeds to access every package file in the list, mostly to regenerate the metadata we already loaded into memory from our db. For repositories containing thousands or tens of thousands of packages (often within a single debpkgr call) this is a real resource bottleneck.

The implementation of this pull request avoids all that by using the metadata from the db directly. This does mean reimplementing some of the functionality previously supplied by debpkgr, but I believe this to be justified both by the performance improvement, as well as the greater level of control it gives us over the MetadataStep.

Once merged, it will become much easier to implement certain features and fix certain bugs I have identified, since we will no longer depend on debpkgr supplying us with the needed interfaces.

Possible New Features post merge:
* Adding an 'InRelease' file (instead of just a 'Release' file)
* Compressing 'Packages' files using xz instead of bz2 as is common in official Debian repositories.
* Improving our pool folder structure. (Currently we just use "pool/<component>/<files>", whereas Debian uses "pool/<component>/<layer1>/<layer2>/<files>". Here, <layer1> is a single letter or number, or "lib" concatenated with a single letter or number. <layer2> is a source package name.

Possible further performance improvements:
In it's current form this pull request does not change what is stored in the db. There are several problems with this. Firstly, the db currently only stores a single checksum (along with it's type) for debian packages. However, we want to write SHA1, SHA256, and MD5sum into 'Packages' files. Currently this pull request has to regenerate these hashes for every package file in the repository and monkey patch them into the objects retrieved from the db. It would be better to store all three hashes in the db directly.

Secondly there should be arch_units_deb in additon to release_units_deb, comp_units_deb, and units_deb. Mainly to fix the following bug (https://pulp.plan.io/issues/4094), but also for some further performance gains in the MetadataStep, as well as for reasons of conceptual completeness.

quba42:
It is possible this PR also fixes https://pulp.plan.io/issues/3750
I was able to synchronize the repository described in the issue report using the changes in this PR.
However, I was not able to reproduce the issue without the changes in this PR either, since my test setup ran out of memory.
I will have another go at it with a more powerful test system tomorrow.

quba42:
I am about ready to rebase this PR one more time with the following new features:
1) Using deb822 from python-debian to write the Release and Packages files
2) A change to the db model to store all needed hashes in the mongo db

common/pulp_deb/common/ids.py
```diff
@@ -3,7 +3,7 @@
 
 TYPE_ID_DEB = "deb"
 UNIT_KEY_DEB = (
-    "name", "version", "architecture", "checksumtype", "checksum")
+    "name", "version", "architecture", "sha256")
```
mibanescu: I would prefer if we kept the debian units in sync with other pulp plugins' units that have a checksumtype and a checksum. In those cases, as well, the checksum is pretty much guaranteed to be a sha256. You carefully added them as fields on lines 90-92, but I have a lot of code that expects all unit types to have a "checksum" unit key (not just field). I don't have a problem with an extra sha256 field, even if redundant, as long as it's not part of the key.

plugins/pulp_deb/plugins/distributors/configuration.py
```diff
@@ -201,19 +201,21 @@ def get_repo_relative_path(repo, config=None):
...
     cmd = cfg.get(GPG_CMD)
     if not cmd:
         return None
     key_id = cfg.get(GPG_KEY_ID)
     repository_name = None if repo is None else repo.id
-    return signer.SignOptions(cmd, repository_name=repository_name,
-                              key_id=key_id)
+    sign_options = signer.SignOptions(cmd, repository_name=repository_name,
+                                      key_id=key_id)
+    return signer.Signer(sign_options)
```
mibanescu: I like the change, I just need to remember to port it into pulp_rpm, where we have almost the same code for signing repository metadata files.

plugins/pulp_deb/plugins/distributors/metadata_files.py
```diff
@@ -0,0 +1,243 @@
...
+            if not data:
+                break
+            checksums['md5sum'].update(data)
+            checksums['sha1'].update(data)
+            checksums['sha256'].update(data)
+
+    checksums['md5sum'] = checksums['md5sum'].hexdigest()
+    checksums['sha1'] = checksums['sha1'].hexdigest()
+    checksums['sha256'] = checksums['sha256'].hexdigest()
+
```
mibanescu: There is a lot of repetition here. May I suggest replacing lines 186-192 with:

            for halg in checksums.values():
                halg.update(data)
    checksums = dict((hname, halg.hexdigest()) for hname, halg in checksums.items())

... re-editing after I read more of the code: any reason not to use pulp.server.util.calculate_checksums?

quba42:
> I ran out of time (and patience) following the big code change, but I promise to come back to it :-)

I do appreciate all the detailed feedback! I understand this is a large potential change.

common/pulp_deb/common/ids.py
quba42: I will revert the unit key change. I was only thinking I should provide a best effort to use the newer fields, but I was unaware of the parallel structure with other plugins. Keeping this in sync makes sense to me.

quba42: The commit dfe46ef7c7da8853c79b1c77ca02ae9db8c142a9 "REBASE_BEFORE_MERGE: Revert change of unit key" now reverts this change.

plugins/pulp_deb/plugins/distributors/metadata_files.py
quba42: I kept my function as a wrapper of the server function to ensure the return dict uses keys I can loop over. (See 4afbb32bfd459d932c05877d1bfcbd0f16a8acf5 "REBASE_BEFORE_MERGE: Use server hash function everywhere").

Let me know if you have a better idea.

mdellweg: Maybe call it `deb_checksums`, to clarify that this naming is rather specific.

plugins/pulp_deb/plugins/migrations/0002_add_multiple_hashes.py
```diff
@@ -0,0 +1,40 @@
...
+
+_logger = logging.getLogger(__name__)
+
+
+def migrate(*args, **kwargs):
+    """
+    Add seperate fields for md5sum, sha1, sha256
+    """
+    deb_package_collection = connection.get_collection('units_deb')
+    for deb_package in deb_package_collection.find({}):
```
goosemania: @quba42 , add `batch_size(100)` to avoid cursor timeout.
Like here https://github.com/pulp/pulp_rpm/blob/2-master/plugins/pulp_rpm/plugins/migrations/0040_errata_pkglist_collection.py#L37

Also take a look at the line above, you can print progress of your migration which helps users not to worry that something is stuck

quba42:
> If you made `CHECKSUM_TYPES` a dict, then you had the translation for the different names in one place, and it would be easier to include e.g. `SHA512` somewhere in the future.
> 
> I think of somthing along the lines of `util.calculate_checksums(input_file, list(CHECKSUM_TYPES.values()))` and
> `return { k, checksums[v] for k,v in CHECKSUM_TYPES.items() }`

I tried to incorporate these changes in d13f5a4702e056f7a82e29426d11191779e4ada3 "REBASE_BEFORE_MERGE: Incorporate calculate_deb_checksums() suggestions".

I also renamed the function to `calculate_deb_checksums` as suggested above.

quba42:
I found another issue during testing:
https://pulp.plan.io/issues/4176

I will need to add an additional fix for this bug to the PR.

plugins/pulp_deb/plugins/db/models.py
```diff
@@ -100,9 +105,23 @@ def from_metadata(cls, unit_md, user_metadata=None):
...
+
+        :param file_path: the path to the file for which checksums are needed
+        :returns: a dict containing the checksums
+        """
+        CHECKSUM_TYPES = {
+            'md5sum': util.TYPE_MD5,
+            'sha1': util.TYPE_SHA1,
+            'sha256': util.TYPE_SHA256,
+        }
+        with open(input_file_path) as input_file:
```
quba42: Should I be catching `IOError` and raise it as `Error` similar to lines 174 to 175?

quba42:
Since it was getting rather messy, I have rebased the branch.
The current version still cannot deal with this issue: https://pulp.plan.io/issues/4176
(Which should really be resolved in a separate PR.)

quba42:
https://github.com/pulp/pulp_deb/pull/61 fixes https://pulp.plan.io/issues/4176.
It also ads a `control_fields` dict containing all control file fields.
(This is essentially the suggestion by @m-bucher above).

This PR is blocked until https://github.com/pulp/pulp_deb/pull/61 is merged at which point I will need to rebase and make minor adjustments.

plugins/pulp_deb/plugins/distributors/metadata_files.py
quba42: The function is now a part of the `DebPackage` class. It is named `calculate_deb_checksums` and essentially replaces `_compute_checksum`.

plugins/pulp_deb/plugins/migrations/0003_add_multiple_hashes.py
```diff
@@ -0,0 +1,46 @@
...
+    Add seperate fields for md5sum, sha1, sha256
+    """
+    deb_package_collection = connection.get_collection('units_deb')
+    deb_package_count = deb_package_collection.count()
+
+    with migration_utils.MigrationProgressLog('Deb Package', deb_package_count) as progress_log:
+        for deb_package in deb_package_collection.find({}).batch_size(100):
+            storage_path = deb_package['_storage_path']
+            if not os.path.exists(storage_path):
+                continue
```
m-bucher: Should we log these orphans?

quba42: You are right, I should add the same logging as I did in migration 0002.

quba42: I added the same logging as I did in migration 0002.
I will mark this conversation as resolved. Feel free to reopen.

plugins/pulp_deb/plugins/importers/sync.py
```diff
@@ -317,20 +317,33 @@ def __init__(self, *args, **kwargs):
...
+                unit.sha1 = checksums['sha1']
+            if unit.md5sum:
+                self.verify_checksum(checksums['md5sum'], unit.md5sum, path)
+            else:
+                unit.md5sum = checksums['md5sum']
+            if unit.sha256:
+                self.verify_checksum(checksums['sha256'], unit.sha256, path)
+            else:
+                unit.sha256 = checksums['sha256']
+            self.verify_checksum(unit.checksum, unit.sha256, path)
```
quba42: Is this what we want? Do we want to check every checksumtype? Do we want to throw an error as soon as we encounter a broken checksum, or collect any errors from all three types and then throw a single error?

plugins/pulp_deb/plugins/distributors/distributor.py
```diff
@@ -282,103 +286,132 @@ def process_main(self, item=None):
...
+            # Create the 'Release' file (for each release):
+            release_meta_data['architectures'].remove('all')
+            if repo.description:
+                release_meta_data['description'] = repo.description
+            release_meta_data['label'] = repo.id
+            release_file_path = write_release_file(release_path,
+                                                   release_meta_data,
+                                                   release_meta_files,)
+
+            signer = configuration.get_gpg_signer(repo, config)
```
quba42: Should I move this somewhere up to line 291 where similar variables are being initialized or keep it down here where it is actually used?

plugins/pulp_deb/plugins/importers/sync.py
quba42: I am leaning heavily towards we do want to check every checksumtype. We just don't want broken checksums/files to make it into our data base...
I am unsure how `PulpCodedTaskFailedException` works, so I would have to look that up to make any changes to what is being thrown as an error.

plugins/pulp_deb/plugins/distributors/distributor.py
```diff
@@ -282,103 +286,132 @@ def process_main(self, item=None):
...
+
+                        # Compress and record 'Packages' files:
+                        release_meta_files.append(packages_file_path)
+                        gz_file_path = gzip_compress_file(packages_file_path)
+                        release_meta_files.append(gz_file_path)
+                        bz2_file_path = bz2_compress_file(packages_file_path)
+                        release_meta_files.append(bz2_file_path)
+
+            # Create the 'Release' file (for each release):
+            release_meta_data['architectures'].remove('all')
```
m-bucher: Not sure if this is a use-case, but I had the problem that pulp_deb tried to publish an empty repository (because I did not specify the components correctly).
In this special case, the `all`-component is not added to `release_meta_data['architectures']` and therefore the `remove()` throws a `KeyError`.

So it probably makes no sense to publish an empty-repository, but in that case we should throw a human-readable error-message. I made the Exception go away by changing  the `remove()` to `discard()` in order to see what happens.

The other thing, which will probably have to be a different PR is, whether the importer should have thrown an error because of my invalid component-definition (e.g. 'could not find component XYZ').

mibanescu: I think publishing an empty repository is a valid case. You should get something that apt won't choke on. A missing repository is unlikely going to make apt happy.

m-bucher: @quba42 I suppose for a valid empty apt-repository the `Packages` should be there but can be empty, which is currently not the case.

quba42: I will have a go at comparing the pulp_deb behaviour with and without my PR.
I will also try to find out what exactly a valid but empty Debian repository would look like.

The fact that there is no relevant error message if the user requests non existent components strikes me as an independent issue, but I will have a look at that as well.

quba42: So, the existing behaviour (in master) is for the sync task to report success, but not actually publish anything (not even a `Release` file). I could easily recreate this behaviour, but I am not sure it's what we want.

I could just recreate the behaviour from master, and open a new ticket that we should look into empty repository publishing. (This PR was never meant to add any features after all.)

plugins/pulp_deb/plugins/distributors/distributor.py
```diff
@@ -282,103 +286,136 @@ def process_main(self, item=None):
...
-                                 with_symlinks=True)
-
-        # Prepare generic releases containing all packages in one component
-        generic_release_names = []
-
-        # In case, no release_units were available (old style repository),
-        # publish as 'stable/main'
+        # Do nothing, if the repository is empty (review this behaviour):
+        if len(unit_dict) == 0:
+            return
```
quba42: My proposal to resolve the above conversation.
@mibanescu  Let me know if you prefer something else...

PR #59, opened 2018-11-08
Satisfy flake8

quba42:
Without this change Travis CI builds will always fial.


quba42:
Thanks for merging, I have now deleted the issue branch.

PR #61, opened 2018-12-08
Ensure the db is used consistently

quba42:
This PR aims to resolve the following issue: https://pulp.plan.io/issues/4176

~~This PR is still "WIP".
I have not yet written any tests or fixed any tests I may have broken.~~

However, the basic implementation is done and appears to be working, so I opened this PR to start discussing it.


plugins/pulp_deb/plugins/db/models.py
```diff
@@ -62,6 +85,126 @@ class DebPackage(FileContentUnit):
...
+    def _parse_rel_fields(cls, field_dict):
+        """
+        Converts any relatinal fields found in field_dict from plain Debian
+        package relation strings to a parsed dict.
+        """
+        for rel_field in cls.REL_FIELDS:
+            if rel_field in field_dict:
+                #TODO: I think the initial_parse should really be performed
+                #within the DependencyParser class. However, first we must
+                #ensure the old version is no longer in use anywhere...
```
quba42: I suppose this TODO is really a question for @mibanescu
Do you agree with this assessment or should I leave the DependencyParser class (and it's interfaces) untouched?

plugins/pulp_deb/plugins/db/models.py
```diff
@@ -33,9 +52,16 @@ class DebPackage(FileContentUnit):
     filename = mongoengine.StringField(required=True)
     relativepath = mongoengine.StringField()
 
+    REQUIRED_FIELDS = ['name', 'version', 'architecture', 'checksumtype',
+                       'checksum', 'filename']
```
quba42: This adds a second source of truth to the field definitions. Does anyone know of a better way to iterate over all required mongoengine fields?

plugins/pulp_deb/plugins/db/models.py
```diff
@@ -51,7 +77,6 @@ class DebPackage(FileContentUnit):
     _content_type_id = mongoengine.StringField(required=True, default=TYPE_ID)
 
     # Non-key fields
```
quba42: I am not sure what "Non-key fields" means in this context... If someone could enlighten me that would be great!

plugins/pulp_deb/plugins/db/models.py
```diff
@@ -62,6 +87,128 @@ class DebPackage(FileContentUnit):
...
+
+    @classmethod
+    def from_packages_paragraph(cls, packages_paragraph):
+        """
+        Creates a DebPackage object (and by extension a mongodb entry) from a
+        single 'Packages' file paragraph as parsed by python-debian's
+        deb822.Packages.iter_paragraphs function.
+        """
+
+        POSSIBLE_EXTRA_FIELDS = [
```
quba42: Should I make this a class constant next to `UNIT_KEY_TO_FIELD_MAP` or keep it down here?

plugins/pulp_deb/plugins/db/models.py
```diff
@@ -62,6 +87,128 @@ class DebPackage(FileContentUnit):
...
+        return return_value
+
+    @classmethod
+    def _parse_rel_fields(cls, field_dict):
+        """
+        Converts any relatinal fields found in field_dict from plain Debian
+        package relation strings to a parsed dict.
+        """
+        for rel_field in cls.REL_FIELDS:
+            if rel_field in field_dict:
```
quba42: The old behaviour (of the `from_file()`) function would place empty lists into the db for any `REL_FIELDS` that were not needed for the package in question. The new behaviour does not set these fields at all. This strikes me as more consistent with how we handle other optional fields. If this is a problem for software dependent on these fields it would be a trivial change to add the empty lists back in (please let me know @mibanescu).

plugins/pulp_deb/plugins/db/models.py
mdellweg: Looking at 625b644c718d61ee1122ef8b637d8efa5fc14551, it seems, that introducing the relational fields, every field above was somehow involved in identifying / finding the package.

plugins/pulp_deb/plugins/db/models.py
```diff
@@ -62,6 +87,128 @@ class DebPackage(FileContentUnit):
...
+    @classmethod
+    def from_deb_file(cls, input_file_path, user_metadata={}):
+        """
+        Creates a DebPackage object (and by extension a mongodb entry) from a
+        .deb package file.
+        """
+        try:
+            control_fields = debfile.DebFile(input_file_path).debcontrol()
+        except debfile.ArError as invalid_package_error:
+            raise InvalidPackageError(str(invalid_package_error))
```
mdellweg: Nice catch

plugins/pulp_deb/plugins/db/models.py
```diff
@@ -62,6 +87,128 @@ class DebPackage(FileContentUnit):
...
+
+        checksum = packages_paragraph['SHA256']
+
+        initialization_params = cls._to_internal_dict_style(packages_paragraph)
+        initialization_params = cls._parse_rel_fields(initialization_params)
+        filename = cls.filename_from_unit_key(initialization_params)
+
+        # Since 'Size' is not a control file field we need to handle it separately:
+        if 'Size' in packages_paragraph:
+            initialization_params.update(size=packages_paragraph['Size'])
```
mdellweg: Nitpicking: Calling `update` to store a single value in a dict seems a little over engineered.

quba42: Are you saying `initialization_params['size']=packages_paragraph['Size']` would be cleaner/more efficient?

quba42: I think I need to have a look at the docs for the `update` function.

mdellweg: Yes, that's what i meant. As i understand, update loops over it's input, which is one unnecessary loop here.

plugins/pulp_deb/plugins/db/models.py
quba42: Since this is no longer how the key is being built, I guess it is safe to change this comment? :smile:

plugins/pulp_deb/plugins/db/models.py
quba42: Done.

plugins/pulp_deb/plugins/db/models.py
quba42: Thanks! But I am afraid I cannot claim credit. Copied it from the `_read_metadata` function.

quba42:
Apart from any further requests for changes, this is just about done, so I will go ahead and rebase the branch.

plugins/pulp_deb/plugins/db/models.py
quba42: I now found a solution that achieves everything I want by adding an extra method to DependencyParser, without touching any existing interfaces.

I therefore consider this conversation resolved.

quba42:
I will now leave this PR be as it is (unless I find bugs that need fixing).
Do feel free to request additional changes!
I recommend we squish everything into a single commit before merging.

plugins/pulp_deb/plugins/db/models.py
```diff
@@ -62,42 +73,150 @@ class DebPackage(FileContentUnit):
...
+        section="Section",
+        priority="Priority",
+        multi_arch="Multi-Arch",
+        homepage="Homepage",
+        description="Description",
+        original_maintainer="Original-Maintainer",
+    )
+
+    @classmethod
+    def from_deb_file(cls, input_file_path, user_metadata={}):
```
mibanescu: Is there a reason for renaming from_file to from_deb_file? If it changed signature, I understand, but it doesn't seem to be the case, and it's kind of implied it's reading a Deb file anyway.

plugins/pulp_deb/plugins/db/models.py
mibanescu: I believe that's fine. I may have used empty lists as placeholders so that I don't break mongo queries on fields that don't exist, but I think mongo is forgiving enough.

plugins/pulp_deb/plugins/db/models.py
```diff
@@ -62,42 +73,150 @@ class DebPackage(FileContentUnit):
...
+    @classmethod
+    def _check_for_required_fields(cls, metadata):
+        """
+        Checks if the dict in metadata contains a value for all required db
+        fields except 'filename', which needs to be generated from the other
+        required fields.
+        """
+        for field in cls.REQUIRED_FIELDS:
+            if field == 'filename':
+                pass
```
mibanescu: Style preference (and you can disagree with that): I would use "continue" instead of "pass" here, and that changes the next line to a plain if instead of elif.

plugins/pulp_deb/plugins/db/models.py
```diff
@@ -62,42 +73,150 @@ class DebPackage(FileContentUnit):
...
+            elif field not in metadata:
+                raise Error('Required field is missing: {}'.format(field))
```
mibanescu: Functionally, I would like to know all the fields that are missing, instead of finding out about them one at the time. So maybe collect all the missing fields in a list, and if the list is non-empty, raise the error with the whole list.

plugins/test/unit/plugins/importers/test_importer.py
```diff
@@ -222,6 +222,7 @@ def test_upload_unit_deb(self, _plugin_api, from_file,
             homepage=None,
             description=None,
             size=None,
+            control_fields=None,
         )
         for fname in models.DebPackage.REL_FIELDS:
             metadata[fname] = None
```
mibanescu: These changes wouldn't be here if it weren't for the function being renamed :-)

plugins/pulp_deb/plugins/db/models.py
quba42: There is a reason for renaming it, but not for keeping it renamed:
When I first added `from_deb_file` I still kept the old `from_file` function since the first db migration was still dependent on it. I also wasn't sure from the outset that they would ultimately end up so similar.

Now that all dependencies to the old function were removed there is no reason not to revert the name change. As you noticed this will make a number of lines of diff disappear, which should make the entire thing a bit more tidy.

I will revert the name change.

plugins/pulp_deb/plugins/db/models.py
```diff
@@ -62,42 +73,153 @@ class DebPackage(FileContentUnit):
...
-                fobj = open(filename, "r")
-            except IOError as e:
-                raise Error(str(e))
-        unit_md = cls._read_metadata(filename)
-        unit_md.update(checksumtype=util.TYPE_SHA256,
-                       checksum=cls._compute_checksum(fobj),
-                       size=fobj.tell())
-
-        return cls.from_metadata(unit_md, user_metadata)
+    def from_file(cls, input_file_path, user_metadata=None):
```
quba42: Should I rename the `input_file_path` parameter to `filename` for consistency with the old version of the `from_file` function?

plugins/pulp_deb/plugins/db/models.py
quba42: I will go ahead and mark this as "resolved" then.

plugins/pulp_deb/plugins/db/models.py
quba42: Done in 5f53f37e0f9cb64c7b8d3a3c8b286175f4a80e28.

quba42:
Thanks for the approval/merge, but shouldn't I have squashed some of those commits one more time?
(My apologies for not using auto-squash style commit messages, if that would have made a difference).

mibanescu:
If the commits made logical sense as a progression, I don't think we need to squash.

quba42:
Just because some of them still have "fixup" in the commit message...

But unless someone with the relevant permissions wants to unprotect master and force push to it, I guess they are in now. :stuck_out_tongue_winking_eye:

mibanescu:
meh. I should probably pay more attention, but it's fine.

quba42:
Well thanks for the insightful code review. I feel like you always catch some great adjustments that seem self evident in retrospect but hadn't been spotted by anyone else.

PR #60, opened 2018-11-14
Basic implementation

mdellweg:
~~OK, far from ready, but~~ this way, we can start discussing...

PR #63, opened 2018-12-18
Re-bootstrap

dralley:


mdellweg:
I think i have most of the changes already on #60 which i intent to merge soon, as it is in a state compatible with pulpcore now.

dralley:
It looks like you captured everything with that PR, so I'll close this one.

PR #64, opened 2019-01-03
Fix test_VerifySignature on devel

quba42:
The `test_VerifySignature` test will not run on my `devel_pulp2_dev` Vagrant box without this line.
Instead it throws a `raise Exception("No GPG-keys in keyring, did the import fail?")`.
Since Travis appears to be doing fine without this fix, it would be good to know if others are having the same problem, or what the difference between Travis and my local Vagrant box is...

Edit: I added a pulp.plan.io issue to go with this PR:
https://pulp.plan.io/issues/4332


mibanescu:
A riddle, wrapped in a mystery. Thanks!

PR #66, opened 2019-01-16
Force digest-algo to SHA256

TiJof:
Support for keys that state a preference for SHA-1 encryption has been turned off in APT as of Debian 9. (Specifically, it was turned off in APT version 1.4~beta1, and Debian 9 has version 1.4.7.)


mibanescu:
Thank you for the patch!

PR #73, opened 2019-02-26
Safely select packages to symlink in components

mdellweg:
When publishing an old copy of a Repository, it can happen, that the
component references packages not yet in that version.


quba42:
I can confirm that not having this change appears to cause problems when promoting Katello content views. (This pretty much systematically involves this rather special use case.)

The proposed change appears to fix this problem. Using `unit_dict.get(package_id)` instead of `unit_dict[package_id]` is also just plain good style, consistent with a similar usage further on in the file. Since the change essentially introduces an extra sanity check it is highly unlikely to break anything.

For all of these reasons I would recommend merging this ASAP, if possible prior to any release!

mibanescu:
Belated +2.

PR #76, opened 2019-03-05
Add documentation on InRelease files

quba42:
I have been asked regarding InRelease file support in pulp_deb by several different people by now.

Well, it turns out, this is merely a matter of configuration!

To let people know of this happy news, I have added some relevant documentation to the `README.md` file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,15 +199,19 @@ A minimal sign command using GPG could be:
 ```Shell
 #!/bin/bash -e
 
-KEYID=${GPG_KEY_ID:-45BA0816}
+if [ -z "${GPG_KEY_ID}" ]; then
+  echo "No Key specified for signing."
+  exit 1
+fi
 
-rm -f ${1}.gpg
+rm -f "${1}.gpg"
 
-gpg --homedir /var/lib/pulp/gpg-home \
-    --batch \
-    --digest-algo SHA256 \
-    --detach-sign --default-key $KEYID \
-    --armor --output ${1}.gpg ${1}
+gpg --batch --armor --detach-sign \
+  --digest-algo SHA256 \
+  --homedir /var/lib/pulp/gpg-home \
+  --default-key ${GPG_KEY_ID} \
+  --output "${1}.gpg" \
+  "${1}"
 ```
 
 You could import your password-less GPG keys like this:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -199,15 +199,20 @@ A minimal sign command using GPG could be:
 ```Shell
 #!/bin/bash -e
 
-KEYID=${GPG_KEY_ID:-45BA0816}
+if [ -z "${GPG_KEY_ID}" ]; then
+  echo "No Key specified for signing."
+  exit 1
+fi
```
quba42: I thought it better not to default to some random signing key in the example.
I wanted something people can copy and paste without changes.

README.md
```diff
@@ -199,15 +199,20 @@ A minimal sign command using GPG could be:
...
 
-rm -f ${1}.gpg
+PULP_WORKER_PATH="$(dirname ${1})"
 
-gpg --homedir /var/lib/pulp/gpg-home \
-    --batch \
-    --digest-algo SHA256 \
```
quba42: I just noticed I dropped the `--digest-algo SHA256` option. Should I add it back in for all three examples?

README.md
```diff
@@ -199,15 +199,20 @@ A minimal sign command using GPG could be:
 ```Shell
 #!/bin/bash -e
 
-KEYID=${GPG_KEY_ID:-45BA0816}
+if [ -z "${GPG_KEY_ID}" ]; then
+  echo "No Key specified for signing."
+  exit 1
```
quba42: I thought it better not to default to some random signing key in the example.
I wanted something people can copy and paste without changes.

README.md
quba42: So, I just added it back in. Debian buster also uses `SHA256`, and it is better to make the option explicit for the example.

README.md
mibanescu: Agreed.

README.md
```diff
@@ -199,15 +199,21 @@ A minimal sign command using GPG could be:
...
-    --detach-sign --default-key $KEYID \
-    --armor --output ${1}.gpg ${1}
+rm -f "${PULP_WORKER_PATH}/Release.gpg"
+
+gpg --batch --armor --detach-sign \
+  --digest-algo SHA256 \
+  --homedir /var/lib/pulp/gpg-home \
+  --default-key ${GPG_KEY_ID} \
+  --output "${PULP_WORKER_PATH}/Release.gpg" \
+  "${PULP_WORKER_PATH}/Release"
```
mibanescu: This is probably my only quibble with the change. I would have kept the "${1}" (and "${1}.gpg" in the line before). I understand  you made these changes to support InRelease, and it makes sense for that. Assuming ${1} is Release is fine for now, but should debpkgr ever change to send other things to the signer, you'd have to change the instructions again.

README.md
```diff
@@ -224,3 +230,63 @@ use. Unprotected GPG keys may be easily stolen. You may want to consider
...
+  --default-key ${GPG_KEY_ID}"
+
+gpg ${COMMON_GPG_OPTS} \
+  --detach-sign \
+  --output "${PULP_WORKER_PATH}/Release.gpg" \
+  "${PULP_WORKER_PATH}/Release"
+
+gpg ${COMMON_GPG_OPTS} \
+  --clearsign \
+  --output "${PULP_WORKER_PATH}/InRelease" \
```
mibanescu: This is the only place you're really using $(basename ${1}) outside of ${1}.

All I am saying is, use the path provided to the script as much as possible, without making an assumption it ends in Release.

Maybe I am thoroughly biased because I wrote the original script :-)

PR #81, opened 2019-03-28
Ignore invalid paragraphs in package indices

mdellweg:
If the long-description of a package contains an erroneous newline, the
rest of the paragraph is considered to describe a new package, but it is
incomplete. We try to ignore those and hope for the best.
Usually there is only the 'Homepage' field, so no functionality should be
affected.

Thanks to @sbernhard for analysing the problem.

### DIFF
--- a/plugins/pulp_deb/plugins/importers/sync.py
+++ b/plugins/pulp_deb/plugins/importers/sync.py
@@ -272,13 +272,17 @@ class ParsePackagesStep(publish_step.PluginStep):
                     if dlr.url in self.parent.packages_urls[release]])
             for ca in repometa.iter_component_arch_binaries():
                 for pkg in ca.iter_packages():
-                    checksum = pkg['SHA256']
-                    self.parent.unit_relative_urls[checksum] = pkg['Filename']
-                    if checksum in units:
-                        unit = units[checksum]
-                    else:
-                        unit = models.DebPackage.from_packages_paragraph(pkg)
-                        units[checksum] = unit
+                    try:
+                        checksum = pkg['SHA256']
+                        self.parent.unit_relative_urls[checksum] = pkg['Filename']
+                        if checksum in units:
+                            unit = units[checksum]
+                        else:
+                            unit = models.DebPackage.from_packages_paragraph(pkg)
+                            units[checksum] = unit
+                    except KeyError:
+                        _logger.warning(_("Invalid package record found. {}").format(pkg))
+                        continue
                     self.parent.component_packages[release][ca.component].append(unit.unit_key)
         self.parent.available_units = units.values()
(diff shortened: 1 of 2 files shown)


dkliban:
ok test

mdellweg:
@mibanescu Any objections to this?

plugins/pulp_deb/plugins/importers/sync.py
```diff
@@ -272,14 +272,17 @@ def process_main(self, item=None):
...
                 for pkg in ca.iter_packages():
-                    checksum = pkg['SHA256']
-                    self.parent.unit_relative_urls[checksum] = pkg['Filename']
-                    if checksum in units:
-                        unit = units[checksum]
-                    else:
-                        unit = models.DebPackage.from_packages_paragraph(pkg)
-                        units[checksum] = unit
-                    self.parent.component_packages[release][ca.component].append(unit.unit_key)
+                    try:
```
mibanescu: This try/except is a fairly large hammer. There are so many dictionary accesses here, that it's hard to tell which of the KeyError exceptions you're trying to ignore.

If it's from pkg['SHA256'] or pkg['Filename'], can you wrap just those two lines in try/except, and continue on the except branch?

If it's from models.DebPackage.from_packages_paragraph, then I'd do it there.

mibanescu: You knew I was going to say this - but is there any chance you can add a test for this? If there were a test, maybe I would have followed the code change easier. I know isolating the exact failure is typically hard, but if writing a test is impossible, then the code probably has to be refactored to allow for easier unit testing. No, I am not blaming anyone but me for the way the code looks like :-)

You could isolate that whole block into a class method so you can mock things easier.

mdellweg: I should have known... And i should be able to write a test.

mdellweg: You know, i'm the one to blame for the convoluted state the sync unittests are in.

mdellweg: I want to ignore whatever failed key acces from `pkg` i stumble across, because the problem is that `pkp` can be an invalid snippet (for example `pkg = {'homepage': 'http://...'}`).
So i think the `units[checksum]` is perfectly safe (guarded by the `if checksum in units`), and the `self.parent.component_packages[release][ca.component].append(unit.unit_key)' i will pull out of the try block.
Do you think that is sufficient?

mibanescu: Definitely better. I think at some point I'd like that to be split into a separate function, but it doesn't have to be now.

PR #83, opened 2019-04-19
Add a "distribution" field

quba42:
**DEPENDS ON:** https://github.com/pulp/pulp/pull/3919 (https://pulp.plan.io/issues/4871)
**Fixes**:
- https://pulp.plan.io/issues/4138
- https://pulp.plan.io/issues/4705
- https://pulp.plan.io/issues/4707

**Relates to**:
- https://pulp.plan.io/issues/3464 (fixes it in pulp, but also needs to be fixed in `python-debpkgr`)

**Background**

Debian repositories are generally structured into multiple "releases" or "distributions", each associated with a single "Release" file somewhere within the `/dists/` folder of the repository.

Most commonly such release files are placed within the folder in one of two ways:

    /dists/<codename>/Release
    /dists/<suite>/Release

Pulp currently publishes all `Release` files under the `<codename>` path.

However, the official specification allows for the following structure:

    /dists/<distribution>/Release

where `<distribution>` may include an arbitrary number of folder levels, and may differ from both `codename` and `suite`.

This leads to a whole class of unpredictable (read undesirable or buggy) behavior within pulp whenever upstream repositories use a `<distribution> != <codename>`.

**This Fix**

This fix aims to conclusively solve all related problems by introuducing internal "distribution" fields.
This field is then used instead of "codename" wherever relevant.

### DIFF
--- a/common/pulp_deb/common/ids.py
+++ b/common/pulp_deb/common/ids.py
@@ -13,12 +13,12 @@ EXTRA_FIELDS_DEB = set([
 
 TYPE_ID_DEB_COMP = 'deb_component'
 UNIT_KEY_DEB_COMP = (
-    'name', 'release', 'repoid')
+    'name', 'distribution', 'repoid')
 EXTRA_FIELDS_DEB_COMP = set([
     'packages'])
 
 TYPE_ID_DEB_RELEASE = 'deb_release'
 UNIT_KEY_DEB_RELEASE = (
-    'codename', 'repoid')
+    'distribution', 'repoid')
 
 SUPPORTED_TYPES = set([TYPE_ID_DEB, TYPE_ID_DEB_COMP, TYPE_ID_DEB_RELEASE])
--- a/extensions_admin/pulp_deb/extensions/admin/units_display.py
+++ b/extensions_admin/pulp_deb/extensions/admin/units_display.py
@@ -47,7 +47,7 @@ def _details_release(release):
     :return: The display string of the release
     :rtype: str
     """
-    return '%s' % (release['codename'])
+    return '%s' % (release['distribution'])
 
 
 def _details_component(component):
--- a/plugins/pulp_deb/plugins/db/models.py
+++ b/plugins/pulp_deb/plugins/db/models.py
@@ -300,7 +300,7 @@ class DebPackage(FileContentUnit):
 
 class DebComponent(ContentUnit):
     """
-    This unittype represents a deb release component
+    This unittype represents a deb release/distribution component
     """
     TYPE_ID = ids.TYPE_ID_DEB_COMP
     meta = dict(collection="units_deb_component",
@@ -308,6 +308,7 @@ class DebComponent(ContentUnit):
     unit_key_fields = ids.UNIT_KEY_DEB_COMP
 
     name = mongoengine.StringField(required=True)
+    distribution = mongoengine.StringField(required=True)
     release = mongoengine.StringField(required=True)
     repoid = mongoengine.StringField(required=True)
     packages = mongoengine.ListField()
@@ -316,11 +317,21 @@ class DebComponent(ContentUnit):
     _ns = mongoengine.StringField(required=True, default=meta['collection'])
     _content_type_id = mongoengine.StringField(required=True, default=TYPE_ID)
 
+    @property
+    def plain_component(self):
+        return self.name.strip('/').split('/')[-1]
+
+    @property
+    def prefixed_component(self):
+        prefix = '/'.join(self.distribution.split('/')[1:]).strip('/')
+        return (prefix + '/' + self.plain_component).strip('/')
+
     @classmethod
     def get_or_create_and_associate(cls, repo, release_unit, name):
         unit = cls()
         unit.name = name
         unit.repoid = repo.id
+        unit.distribution = release_unit.distribution
         unit.release = release_unit.codename
         try:
             unit.save()
@@ -338,12 +349,14 @@ class DebComponent(ContentUnit):
             # find the corresponding unit
             unit = self.__class__.objects.filter(repoid=repo.repo_id,
                                                  name=self.name,
-                                                 release=self.release).first()
+                                                 distribution=self.distribution).first()
             if unit is None:
                 # create a new one
                 unit = self.__class__()
+                # set the key fields
                 unit.repoid = repo.repo_id
                 unit.name = self.name
+                unit.distribution = self.distribution
                 unit.release = self.release
 
             # update data
@@ -356,7 +369,7 @@ class DebComponent(ContentUnit):
 
 class DebRelease(ContentUnit):
     """
-    This unittype represents a deb release
+    This unittype represents a deb release (also referred to as a "distribution")
     """
     TYPE_ID = ids.TYPE_ID_DEB_RELEASE
     meta = dict(collection="units_deb_release",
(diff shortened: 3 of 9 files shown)


quba42:
I added "WIP" to the title, since I was not yet able to test these changes in the real world.
The tests are all green on my `pulp2_dev` Vagrant box, but the pulp workers are broken so I could not test any real repositories.

Note also, that this change is not actually as big as it looks at first glance. Everything from commit 74e0cd568dd01caa1a38a0713b0093287c8ae5a2 just renames internal names for consistency and readability without actually changing any logic.

common/pulp_deb/common/ids.py
```diff
@@ -13,12 +13,12 @@
...
 UNIT_KEY_DEB_COMP = (
-    'name', 'release', 'repoid')
+    'name', 'distribution', 'repoid')
 EXTRA_FIELDS_DEB_COMP = set([
     'packages'])
 
 TYPE_ID_DEB_RELEASE = 'deb_release'
 UNIT_KEY_DEB_RELEASE = (
-    'codename', 'repoid')
+    'distribution', 'repoid')
```
quba42: The old keys are not actually guaranteed to be unique. That being said, I am not sure what pitfalls there may be in changing the unit keys. I kept this change in a separate commit so it is very easy to revert. ;-)

extensions_admin/pulp_deb/extensions/admin/units_display.py
```diff
@@ -47,7 +47,7 @@ def _details_release(release):
     :return: The display string of the release
     :rtype: str
     """
-    return '%s' % (release['codename'])
+    return '%s' % (release['distribution'])
```
quba42: This is the only user facing change...

plugins/pulp_deb/plugins/db/models.py
```diff
@@ -308,6 +308,7 @@ class DebComponent(ContentUnit):
     unit_key_fields = ids.UNIT_KEY_DEB_COMP
 
     name = mongoengine.StringField(required=True)
+    distribution = mongoengine.StringField(required=True)
     release = mongoengine.StringField(required=True)
```
quba42: This `release` field is essentially replaced by `distribution` (as in it isn't actually needed anymore). I kept it in anyway in case I've missed something...

plugins/pulp_deb/plugins/migrations/0004_add_distribution.py
```diff
@@ -0,0 +1,43 @@
...
+from pulp.server.db.migrations.lib import utils as migration_utils
+
+
+_logger = logging.getLogger(__name__)
+
+
+def migrate(*args, **kwargs):
+    """
+    Add a distribution field to both deb releases and components.
+    """
```
quba42: The DB migration simply copies the content of the existing field (containing the "codename") into the new `distribution` field (for both DebRelease and DebComponent).

This minimalist migration means that the behaviour of existing repositories will not change at all until they are next synced. (At which point the new "correct" behaviour will take effect.)

quba42:
After a first round of testing, I can confirm that this fixes the pulp issues #4705 and #4707 (as well as #4138 which describes the root cause).

I am still getting an error with issue #3464. I will continue to investigate (https://pulp.plan.io/issues/3464).

I also have yet to test the DB migration.

dkliban:
ok test

quba42:
It appears https://pulp.plan.io/issues/3464 is a separate issue within python-debpkgr. (As far as I can tell debpkgr may be tripping over the same issue of using some internal keys that aren't actually unique in all cases.)

That being said there are still several issues with the DB migration:
* It appears pulp is trying to use the new unit keys before running the migration, which does not work because the key fields need to be created by the migration. (Not sure how this is meant to work).
* Existing repositories do not update correctly with the first sync post migration. (New repositories work as expected).

common/pulp_deb/common/ids.py
quba42: It turns out this breaks the DB migration, since pulp-manage-db attempts to use the new keys before the migration has run. (This does not work since the distribution field does not yet exist at that time.)

Does anyone know how this sort of thing should be done? (We really should change the keys, since the old ones are not unique in all cases...)

quba42:
This is dependent on these changes within pulp itself: https://github.com/pulp/pulp/pull/3919

plugins/pulp_deb/plugins/migrations/0004_add_distribution.py
quba42: This is no longer true.

common/pulp_deb/common/ids.py
quba42: I solved this via a PR against pulp...
https://github.com/pulp/pulp/pull/3919

quba42:
Rebased and fixed merge conflict.

plugins/pulp_deb/plugins/migrations/0004_add_distribution.py
```diff
@@ -0,0 +1,134 @@
...
+def pre_index_migrate(*args, **kwargs):
+    """
+    Add a naive distribution field to both deb releases and components, to
+    ensure we will survive uniqueness constraints post index creation.
+    """
+    release_collection = connection.get_collection('units_deb_release')
+    component_collection = connection.get_collection('units_deb_component')
+
+    # Migrate the units_deb_release collection:
+    for release in release_collection.find({}).batch_size(100):
```
goosemania: @quba42, In my opinion, for pre_index_migrate especially, it's important to focus on idempotence.
There is no indication how well migration went and if pre-index part failed or not, so it's very possible that it will be re-run multiple times.
I don't know the typical size of the collection which you update, but in any case, I suggest to search for the documents without new field and update only those. So when re-run, there won't be any useless work done.
E.g. only documents without the `distribution` field and output is limited to the fields you need (I think `_id` is always there):
```
for release in release_collection.find({'distribution': {'$exists': False}},
                                       ['codename']).batch_size(100):
       ...
```

quba42: Only applying to units missing the distribution field makes sense.
I was kind of relying on the fact that these collections should not be that large, but of course I should not be assuming anything!

Syntax question: Is `.find({}, ['codename'])` the same as `.find({},{codename: true})` or does the latter only work in mongo shell?

goosemania: It seems like it's a bit different. See [pymongo docs for `find`](http://api.mongodb.com/python/current/api/pymongo/collection.html#pymongo.collection.Collection.find).
The `projection` is a parameter you are interested in. I understood it as if you use `list` -  `_id` is included, if you use `dict` - it's not (by default). An experiment will tell ;)

quba42:
Updated to reflect the latest changes in https://github.com/pulp/pulp/pull/3919!

plugins/pulp_deb/plugins/migrations/0004_add_distribution.py
quba42: Added the change, thanks for the suggestion.

quba42:
@mibanescu Since it now looks like the needed pulp changes will be accepted, it would be great if you could have a look at this.
We did add tests, and I am afraid the key changes are unavoidable (the old keys are part of the root cause of the problem). :wink: 

Note that some of the change is just renaming of internal fields for clarity (without really changing the logic).

quba42:
https://github.com/pulp/pulp/pull/3919 is now merged!

mibanescu:
Looks good go me.

PR #84, opened 2019-04-22
Update distributor.py

mcen1:
When attempting to sync a debian repo missing "all", an exception is generated with this line trying to remove "all". So I added an "if" to only remove if it's there.

### DIFF
--- a/plugins/pulp_deb/plugins/distributors/distributor.py
+++ b/plugins/pulp_deb/plugins/distributors/distributor.py
@@ -406,7 +406,8 @@ class MetadataStep(PluginStep):
                         release_meta_files.append(bz2_file_path)
 
             # Create the 'Release' file (for each release):
-            release_meta_data['architectures'].remove('all')
+            if 'all' in release_meta_data['architectures']:
+              release_meta_data['architectures'].remove('all')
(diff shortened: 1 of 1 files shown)


plugins/pulp_deb/plugins/distributors/distributor.py
```diff
@@ -406,7 +406,8 @@ def process_main(self, item=None):
                         release_meta_files.append(bz2_file_path)
 
             # Create the 'Release' file (for each release):
-            release_meta_data['architectures'].remove('all')
+            if 'all' in release_meta_data['architectures']:
+              release_meta_data['architectures'].remove('all')
```
quba42: I think this can also be solved by using `release_meta_data['architectures'].discard('all')` (`discard` instead of `remove`).

If my hunch is correct I would prefer this as a more lightweight alternative.
@mcen1 Can you test if that would also work for your problem?

mcen1: yup that's much better and cleaner, thank you!

quba42:
If you squish the two commits together, you have my ACK.
Thanks for the contribution!

PR #87, opened 2019-05-20
Single shot upload

mdellweg:


mdellweg:
@bmbouter See my take on the single-shot-upload.
In contrast to the rpm-plugin, the validation moved completely into the serializer (i know, we talked about those drf serializers earlier) and the task is in the right directory.

mdellweg:
Test still pending...

pulp_deb/app/serializers/content_serializers.py
```diff
@@ -0,0 +1,253 @@
...
+            ('release', 'component', 'architecture', 'relative_path')
+        model = InstallerFileIndex
+
+
+class PackageSerializer(SingleArtifactContentSerializer):
+    """
+    A Serializer for Package.
+    """
+
+    essential = YesNoField(
```
bmbouter: Does Debian just use 'yes' and 'no' instead?

pulp_deb/app/serializers/upload_serializers.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def validate(self, data):
+        """
+        Validate uploaded file, prepare artifact and collect shared resources.
+        """
+        data = super().validate(data)
+        data['filename'] = data['file'].name
+        artifact = Artifact.init_and_validate(data['file'])
+        try:
+            artifact.save()
```
bmbouter: I expected a lot of this to be in the viewset.  What parts of this method are validation? I would keep only those parts here. For instance the validation saves as a side-effect.

I wanted to share this observation; you're free to code it how you would like.

pulp_deb/app/tasks/upload.py
```diff
@@ -0,0 +1,63 @@
...
+log = logging.getLogger(__name__)
+
+
+def one_shot_upload(artifact, filename, repository=None):
+    """
+    Create a Package from an uploaded file and attach it to a Repository if provided.
+
+    Args:
+        artifact (Artifact): Create a Package from this artifact.
+    Optional:
```
bmbouter: Usually 1 line of whitespace separating sections.

pulp_deb/app/tasks/upload.py
```diff
@@ -0,0 +1,63 @@
...
+        package = Package.objects.get(sha256=artifact.sha256, relative_path=filename)
+    except ObjectDoesNotExist:
+        package_paragraph = debfile.DebFile(
+            fileobj=artifact.file
+        ).debcontrol()
+        package_dict = Package.from822(package_paragraph)
+        package_dict['relative_path'] = filename
+        package_dict['sha256'] = artifact.sha256
+        package = Package(**package_dict)
+        package.save()
```
bmbouter: Consider wrapping these various objects that are related into an `atomic with transaction():` statement.

pulp_deb/app/tasks/upload.py
```diff
@@ -0,0 +1,63 @@
...
+
+        resource = CreatedResource(content_object=package)
+        resource.save()
+
+    if repository:
+        content_to_add = Package.objects.filter(pk=package.pk)
+
+        # create new repo version with uploaded package
+        with RepositoryVersion.create(repository) as new_version:
+            new_version.add_content(content_to_add)
```
bmbouter: Don't you also want to make a `CreatedResource` for this too?

pulp_deb/app/viewsets.py
```diff
@@ -130,6 +131,51 @@ class InstallerPackageViewSet(ContentViewSet):
...
+    ViewSet for One Shot Deb Upload.
+
+    Args:
+        file@: package to upload
+    Optional:
+        repository: repository to update
+    """
+
+    serializer_class = serializers.OneShotUploadSerializer
+    # parser_classes = (MultiPartParser, FormParser)  # Why is that useful?
```
bmbouter: I don't know but maybe remove it instead of comment it?

pulp_deb/app/viewsets.py
```diff
@@ -130,6 +131,51 @@ class InstallerPackageViewSet(ContentViewSet):
...
+        serializer = serializers.OneShotUploadSerializer(
+            data=request.data, context={'request': request})
+        serializer.is_valid(raise_exception=True)
+        artifact = serializer.validated_data['artifact']
+        filename = serializer.validated_data['filename']
+        repository = serializer.validated_data.get('repository')
+
+        async_result = enqueue_with_reservation(
+            tasks.one_shot_upload,
+            serializer.shared_resources,
```
bmbouter: This is pretty opaque here. If I could see the data here I would understand the lock better. If `repository` is in use I think it needs to lock on the repository so that RepoVersion creation won't race causing one to fail. If `repository` is not in use I think the tasks can be 100% parallel and so a lock on the artifact ID would make sense there.

pulp_deb/app/serializers/content_serializers.py
mdellweg: Yes, it is `yes` or `no`.

http://www.sosst.sk/doc/debian-policy/policy.html/ch-controlfields.html

pulp_deb/app/tasks/upload.py
```diff
@@ -0,0 +1,63 @@
...
+
+from pulp_deb.app.models import (
+    Package,
+)
+
+
+log = logging.getLogger(__name__)
+
+
+def one_shot_upload(artifact, filename, repository=None):
```
bmbouter: Consider changing `artifact` to `artifact_id` and requerying the database for the object. The thinking is it keeps the Redis payload small and keeps the user's codepath using objects clearly initialized in the local runtime (versus rehydrated ones that were serialized in another Python interpreter).

Similarly with `repository` if repository isn't an ID but the actualy `Repository` object.

pulp_deb/app/serializers/upload_serializers.py
mdellweg: You are right, maybe saving the artifact can be pulled out again. But i think `init_and_validate` calls for inclusion in the validate step of the serializer.

pulp_deb/app/viewsets.py
```diff
@@ -130,6 +131,51 @@ class InstallerPackageViewSet(ContentViewSet):
...
+    Args:
+        file@: package to upload
+    Optional:
+        repository: repository to update
+    """
+
+    serializer_class = serializers.OneShotUploadSerializer
+    # parser_classes = (MultiPartParser, FormParser)  # Why is that useful?
+
+    @swagger_auto_schema(
```
bmbouter: :+1:

pulp_deb/tests/functional/api/test_upload.py
```diff
@@ -0,0 +1,76 @@
...
+
+from pulp_deb.tests.functional.constants import (
+    DEB_PACKAGE_PATH,
+    DEB_PACKAGE_URL,
+    DEB_SINGLE_REQUEST_UPLOAD_PATH,
+)
+from pulp_deb.tests.functional.utils import set_up_module as setUpModule  # noqa:F401
+
+
+class SingleRequestUploadTestCase(unittest.TestCase):
```
bmbouter: :100:

bmbouter: :100:

pulp_deb/app/tasks/upload.py
mdellweg: I thought, my observation was, that RepoVersion does that on its own.
(At some point, i had that version twice in the task_report.)

bmbouter: Oh yeah, you are right. Thanks.

pulp_deb/app/viewsets.py
mdellweg: Well, i did not want to keep it commented. Also it does not seem to be necessary. Still i want to understand why it is there in the rpm plugin.

bmbouter: I agree, maybe ask in #pulp-rpm or git-blame that line in RPM and ask them?

pulp_deb/app/tasks/upload.py
mdellweg: Good to know. I guess, that also means to clean up the rpm plugin.

pulp_deb/app/viewsets.py
mdellweg: My thinking here is that the serializer is the owner of all the logic, and the viewset only takes the serialized data to save objects & trigger tasks. Since the serializer already decides about the availability of the repository, it can also handle its inclusion in the locked resources. And if you liked it, we could follow that principle in all serializers, that are meant to prepare data for a task.

pulp_deb/app/tasks/upload.py
bmbouter: I didn't realize we did that there. I would recommend the same to the rpm team.

pulp_deb/app/tasks/upload.py
mdellweg: Also i want this to be part of the test. See TODO there.

pulp_deb/app/viewsets.py
mdellweg: https://github.com/pulp/pulp_rpm/commit/0e9c70dd46791d1d5d464b0c823b83920d2348e7#r33606083

pulp_deb/app/serializers/content_serializers.py
mdellweg: Anyway, this only lit up by refactoring the serializers into separate files. ;)

pulp_deb/app/viewsets.py
bmbouter: I think of the viewset as the place where the logic lives. The django view docs talk about how views contain logic in the first paragraph https://docs.djangoproject.com/en/2.2/topics/http/views/#writing-views

To me, serializers deal with translating python objects to a serialized form and vice-versa (which must include validation when hydrating the Python object). I wouldn't include handling logic in that place.

That being said you can organize pulp_deb however makes the most sense to you.

pulp_deb/app/serializers/content_serializers.py
bmbouter: Yup mainly just more education for me. No changes requested. :+1:

bmbouter:
@mdellweg lmk whenever I should retest this. Also I see it needs some conflict resolution too. I'm excited about this feature!

mdellweg:
Uups, i rebased on an old version of master.
Sadly, we are still waiting on the arfile fix in python-debian.

bmbouter:
I commented just now:  https://salsa.debian.org/python-debian-team/python-debian/merge_requests/11#note_79755

mdellweg:
The arfile patch got merged. ;)

pulp_deb/app/models.py
```diff
@@ -270,7 +270,7 @@ class Package(BasePackage):
 
     class Meta:
         default_related_name = "%(app_label)s_%(model_name)s"
-        unique_together = (("sha256",),)
+        unique_together = (("relative_path", "sha256"),)
```
bmbouter: I didn't expect the addition of `relative_path` here. It makes some sense, butI'm not sure at the model level was where the placement of content into a repo would be enforced.

pulp_deb/app/viewsets.py
```diff
@@ -120,6 +124,61 @@ class InstallerPackageViewSet(ContentViewSet):
     serializer_class = serializers.InstallerPackageSerializer
 
 
+class OneShotUploadViewSet(ViewSet):
```
bmbouter: Inheriting in this way will put the viewset outside of the /pulp/api/v3/ part of the API. I had done the same on pulp_ansible, but I'm going to switch it back here soon. I recommend inheriting from Pulp's `NamedViewSet` to bring this back into `pulp/api/v3`. Wdty?

pulp_deb/app/models.py
mdellweg: It is kind of a submarine change here, yes. Maybe it _is_ a partial fix for https://pulp.plan.io/issues/5377 .

mdellweg: And it _is_ needed in order to sync two repositories sharing a file with different relative_paths.

bmbouter: I see, thank you.

pulp_deb/app/tasks/upload.py
```diff
@@ -0,0 +1,61 @@
...
+    Create a Package from an uploaded file and attach it to a Repository if provided.
+
+    Args:
+        artifact_pk (str): Create a Package from this artifact.
+
+    Optional:
+        repository_pk (str): Insert the Package into this Repository.
+
+    """
+    log.info("{} {}".format(artifact_pk, repository_pk))
```
dkliban: i assume that the above is for debugging purposes only. maybe remove it?

pulp_deb/app/tasks/upload.py
```diff
@@ -0,0 +1,61 @@
...
+    artifact = Artifact.objects.get(pk=artifact_pk)
+    repository = Repository.objects.get(pk=repository_pk) if repository_pk else None
+    log.info(
+        _("Uploading deb package: artifact={artifact}, repository={repo}").format(
```
dkliban: I am not sure how useful it is to log this at INFO level.

mdellweg: DEBUG seems ok, right?

pulp_deb/app/viewsets.py
mdellweg: I do not find a `NamedViewSet`, only a `NamedModelViewSet`, which does not work here.
Also in a way, this url is available both with _and_ without the  `pulp/api/v3` part.
Therefore i would vote for not blocking this here and refactor it once there is a `NamedViewSet`.

pulp_deb/app/tasks/upload.py
dkliban: yes, debug is good.

pulp_deb/app/viewsets.py
bmbouter: Indeed I had meant `NamedModelViewSet`. Is the observation that it won't work here due to this ViewSet not having a model?

The trend I've seen more recently is that the one-shot uploader code is taking the place of the generic `create()` method on the Content viewset rather than having a separate viewset at all. What do you think about that?

Also you mention this is available with the `pulp/api/v3` currently? How does that work?

mdellweg: Yes, it is not having a model. And it complains about it.
I am not sure, i like completely replacing the create with upload logic. Can you still pass the relative_path parameter along an uploaded file? What about a file that is so big, it must be uploaded in chunks, assembled into an artifact, and then incorporated into a content unit?
If we can make the create serializer accept either an artifact or a file, i would like it.
But i would still merge this one, and start experimenting from there.
It seems to be accessible by both paths and i don't have the slightest idea, why.

bmbouter: I expected `relative_path` to be only used at repo-association time when the `ContentArtifact` instance is created. Is that what you think also? Is this upload also doing the repository association?

Regarding chunking, what you're describing sounds right. To restate it, a content unit creation would accept either a file delivered in that request OR a reference to an existing Artifact, which could have been uploaded with the chunked upload API. To help accomplish this the pulpcore-plugin could provide a serializer that provides exactly ^ generally and then it could be "re-used" as the serializer for `create()` specifically in pulp_deb and all plugins. How does this sound?

I would probably merge this now and rework later also. Definitely feel free to merge what you feel is right. You are ultimately it's developer so I defer to you.

mdellweg: Oh relative_path is used here like it is in the file repository (for the verbatim publisher). Everything else can be derived from the artifact. What you refer to, i think would be `_relative_path`. Also the (not-verbatim) publisher publishes the packages with its own rule for relative paths, ignoring this database field.

What you envision for the `create` is exactly matching my expectations. Especially the `pulpcore-plugin` helps implementing this part. ;)
But i really see that as the next step, therefore i am going to merge this one, as it gives some important functionality already.

PR #90, opened 2019-05-29
Fix the logging in migration 0004

quba42:
fixes #4884
https://pulp.plan.io/issues/4884


quba42:
I am calling it "WIP" until I have tested it.

quba42:
I successfully reproduced the bug, and then verified that it does not occur with this fix.

bherrin3:
Just adding a note that the re-test of 2-master in the broken state worked with QE ansible playbook retry.

PR #92, opened 2019-06-06
Fix migration 0004 for importers with empty config

quba42:
This fix prevents KeyErrors during migration 0004.
These can happen when there are importers with empty "config" dict.

fixes #4924
https://pulp.plan.io/issues/4924

### DIFF
--- a/plugins/pulp_deb/plugins/migrations/0004_add_distribution.py
+++ b/plugins/pulp_deb/plugins/migrations/0004_add_distribution.py
@@ -1,4 +1,5 @@
 import logging
+from six import string_types
 from pulp.server.db import connection
 from pulp_deb.common import ids
 from pymongo.errors import OperationFailure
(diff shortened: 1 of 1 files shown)


quba42:
So, I now went over every db field access and asked myself if I need to guard it against missing field (KeyError), None value, and/or empty string value.
As a result this has turned into a somewhat bigger/more confusing change.

I changed the status to "WIP:" to ensure I get a chance to retest everything before it is merged.
Once I have tested everything I will remove the WIP status.

plugins/pulp_deb/plugins/migrations/0004_add_distribution.py
```diff
@@ -62,24 +62,36 @@ def migrate(*args, **kwargs):
...
+    deb_importer_filter = {
+        'importer_type_id': ids.TYPE_ID_IMPORTER,
+        'config.releases': {'$exists': True},
+    }
 
     # Perform a best effort to recreate the upstream distribution:
     for repo_importer in importer_collection.find(deb_importer_filter).batch_size(100):
-        distributions = repo_importer['config']['releases'].split(',')
+        importer_releases = repo_importer['config']['releases']
+        if isinstance(importer_releases, basestring) and importer_releases:
```
quba42: Apparently python 3.x versions are not happy with this line.
Does anyone know how to make this compatible across the python 2.x vs. 3.x divide?

quba42: https://stackoverflow.com/questions/11301138/how-to-check-if-variable-is-string-with-python-2-and-3-compatibility

quba42: That should do it.

quba42:
I have now tested this in a variety of ways.
Since this should be considered a release blocker I recommend we merge it soon.

quba42:
I have now tested this in every way I can come up with, and it all seems to work as designed.

@mibanescu If we could merge this soon that would be great. That way only the packaging change of this fix will still be missing, and I can start trying to get in touch with the pulp_deb release team.

PR #93, opened 2019-06-11
Remove /usr/lib/pulp/plugins/types/deb.json from setup.py

quba42:
This legacy file will prevent migration 0004 from working.

WARNING: This will also require removing the file from the .spec file in
pulp-packaging repository!

fixes #4951
https://pulp.plan.io/issues/4951

I recommend we hold out on merging this until we have had some reaction from the pulp-packaging maintainers.


mdellweg:
I think, we can complete remove the file from the repository, not just not installing it.

quba42:
I also added a PR to pulp-packaging:
https://github.com/pulp/pulp-packaging/pull/97

quba42:
Updated to reflect @mdellweg's suggestion.

dkliban:
ok test

PR #94, opened 2019-06-13
Repair test_publish_repo

quba42:
Some assertions were silently not being called.
Added extra assertions to ensure this cannot happen in the future.

### DIFF
--- a/plugins/test/unit/plugins/distributors/test_distributor.py
+++ b/plugins/test/unit/plugins/distributors/test_distributor.py
@@ -250,6 +250,8 @@ class PublishRepoMixIn(object):
                 id='allid',
                 distribution='default',
             ))
+        release_file_counter = 0
+        packages_file_counter = 0
         for release in release_units:
             comp_dir = os.path.join(
                 repo_config['http_publish_dir'],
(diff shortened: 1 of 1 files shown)


plugins/test/unit/plugins/distributors/test_distributor.py
```diff
@@ -258,17 +260,22 @@ def mock_get_units(repo_id, model_class, *args, **kwargs):
                 release.distribution)
             release_file = os.path.join(comp_dir, 'Release')
             self.assertTrue(os.path.exists(release_file))
+            release_file_counter += 1
             # Make sure the components Packages files exist
             for comp in [comp.plain_component for comp in component_units
-                         if comp.release == release.codename]:
+                         if comp.distribution == release.distribution]:
```
quba42: This is what was actually broken. Because of this the list comprehension returned an empty list, so the `for` loop was never entered, rendering all assertions within it useless.

I have  now added counters to all relevant tests to ensure that the loop is traversed the expected number of times for each test.

mibanescu: Looks good to me. You can merge this, unless you want me to.

quba42:
I was just about to say that while I don't have merge rights, I can have @mdellweg do it...

PR #95, opened 2019-06-13
Use dictionaries for units_to_check

mdellweg:
Since lookup in dicts is logarithmic, this should improve performance
significantly if a large number of old content units are to be checked
for removal (i.e. when switching a huge repository to another
architecture).


mdellweg:
@mibanescu as this is pulp2, may i ask you to have a look?

PR #96, opened 2019-06-28
Katello support for migration 0004

quba42:
I am adding this PR so I can ask if this could still be included (release schedule wise), and whether we would want a Katello specific work around in pulp.

While non essential, this change will greatly improve the user experience of Katello pulp_deb users and is highly unlikely to affect anyone else.
ATIX will almost certainly add this change (or something feature equivalent) to the pulp instance we package with our orcharhinos.

In addition to the Katello work around, I have also improved the logging. This should be particularly valuable for analysis if the migration produces unexpected results.

My apologies for springing this on everyone after the dev freeze, but I wanted to at least put it up for discussion.

### DIFF
--- a/plugins/pulp_deb/plugins/migrations/0004_add_distribution.py
+++ b/plugins/pulp_deb/plugins/migrations/0004_add_distribution.py
@@ -1,4 +1,5 @@
 import logging
+import re
 from six import string_types
 from pulp.server.db import connection
 from pulp_deb.common import ids
@@ -63,10 +64,62 @@ def migrate(*args, **kwargs):
         _logger.info("Ignoring expected OperationFailure exception:")
         _logger.info(str(exception))
 
+    # Copy Katello repo_importer releases fields from the relevant root repositories:
+    # The following for loop is exclusive to Katello's pulp usage. Anyone not using
+    # pulp_deb as part of Katello should be completely unaffected.
+    deb_importer_filter = {
+        'importer_type_id': ids.TYPE_ID_IMPORTER,
+        'repo_id': {'$exists': True},
+        'config.releases': {'$exists': False},
+    }
+
+    pattern = re.compile('^.+-([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})$')
+
+    for repo_importer in importer_collection.find(deb_importer_filter).batch_size(100):
+        repo_id = repo_importer['repo_id']
+
+        if not isinstance(repo_id, string_types):
+            continue
+
+        match = re.match(pattern, repo_id)
+        if match is not None:
+            root_repo_id_candidate = match.group(1)
+        else:
+            continue
+
+        id_filter = {
+            'importer_type_id': ids.TYPE_ID_IMPORTER,
+            'repo_id': root_repo_id_candidate,
+            'config.releases': {'$exists': True},
+        }
+
+        projection = {
+            'config': True,
+        }
+
+        root_repo_importer_candidates = importer_collection.find(id_filter, projection)
+
+        if root_repo_importer_candidates.count() == 1:
+            katello_root_repo_importer = list(root_repo_importer_candidates)[0]
+        else:
+            continue
+
+        root_repo_releases = katello_root_repo_importer['config']['releases']
+
+        msg = 'Adding config.releases from root repo_importer {} to repo_importer {}.'\
+              ''.format(katello_root_repo_importer['_id'], repo_importer['_id'])
+        _logger.debug(msg)
+        importer_collection.update_one(
+            {'_id': repo_importer['_id']},
+            {'$set': {'config.releases': root_repo_releases}},
+        )
+
     deb_importer_filter = {
         'importer_type_id': ids.TYPE_ID_IMPORTER,
         'config.releases': {'$exists': True},
     }
+
+    changed_repos = set()
 
     # Perform a best effort to recreate the upstream distribution:
     for repo_importer in importer_collection.find(deb_importer_filter).batch_size(100):
@@ -76,9 +129,9 @@ def migrate(*args, **kwargs):
         else:
             continue
 
-        release_filter = {'repoid': repo_importer['repo_id']}
+        repo_id = repo_importer['repo_id']
 
-        for release in release_collection.find(release_filter).batch_size(100):
+        for release in release_collection.find({'repoid': repo_id}).batch_size(100):
             original_distribution = None
             suite = ''
             codename = release['codename']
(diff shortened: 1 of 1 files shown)


dkliban:
ok test

dkliban:
@mibanescu Please take a look at this PR.

plugins/pulp_deb/plugins/migrations/0004_add_distribution.py
```diff
@@ -63,11 +63,55 @@ def migrate(*args, **kwargs):
...
+        'importer_type_id': ids.TYPE_ID_IMPORTER,
+        'repo_id': {'$exists': True},
+        'config.releases': {'$exists': False},
+    }
+
+    for repo_importer in importer_collection.find(deb_importer_filter).batch_size(100):
+        repo_id = repo_importer['repo_id']
+
+        if isinstance(repo_id, string_types) and len(repo_id) > 36:
+            root_repo_id_candidate = repo_id[-36:]
```
quba42: I could still adjust this to use a regex match even more specific to the Katello naming scheme.
That would lessen the risk of this part of the migration affecting anything other than the intended Katello use case to something very close to zero.

plugins/pulp_deb/plugins/migrations/0004_add_distribution.py
```diff
@@ -139,16 +191,39 @@ def migrate(*args, **kwargs):
...
+            changed_repos.add(repo_id)
+
+    output_file_path = '/var/lib/pulp/deb_repo_republish_list.txt'
+
+    _logger.info('List of repos modified by the best effort migration (by repo_id):')
+
+    with open(output_file_path, 'w') as output_file:
+        for repo_id in changed_repos:
+            output_file.write('{}\n'.format(repo_id))
+            _logger.info(repo_id)
```
quba42: I could use some feedback on whether leaving behind a file like this (in addition to the logging) is desirable/best practice/etc.

My rational is that this would make it very easy to provide some post migration script to republish all affected repositories (which is the only way to ensure that these changes go fully into effect).

plugins/pulp_deb/plugins/migrations/0004_add_distribution.py
quba42: Done.

plugins/pulp_deb/plugins/migrations/0004_add_distribution.py
```diff
@@ -139,16 +198,39 @@ def migrate(*args, **kwargs):
...
             release_collection.update_one(
                 {'_id': release['_id']},
-                {'$set': new_release_fields},
+                {'$set': {'distribution': original_distribution}},
             )
+            changed_repos.add(repo_id)
+
+    output_file_path = '/var/lib/pulp/deb_repo_republish_list.txt'
+
+    _logger.info('List of repos modified by the best effort migration (by repo_id):')
```
mdellweg: Can you only output this list (with its heading) if it is nonempty?

plugins/pulp_deb/plugins/migrations/0004_add_distribution.py
```diff
@@ -112,15 +165,21 @@ def migrate(*args, **kwargs):
...
                         {'$set': new_component_fields},
                     )
                 elif original_distribution:
-                    new_component_fields = {'distribution': original_distribution}
+                    msg = 'Updating distribution of component {} from {} to {}.'.format(
+                        component['_id'],
+                        component['distribution'],
+                        original_distribution,
+                    )
+                    _logger.info(msg)
```
mdellweg: This seems to be a bit verbose. Can you restrict it to the log level debug?

plugins/pulp_deb/plugins/migrations/0004_add_distribution.py
quba42: Done.

plugins/pulp_deb/plugins/migrations/0004_add_distribution.py
quba42: Done.

quba42:
I have now incorporated some feedback by @mdellweg and rebased.

plugins/pulp_deb/plugins/migrations/0004_add_distribution.py
```diff
@@ -54,20 +55,70 @@ def migrate(*args, **kwargs):
...
 
     try:
         component_collection.drop_index("name_1_release_1_repoid_1")
     except OperationFailure as exception:
-        _logger.info("Ignoring expected OperationFailure exception:")
-        _logger.info(str(exception))
+        _logger.debug("Ignoring expected OperationFailure exception:")
+        _logger.debug(str(exception))
+
+    # Copy Katello repo_importer releases fields from the relevant root repositories:
```
mdellweg: You could be a bit more specific, that this is meant to only affect katello installations.

plugins/pulp_deb/plugins/migrations/0004_add_distribution.py
```diff
@@ -54,20 +55,70 @@ def migrate(*args, **kwargs):
...
+        _logger.debug("Ignoring expected OperationFailure exception:")
+        _logger.debug(str(exception))
 
     try:
         component_collection.drop_index("name_1_release_1_repoid_1")
     except OperationFailure as exception:
-        _logger.info("Ignoring expected OperationFailure exception:")
-        _logger.info(str(exception))
+        _logger.debug("Ignoring expected OperationFailure exception:")
+        _logger.debug(str(exception))
```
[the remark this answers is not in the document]
quba42: Done.

plugins/pulp_deb/plugins/migrations/0004_add_distribution.py
quba42: Added two more lines of comments for clarification.

plugins/pulp_deb/plugins/migrations/0004_add_distribution.py
```diff
@@ -139,16 +200,40 @@ def migrate(*args, **kwargs):
...
+            )
+            _logger.debug(msg)
             release_collection.update_one(
                 {'_id': release['_id']},
-                {'$set': new_release_fields},
+                {'$set': {'distribution': original_distribution}},
             )
+            changed_repos.add(repo_id)
+
+    output_file_path = '/var/lib/pulp/deb_repo_republish_list.txt'
```
quba42: My one remaining question: Is it ok to write an additional output file (in addition to the logging) as part of this migration?
Is there a better place for it than `/var/lib/pulp/deb_repo_republish_list.txt`?

quba42: Looking at this example (https://github.com/pulp/pulp_puppet/blob/2-master/pulp_puppet_plugins/pulp_puppet/plugins/migrations/0005_puppet_module_name_change.py#L54, thanks @goosemania) I have included the migration number in the output file name.

mibanescu:
Feel free to merge it whenever.

mdellweg:
Thanks to all!

PR #99, opened 2019-07-18
Make pulp_deb publish empty repositories.

quba42:
fixes #4345
https://pulp.plan.io/issues/4345

The current behavior of pulp_deb is not to publish anything if repositories are empty (contain no packages).

This prevents pulp to pulp syncs and may be a problem for clients configured to use a repository with an empty publish directory.

By simply removing a guard the publish mechanism works as expected producing a valid empty repository complete with Release and (empty) Packages file.

The only limitation is that because of https://pulp.plan.io/issues/4094 the only architecture that will be published is the `all` architecture. However, this can be fixed as a separate issue.

### DIFF
--- a/plugins/test/unit/plugins/distributors/test_distributor.py
+++ b/plugins/test/unit/plugins/distributors/test_distributor.py
@@ -538,6 +538,34 @@ class TestPublishRepoMultiCompArchDeb(PublishRepoMixIn, BaseTest):
     packages_file_count = 12
 
 
+class TestPublishEmptyRepo(PublishRepoMixIn, BaseTest):
+    """
+    Empty repositories (no packages) should still publish a valid Release file
+    allong with empty Packages files.
+    """
+    Sample_Units = {
+        models.DebPackage: [],
+        models.DebComponent: [
+            dict(
+                name='main',
+                distribution='stable',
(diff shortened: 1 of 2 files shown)


plugins/test/unit/plugins/distributors/test_distributor.py
```diff
@@ -538,6 +538,34 @@ class TestPublishRepoMultiCompArchDeb(PublishRepoMixIn, BaseTest):
...
+    allong with empty Packages files.
+    """
+    Sample_Units = {
+        models.DebPackage: [],
+        models.DebComponent: [
+            dict(
+                name='main',
+                distribution='stable',
+                id='mainid',
+                packages=['bbbb', 'cccc', 'dddd', 'ffff']),
```
sbernhard: In case of a empty models.DebPackage[], shouldn't be "packages=[....]" empty, too?

quba42: Probably. Or maybe the line should be removed all together. (In the sense that I do not want to mock anything I don't have to.) I will check it out.

quba42: Fixed it.

PR #100, opened 2019-07-23
Add validation of existing packages to sync

mdellweg:
If force full mode is active, existing packages are validated and if needed
rescheduled for download.

closes #5170
https://pulp.plan.io/issues/5170

### DIFF
--- a/plugins/pulp_deb/plugins/importers/sync.py
+++ b/plugins/pulp_deb/plugins/importers/sync.py
@@ -33,6 +33,19 @@ DEBSYNC002 = Error(
     ["repo_id", "feed_url", "filename", "checksum_expected", "checksum_actual"])
 
 
+def verify_unit_callback(unit):
+    if not os.path.isfile(unit.storage_path):
+        return False
+    checksums = unit.calculate_deb_checksums(unit.storage_path)
+    if unit.sha1 and unit.sha1 != checksums['sha1']:
+        return False
+    if unit.md5sum and unit.md5sum != checksums['md5sum']:
+        return False
+    if unit.sha256 and unit.sha256 != checksums['sha256']:
+        return False
+    return True
+
+
 class RepoSync(publish_step.PluginStep):
     Type_Class_Map = {
         models.DebPackage.TYPE_ID: models.DebPackage,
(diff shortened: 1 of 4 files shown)


mdellweg:
Tests _will_ follow.

plugins/pulp_deb/plugins/importers/sync.py
```diff
@@ -59,6 +73,8 @@ def __init__(self, repo, conduit, config):
         self.components = split_or_none(self.get_config().get('components'))
         self.remove_missing = self.get_config().get_boolean(
             constants.CONFIG_REMOVE_MISSING_UNITS, constants.CONFIG_REMOVE_MISSING_UNITS_DEFAULT)
+        self.validate = repo_controller.check_perform_full_sync(
+            self.repo.id, self.conduit, self.config)
```
mdellweg: I just realized by looking at how the tests fail, that this is _not_ the option i was looking for...

plugins/test/unit/plugins/importers/test_sync.py
```diff
@@ -230,7 +230,7 @@ def test_SaveDownloadedUnits(self):
 
         repo = self.repo.repo_obj
         for path, unit in path_to_unit.items():
-            unit.save_and_associate.assert_called_once_with(path, repo)
+            unit.save_and_associate.assert_called_once_with(path, repo, force=False)
```
quba42: Shouldn't this default to `False` on its own?

plugins/pulp_deb/plugins/importers/sync.py
```diff
@@ -59,6 +73,7 @@ def __init__(self, repo, conduit, config):
         self.components = split_or_none(self.get_config().get('components'))
         self.remove_missing = self.get_config().get_boolean(
             constants.CONFIG_REMOVE_MISSING_UNITS, constants.CONFIG_REMOVE_MISSING_UNITS_DEFAULT)
+        self.validate = self.get_config().get_boolean(importer_constants.KEY_VALIDATE, False)
```
quba42: If I understand this correctly we are now deciding between verifying or not verifying based on an importer setting?
How does this compare to how RPM and others are handling advanced sync options?

plugins/test/unit/plugins/importers/test_sync.py
mdellweg: Maybe, but the function will actually be called with an explicit `False` in the new version.

mdellweg:
The test are failing due to https://github.com/pulp/pulp/pull/3946

m-bucher:
Are we ready to merge this?
We also have a PR in katello waiting for this https://github.com/Katello/katello/pull/8239

mdellweg:
I promised tests. Added a test to check the unit validation code.
I think this should be considered ready.

mdellweg:
@mibanescu as this is pulp2, may i ask you to have a look?

PR #101, opened 2019-08-07
Makes plugin compatible with pulpcore-plugin 0.1.0rc4

dkliban:
This PR contains multiple commits that were needed to get this plugin compatible with pulpcore-plugin 0.1.0rc4. 

This PR also introduces the template_config.yml that is needed to manage the Travis config with plugin-template.


.travis/check_commit.sh
```diff
@@ -0,0 +1,36 @@
...
+  exit 0
+fi
+
+if [ "$TRAVIS_COMMIT_RANGE" != "" ]; then
+  RANGE=$TRAVIS_COMMIT_RANGE
+elif [ "$TRAVIS_COMMIT" != "" ]; then
+  RANGE=$TRAVIS_COMMIT
+fi
+
+# Travis sends the ranges with 3 dots. Git only wants one.
```
mdellweg:
```suggestion
# Travis sends the ranges with 3 dots. Git only wants two.
```

mdellweg: Alternatively: "Git only knows why."

dkliban: lol

.travis/install.sh
```diff
@@ -1,36 +1,23 @@
...
-pip install -r test_requirements.txt
+# WARNING: DO NOT EDIT!
+#
+# This file was generated by plugin_template, and is managed by bootstrap.py. Please use
+# bootstrap.py to update this file.
+#
+# For more info visit https://github.com/pulp/plugin_template
 
-pushd ..
+set -v
```
mdellweg: I think, this should at least be `set -e` here and elsewhere.
I prefer `set -euv`.

.travis/before_script.sh
```diff
@@ -1,14 +1,26 @@
...
+#!/usr/bin/env sh
+
+# WARNING: DO NOT EDIT!
+#
+# This file was generated by plugin_template, and is managed by bootstrap.py. Please use
+# bootstrap.py to update this file.
+#
+# For more info visit https://github.com/pulp/plugin_template
+
+set -v
```
mdellweg: There's another one

PR #114, opened 2019-09-05
Problem: Travis does not test building docs

dkliban:
Solution: enable docs testing template_config.yaml

This patch adds the latest plugin template changes in addition to the documentation
building on Travis.

The latest changes make the check for services being started shorter. The logs for
pulp-api are also longer in case of test failures.

[noissue]


CHANGES.rst
```diff
@@ -13,3 +13,39 @@ Changelog
 
 .. towncrier release notes start
 
+2.0.0a2.dev1 (2019-09-06)
```
mdellweg: That should be the new version, i think.

CHANGES.rst
```diff
@@ -13,3 +13,39 @@ Changelog
...
+- Reference the fact you must have both _relative_path and relative_path.
+  `#5376 <https://pulp.plan.io/issues/5376>`_
+- Fix various documentation issues from API changes, plus other misc fixes.
+  `#5380 <https://pulp.plan.io/issues/5380>`_
+
+
+Misc
+----
+
+- `#4681 <https://pulp.plan.io/issues/4681>`_, `#5401 <https://pulp.plan.io/issues/5401>`_
```
mdellweg: I think, the misc entries are important enough to be spelled out. We probably need some configuration for this.

See https://github.com/hawkowl/towncrier/blob/master/pyproject.toml#L7

docs/_templates/restapi.html
```diff
@@ -0,0 +1,24 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>REST API for Pulp 3 RPM Plugin</title>
```
mdellweg:
```suggestion
    <title>REST API for Pulp 3 Deb Plugin</title>
```

docs/Makefile
```diff
@@ -42,6 +42,7 @@ clean:
 	-rm -rf $(BUILDDIR)/*
 
 html:
+	curl -o _static/api.json "http://localhost:24817/pulp/api/v3/docs/api.json?plugin=pulp_rpm"
```
mdellweg:
```suggestion
	curl -o _static/api.json "http://localhost:24817/pulp/api/v3/docs/api.json?plugin=pulp_deb"
```

CHANGES.rst
mdellweg: We could also change the template and ignore the `showcontent` parameter.

dkliban: I've added both of these changes to my branch.

CHANGES.rst
dkliban: I've rebased and re-generated the change log.

docs/_templates/restapi.html
dkliban: Thank you!

PR #118, opened 2019-09-19
Switch to using PublishedMetada.create_from_file()

dkliban:
Required PR: pulp/pulpcore#303

re: #5304
https://pulp.plan.io/issues/5304

### DIFF
--- a/CHANGES/5304.misc
+++ b/CHANGES/5304.misc
@@ -1,0 +1,1 @@
+PublishedMetadata is now a type of Content.
--- a/pulp_deb/app/tasks/publishing.py
+++ b/pulp_deb/app/tasks/publishing.py
@@ -111,16 +111,12 @@ def publish(repository_version_pk, simple=False, structured=False):
                     _add_to_release(release, package_index_path)
                     _add_to_release(release, gz_package_index_path)
 
-                    package_index = PublishedMetadata(
-                        relative_path=package_index_path,
-                        publication=publication,
-                        file=File(open(package_index_path, "rb")),
+                    package_index = PublishedMetadata.create_from_file(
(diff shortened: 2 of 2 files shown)


CHANGES/5304.misc
```diff
@@ -0,0 +1,4 @@
+Required PR: https://github.com/pulp/pulpcore/pull/303/
+
+re: #5304
+https://pulp.plan.io/issues/5304
```
mdellweg: this sounds like the information to go in the commit message.

pulp_deb/app/tasks/publishing.py
```diff
@@ -111,27 +111,21 @@ def publish(repository_version_pk, simple=False, structured=False):
                     _add_to_release(release, package_index_path)
                     _add_to_release(release, gz_package_index_path)
 
-                    package_index = PublishedMetadata(
-                        relative_path=package_index_path,
-                        publication=publication,
-                        file=File(open(package_index_path, "rb")),
+                    package_index = PublishedMetadata.create_from_file(
+                        publication=publication, file=File(open(package_index_path, "rb"))
```
mdellweg: How is the relative_path determined in this PublishedMetadata?

dkliban: If the relative_path is passed in to create_from_file, the that value is used. If it is not passed in, the file name from the file object is used. 

https://docs.pulpproject.org/en/pulpcore-plugin/nightly/api-reference/models.html#pulpcore.plugin.models.PublishedMetadata.create_from_file

CHANGES/5304.misc
dkliban: yep ... fixed it now.

PR #120, opened 2019-09-23
Rename pulp services

lubosmj:
closes #4554
https://pulp.plan.io/issues/4554


CHANGES/4554.misc
```diff
@@ -0,0 +1 @@
+Change the prefix of Pulp services from pulp-* to pulpcore-*
```
mdellweg: I think, this should be `4554.doc`

PR #123, opened 2019-10-02
Bump version

mdellweg:
Generate changelog


pulp_deb/__init__.py
```diff
@@ -1,3 +1,3 @@
-__version__ = "2.0.0b2-dev"
```
goosemania: It should be 2.0.0b2.dev according to PEP 440

goosemania: I would fix it in setup.py as well. It will be an issue for bindings otherwise.

setup.py
```diff
@@ -2,11 +2,11 @@
 
 from setuptools import find_packages, setup
 
-requirements = ["pulpcore-plugin~=0.1rc4", "python-debian>=0.1.36"]
+requirements = ["pulpcore-plugin~=0.1rc6", "python-debian>=0.1.36"]
```
goosemania: It's up to you, however in pulp_rpm we pinned this time to ==0.1.0rc6 because there will be more breaking changes.
It's only for release, master branch should be ~=0.1rc6.

pulp_deb/__init__.py
mdellweg: OK, then the old one was wrong. I will adhere to that in the first commit after the release. ;)

PR #126, opened 2019-10-04
Fix docs due to renaming of pulpcore-content-app to pulpcore-content

goosemania:
[noissue]

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -56,7 +56,7 @@ Run Services
 
 .. code-block:: bash
 
-   sudo systemctl restart pulpcore-content-app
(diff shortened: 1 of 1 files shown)


mdellweg:
Approved the change. But a changelog entry seems to me missing.

goosemania:
Oh, I didn't realise that you've released the other docs change for the same issue. Ok, I'll add a changelog entry.

mdellweg:
Ahh, that's the problem. Then a [noissue] should be good enough, but you must amend the commit with it.

mdellweg:
Thank you!

PR #137, opened 2019-11-12
Typed repos

mdellweg:


dev_requirements.txt
```diff
@@ -8,6 +8,3 @@ flake8-docstrings
 
 flake8-tuple
 flake8-quotes
```
dkliban: you need to add 'requests' to the list so that the script that checks with pulp.plan.io about issues can work.

mdellweg: I don't think, that was the problem. The ticket i referenced to (as all the other plugins) was already in the modified state.
Also this file is freshly generated by the plugin_template. I believe that to work.

PR #146, opened 2020-01-27
Publish components

mdellweg:


pulp_deb/app/tasks/publishing.py
```diff
@@ -68,101 +75,158 @@ def publish(repository_version_pk, simple=False, structured=False):
...
-                release["SHA256"] = []
-                release["SHA512"] = []
-                package_index_files = {}
+                codename = "default"
+                component_name = "all"
+                architectures = (
+                    Package.objects.filter(pk__in=repo_version.content.order_by("-pulp_created"),)
+                    .distinct("architecture")
+                    .values_list("architecture", flat=True)
+                )
```
quba42: I think there are situations, where Users might want to publish certain architectures even if there are no packages specific to that architecture. Are we covering such cases with this approach?

mdellweg: Not with simple, but with the structured approach, we publish every architecture that is represented via a ReleaseArchitecture unit.

pulp_deb/app/tasks/publishing.py
```diff
@@ -68,101 +75,158 @@ def publish(repository_version_pk, simple=False, structured=False):
...
+            {"sha512": artifact.sha512, "size": artifact.size, "name": metadata.relative_path}
+        )
+
+    def finish(self):
+        # Publish Packages files
+        for component in self.components.values():
+            component.finish()
+        # Publish Release file
+        self.release["components"] = " ".join(self.components.keys())
+        release_path = os.path.join("dists", self.release["codename"], "Release")
```
quba42: This should use the `distribution` instead of the `codename`.

PR #147, opened 2020-01-31
Fix synching

mdellweg:
Do not attempt to 'download' generated temporary files. Create artifacts
directly instead.

[noissue]


bmbouter:
This looks right to me.

PR #149, opened 2020-02-10
Update pulpcore requirement

fabricio-aguiar:
[noissue]

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,7 @@
 
 from setuptools import find_packages, setup
 
-requirements = ["pulpcore>=3.1.0.dev,<3.2", "python-debian>=0.1.36"]
+requirements = ["pulpcore>=3.1.0.dev,<3.3", "python-debian>=0.1.36"]
 
 setup(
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -2,7 +2,7 @@
 
 from setuptools import find_packages, setup
 
-requirements = ["pulpcore>=3.1.0.dev,<3.2", "python-debian>=0.1.36"]
+requirements = ["pulpcore>=3.1.0.dev,<3.3", "python-debian>=0.1.36"]
```
mdellweg: So `3.2.0.dev0` is actually not less than `3.2`?
There is something rotten in the state of python versions.

fabricio-aguiar: apparently not, on pulp-dev IRC:
```
<quba42> When I try to provision my source box using pulplift, I get "kg_resources.VersionConflict: (pulpcore 3.2.0.dev0 (/home/vagrant/devel/pulpcore), Requirement.parse('pulpcore<3.2,>=3.1.0.dev'))"
```

PR #154, opened 2020-02-21
Add Release signing to publishing

mdellweg:
Required PR: pulp/pulpcore#558

fixes #6171


mdellweg:
@quba42 @dkliban

pulp_deb/app/tasks/publishing.py
```diff
@@ -218,15 +229,27 @@ def finish(self):
...
         with open(release_path, "wb") as release_file:
             self.release.dump(release_file)
         release_metadata = PublishedMetadata.create_from_file(
             publication=self.publication, file=File(open(release_path, "rb")),
         )
         release_metadata.save()
-        # TODO Sign release
+        if self.signing_service:
+            in_release_path = os.path.join(release_dir, "InRelease")
+            release_gpg_path = os.path.join(release_dir, "Release.gpg")
```
dkliban: it's a little confusing that this is the name of the signature produced by the signing service. I imagined you would capture the output of self.signing_service.sign() and use the values from there. What do you think about refactoring it that way?

mdellweg: Hard to say. This signing service is supposed to produce two files I know the exact names of. The interface is only meant to report one.

dkliban: It's up to you. Maybe provide a comment saying that the service produces that file?

mdellweg: I even thought about subclassing another service, now that I have proof of the concept.

quba42: https://github.com/pulp/pulp_deb/pull/160 is a evolution of this PR, that uses a dedicated "AptReleaseSigningService".

quba42:
Since we now have the dedicated AptReleaseSigningService version merged, I guess we can close this PR?

mdellweg:
Right.

PR #98, opened 2019-07-02
Update version to 1.11 alpha

pcreech:

PR #155, opened 2020-02-24
1.10.1 cps

evgeni:


quba42:
If you could also cherry-pick c5c0647d4a3f4954345809ef30e193f15f046e52 that would be good.
(That commit isn't associated with any issue...)

Note: With the addition of this commit the only difference between `1.10-release` and `master` would be that `master` contains some more tests.
Question: What would need to happen to create an `1.11-release` branch at the current tip of master instead of this merge?

quba42:
Were we meant to have merged this (https://github.com/pulp/pulp_deb/pull/98) at some point?
P.S: Please excuse my great ignorance regarding the pulp2 release process...

evgeni:
@ipanova is @quba42's request reasonable? the code looks so, but I don't know if you need anything in terms of bookkeeping/release notes.

ipanova:
Sounds good, lets pick that one as well.Release notes are usually kept by
plugins, I think debian doesn't have those, or at least I have not find any.

evgeni:
Cool. Picked!

PR #160, opened 2020-03-13
Add Release signing using AptReleaseSigningService

quba42:
Based on the POC by @mdellweg in https://github.com/pulp/pulp_deb/pull/154.
(I am literally reusing the same commit with minor adjustments).

In addition this PR adds a dedicated AptReleaseSigningService instead of (mis-)using the AsciiArmoredDetachedSigningService from pulpcore.


pulp_deb/app/models/signing_service.py
```diff
@@ -0,0 +1,166 @@
...
+        """
+        with tempfile.TemporaryDirectory() as temp_directory_name:
+            test_release_path = os.path.join(temp_directory_name, "Release")
+            with open(test_release_path, "wb") as test_file:
+                test_data = b"arbitrary data"
+                test_file.write(test_data)
+                test_file.flush()
+                try:
+                    return_value = self.sign(test_release_path)
+                except JSONDecodeError:
```
mdellweg: Can we move that to the `sign` function?

pulp_deb/app/tasks/publishing.py
```diff
@@ -218,15 +229,26 @@ def finish(self):
...
         release_metadata = PublishedMetadata.create_from_file(
             publication=self.publication, file=File(open(release_path, "rb")),
         )
         release_metadata.save()
-        # TODO Sign release
+        if self.signing_service:
+            signed = self.signing_service.sign(release_path)
+            for signature_file in signed["signatures"].values():
+                file_name = os.path.basename(signature_file)
+                relative_path = os.path.join(release_dir, file_name)
```
mdellweg: I think, we don't need to specify this `relative_path` as it will be derived from the `File` object.

quba42: Unless the signing service places the signature files in a different folder from the Release file to be signed, which is currently possible.

pulp_deb/app/models/signing_service.py
quba42: Done.

pulp_deb/app/tasks/publishing.py
mdellweg: And could be a requirement if the script needs to run as another user...

Can we make the testing script actually do this then?

quba42: That makes sense, will do.

quba42: Done.

PR #162, opened 2020-03-19
Bindings installation on dev environment

fabricio-aguiar:
https://pulp.plan.io/issues/6221
ref #6221

### DIFF
--- a/CHANGES/6396.doc
+++ b/CHANGES/6396.doc
@@ -1,0 +1,1 @@
+Documented bindings installation on dev environment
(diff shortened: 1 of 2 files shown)


fabricio-aguiar:
> I guess this is this coming from plugin_template?

it did not come from plugin_template, I just thought it would be nice to have this doc matching with pulpcore: https://github.com/pulp/pulpcore/pull/598

mdellweg:
So should it be in plugin_template, too?

mdellweg:
Thank you @fabricio-aguiar !

fabricio-aguiar:
> So should it be in plugin_template, too?

good point! I think it could be valuable

PR #163, opened 2020-03-23
Permit empty Suite fields

quba42:
fixes #6050, #6383

Adds test for both.


pulp_deb/app/tasks/synchronizing.py
```diff
@@ -255,7 +255,9 @@ def __init__(self, remote, *args, **kwargs):
                     release_file.sha256 = release_file_artifact.sha256
                     release_file_dict = deb822.Release(release_file_artifact.file)
                     release_file.codename = release_file_dict["Codename"]
-                    release_file.suite = release_file_dict["Suite"]
+                    suite = release_file_dict.get("Suite")
+                    if suite:
+                        release_file.suite = suite
```
mdellweg:
```suggestion
                    if "suite" in release_file_dict:
                        release_file.suite = release_file_dict.get("Suite")
```

PR #165, opened 2020-04-29
Prepare 2.3.0b1 (beta) release

quba42:
I have prepared this PR for creating a `pulp_deb` `2.3.0b1` beta release, compatible to pulpcore `2.3`.

I believe I do not have the required permissions to actually create this beta release, which is why this is a "Draft pull request".

If I have understood everything correctly, I need someone to do the following in order to create the Release:

1. Review these changes
1. Create a branch named `2.3` at commit 662daab and push it to `pulp/pulp_deb`
1. Create a tag named `2.3.0b1` at the same commit and push it to `pulp/pulp_deb`
1. Check if the pipeline successfully released to https://pypi.org/project/pulp-deb/
1. Merge this PR

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,13 +3,13 @@
 from setuptools import find_packages, setup
 
 requirements = [
-    "pulpcore>=3.3,<3.4",
+    "pulpcore>=3.3",
     "python-debian>=0.1.36",
 ]
 
 setup(
     name="pulp-deb",
-    version="2.3.0b1",
+    version="2.4.0b1.dev",
     description="pulp-deb plugin for the Pulp Project",
     license="GPLv2+",
     author="Matthias Dellweg",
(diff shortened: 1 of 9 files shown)


setup.py
```diff
@@ -2,11 +2,14 @@
 
 from setuptools import find_packages, setup
 
-requirements = ["pulpcore>=3.3.0.dev", "python-debian>=0.1.36"]
+requirements = [
+    "pulpcore>=3.4",
```
quba42, line 6: I noticed that the `pulp_rpm` plugin has still left this requirement at `>=3.3` for the next development version.
Since the version bump in pulpcore happens immediately after the release is branched, this has the practical implication of allowing combinations of the `pulp_rpm` `master` branch with the latest released version of pulpcore.

This did not make sense to me, but If there is a good reason for it, I am all ears.

quba42: And, the failed Travis test has answered my question for me. Travis knows pulpcore versions up to the latest release, and tests against that...

PR #166, opened 2020-04-30
Plugin template updates

quba42:
Updates to the travis comfig from the latest plugin_template plus manual adjustments.


docs/Makefile
```diff
@@ -42,7 +42,7 @@ clean:
 	-rm -rf $(BUILDDIR)/*
 
 html:
-	curl --fail -o _static/api.json "http://localhost:24817/pulp/api/v3/docs/api.json?plugin=pulp_deb&include_html=1"
+	curl -o _static/api.json "http://pulp/api/v3/docs/api.json?plugin=pulp_deb"
```
quba42, line 45: This change makes it impossible to run `make html` from pulplift source development boxes.
This is a shame. Perhaps there is some way to cover both use cases?

mdellweg: good catch. Thank you!

quba42: I will just reference the equivalent discussion on pulp_file: https://github.com/pulp/pulp_file/pull/388#discussion_r420255202
That way I don't need to repeat myself.

PR #167, opened 2020-05-12
Add per repository package uniqueness constraints

Manisha15:
Add uniqueness constraint for debian repository versions such that there are no duplicate packages in repository.

fixes #6429
https://pulp.plan.io/issues/6429

### DIFF
--- a/pulp_deb/app/models/content.py
+++ b/pulp_deb/app/models/content.py
@@ -203,6 +203,8 @@ class BasePackage(Content):
             "{}.{}".format(self.name, self.SUFFIX),
         )
 
+    repo_key_fields = ("package", "version", "architecture")
+
     class Meta:
         default_related_name = "%(app_label)s_%(model_name)s"
         unique_together = (("relative_path", "sha256"),)
--- a/CHANGES/6429.doc
+++ b/CHANGES/6429.doc
@@ -1,0 +1,1 @@
+Added a note on per repository package uniqueness constraints to the feature overview documentation.
--- a/CHANGES/6429.feature
+++ b/CHANGES/6429.feature
@@ -1,0 +1,1 @@
+Added handling of packages with the same name, version, and architecture, when saving a new repository version.
--- a/docs/feature_overview.rst
+++ b/docs/feature_overview.rst
@@ -89,6 +89,14 @@ In general, uploading content works the same way as for any other Pulp plugin, s
    As a result, manually uploaded packages will only show up in your publications, if you are using the "simple" publisher.
    For more information, see :ref:`simple and structured publishing <simple_and_structured_publishing>` below.
 
+.. note::
+   As a matter of best practice, the existence of multiple Debian packages with the same name, version, and architecture (but different content/checksum) should be avoided.
+   Since the existence of such packages may be beyond the control of the ``pulp_deb`` user, the plugin takes a maximally permissive approach:
+   Users can upload arbitrary (valid) packages to the Pulp database, but they cannot add multiple colliding packages of the same type (``.deb`` or ``.udeb``), to a single Pulp repository version.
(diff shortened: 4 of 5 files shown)


pulp_deb/app/migrations/0011_auto_20200505_0937.py
```diff
@@ -0,0 +1,21 @@
...
+    operations = [
+        migrations.AlterUniqueTogether(
+            name='installerpackage',
+            unique_together={('relative_path', 'sha256'), ('package', 'version', 'architecture')},
+        ),
+        migrations.AlterUniqueTogether(
+            name='package',
+            unique_together={('relative_path', 'sha256'), ('package', 'version', 'architecture')},
+        ),
+    ]
```
mdellweg, line 21: This migration is probably outdated (or even unneeded).

quba42: I am pretty sure it is needed (since the change would not show up in my test results until I applied the migration), but probably needs to be regenerated.

mdellweg: There is no connection between `unique_together` and `repo_key_field`.
The former provides uniqueness across all content units (on the database level), and the latter provides uniqueness inside a repository version (with the help of `pulpcore.plugin.repo_version_utils.remove_duplicates`).

Manisha15: This was for older implementation (using uniqueness constraint method). With new approach (repo_key_field) this migration is not generated. Fixed it!

quba42: And now the tests are green, yeah!

pulp_deb/app/models/content.py
```diff
@@ -216,6 +216,8 @@ class Package(BasePackage):
 
     SUFFIX = "deb"
 
+    repo_key_fields = ("package", "version", "architecture")
```
quba42, line 219: @mdellweg Do you think this should move to the `BasePackage` class (so it will apply to `InstallerPackage` as well)?

mdellweg: Good call. I think, yes. (They are insanely similar)

pulp_deb/app/models/content.py
```diff
@@ -198,6 +198,8 @@ def filename(self, component=""):
             "pool", component, prefix, sourcename, "{}.{}".format(self.name, self.SUFFIX)
         )
 
+    repo_key_fields = ("package", "version", "architecture")
```
quba42, line 201: I am thinking we might need to add a checksum to this, in order to achieve what the Debian policy states.

quba42:
I have rebased just the test without the test for now.
The old state is still available on the `deb_repo_uniqueness_test_v1` branch.

pulp_deb/app/models/content.py
quba42: I think this was a nonsense thought...

quba42:
So, having discussed the behaviour with @mdellweg a bit, I have come to the conclusion that this essentially works as designed, though it is not entirely clear to what extent the design accounts for adding several colliding packages in a single command.

The following remains to be done:

- [ ] Test a couple more edge cases and then come to a decision about what sort of test  coverage we need.
- [ ] Mention the effect of this change in the "Feature overview documentation".

This is certainly a significant improvement over current master, which simply creates inconsistent repo versions that will always fail to publish when there are package uniqueness collisions.

quba42:
Due to the importance of this change, as well as the fact that it largely makes use of existing pulpcore features, I will defer the addition of test coverage to a separate issue.

quba42 (approved):

PR #171, opened 2020-05-26
Remove defunct fixtures

quba42:
[noissue]

The tests require up to date fixtures from the pulp-fixtures repo.
As such, these legacy fixtures are out of date duplicates that only
serve to confuse people.

https://github.com/pulp/pulp-fixtures


quba42:
I feel like the fixtures should be mentioned somewhere in a README.md or the docs, but I am unsure what the correct place would be...

@mdellweg Do you have any opinions?

mdellweg:
Having the fixtures in the pulp-fixtures container (which might be served at fixtures.pulp.org or similar) should be a common thing, so maybe in pulpcore?

PR #173, opened 2020-05-28
Fix structured publishing

quba42:
fixes #6787
https://pulp.plan.io/issues/6787

Ensures packages with architecture 'all' are added to all available
package indecies, rather than a non existent 'all' package index.

Note: I believe we can create test coverage for this change, by simply adding some `architecture=all` type packages to the existing fixture repos. I will open a separate PR against https://github.com/pulp/pulp-fixtures

### DIFF
--- a/CHANGES/6787.bugfix
+++ b/CHANGES/6787.bugfix
@@ -1,0 +1,1 @@
+Fixed structured publishing of architecture 'all' type packages.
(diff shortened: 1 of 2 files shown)


quba42:
I cherry-pick https://github.com/pulp/pulp_deb/pull/175 to this issue branch to run the tests.
https://github.com/pulp/pulp_deb/pull/175 should be merged first!

PR #177, opened 2020-06-03
Use the correct relative paths in Release files

quba42:
fixes #6876
https://pulp.plan.io/issues/6876

I also fixed several instances of using the codename in paths. Codenames
should never be used in paths. Always use the "distribution" in paths.

ref #6051
https://pulp.plan.io/issues/6051

### DIFF
--- a/CHANGES/6876.bugfix
+++ b/CHANGES/6876.bugfix
@@ -1,0 +1,1 @@
+Fixed a bug where published Release files were using paths relative to the repo root, instead of relative to the release file.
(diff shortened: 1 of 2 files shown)


quba42:
Not sure if I am referencing the issues correctly.
I meant to reference https://pulp.plan.io/issues/6051 because it relates to the distribution vs. codename issue.
However, this PR does not conclusively fix that issue (only some individual instances I came across as part of the issue this PR actually deals with (https://pulp.plan.io/issues/6876).

I thought that is exactly what "ref #6051" was for. (See https://docs.pulpproject.org/en/master/nightly/contributing/git.html#commits)
However the automation appears to have treated 6051 and 6876 identically (changing both issues to POST).

quba42:
Note: https://pulp.plan.io/issues/6901 would create test coverage for this issue.
I would like to keep this as a separate issue for now, since this is such a critical bug, and the test will probably take a while.

PR #178, opened 2020-06-05
Updating fixture path

daviddavis:
See https://github.com/pulp/pulp-fixtures/pull/160

[noissue]


daviddavis:
This can be merged now.

PR #180, opened 2020-06-12
Refactor (some) class-names to Apt

m-bucher:
Should fix https://pulp.plan.io/issues/6897

Please comment about possible obstacles.


quba42:
Since the original cause for this change, was inconsistent naming between class names, and API paths, I went and had another look at those.
This change appears to bring them exactly into line with each other. (i.e. All the things that have `/apt/` in their API path are now named `Apt<entity>` for their class.

I wanted to draw attention one more time to the verbatim publisher since it is the other thing with an unusual name:
* API path `/pulp/api/v3/publications/deb/verbatim/`, class name `VerbatimPublication`
* API path `/pulp/api/v3/publications/deb/apt/`, class name `AptPublication`

But this now seems structurally consistent.

CHANGES/6897.bugfix
```diff
@@ -0,0 +1,6 @@
+Renamed internal models to Apt.. to keep API consistent with other plugins:
+
+* `DebDistribution` -> `AptDistribution`
+* `DebPublication` -> `AptPublication`
+* `DebRemote` -> `AptRemote`
+* `DebRepository` -> `AptRepository`
```
quba42, line 6: I had not seen multiline changelog entries before, so I went and checked what `towncrier` does with them.
Sure enough, he appends the issue link to the end of the list, resulting in the following error when the docs are built:
```
../CHANGES.rst:42:Bullet list ends without a blank line; unexpected unindent.
```

We can correct this when we create a release, or we can go back to using this as intended and just make it a single line or paragraph...

We also need to use double backticks if we want inline code blocks in the final output, because ReStructuredText.

Personally, I would write something like:
```
Renamed many class names from ``Deb...`` to ``Apt...`` for consistency with the API paths.
```
This will be linked with the upstream issue for anyone that wants to read more.

We could also consider naming the file `6897.misc` instead of `6897.bugfix` since I am not entirely sure to what extent this is really a bugfix. :stuck_out_tongue_winking_eye:

quba42:
> I really like it, but I also do see one possible obstacle. Can you please try the migration with a seeded db. I am unsure if existing objects are properly accounted for. You might need to add a handcrafted step.

I am not sure if this is a meaningful test, but for what it is worth, I was still able to find everything in my source development box after applying the migration.

How else can I test this?

CHANGES/6897.bugfix
mdellweg: What if you add the empty line here?

quba42: Tried that, did not work (for whatever reason...)

quba42: Like I said, if we really want the bullet list it is not a major problem to fix it manually when we create a release, but it might be more trouble than it is worth.

quba42:
@m-bucher Do we need to confirm one more time that this really does fix the issue with the auto generated API bindings?

m-bucher:
@quba42 I will check it out in my forklift.

m-bucher:
@quba42 The tests now no longer produce the broken URL, so I guess that is a good sign :smiley:

quba42:
@mdellweg I will have one more look at a migration tomorrow, and then I would merge this.

PR #181, opened 2020-06-15
Fix nullable DB fields to set to null via the API

Manisha15:
Fix nullable DB fields which could not be set to null via the API

fixes #6908
https://pulp.plan.io/issues/6908

### DIFF
--- a/CHANGES/6908.bugfix
+++ b/CHANGES/6908.bugfix
@@ -1,0 +1,1 @@
+Fixed a bug where some nullable fields for remotes could not be set to null via the API.
--- a/pulp_deb/app/serializers/remote_serializers.py
+++ b/pulp_deb/app/serializers/remote_serializers.py
@@ -16,11 +16,15 @@ class DebRemoteSerializer(RemoteSerializer):
     )
 
     components = CharField(
-        help_text="Whitespace separatet list of components to sync", required=False,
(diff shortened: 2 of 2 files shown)


CHANGES/6908.bugfix
```diff
@@ -0,0 +1 @@
+Fixed nullable DB fields which could not be set to null via the API
```
mdellweg, line 1: I only see a single one, but maybe `architectures` should be treated that way, too.

Manisha15: I thought the ticket was for components only but if it was for all attributes I will fix them all.

quba42: Do we also need it for the `gpgkey` field? Is it a valid use case to add a gpg key to a remote and then later remove it again completely?

quba42, line 1: For the changelog entry, perhaps the following might be a bit clearer for users:
```suggestion
Fixed a bug where some nullable fields for remotes could not be set to null via the API.
```

mdellweg: Probably yes. "Des Menschen Wille ist sein Himmelreich." ("Someones will is his kingdom of heavens.")

quba42:
Same here, we should still fixup/squish these two commits into one.

PR #182, opened 2020-06-17
Prepare release 2.4.0b1 (beta)

quba42:
This is mainly needed so people can install the plugin along with pulpcore `3.4`.

This nevertheless remains a beta release!


quba42:
Now that I see them, I fully agree that two commits are better.

PR #183, opened 2020-06-30
Rework Workflow Documentation

maximiliankolb:
First draft for Issue #6398


docs/workflows/sync.rst
```diff
@@ -1,91 +1,161 @@
...
+      "tls_validation": true,
+      "url": "http://nginx.org/packages/debian",
+      "username": null
+  }
+
+Sync Repository with Remote
+-------------------------------
 
-``$ http POST $BASE_ADDR/pulp/api/v3/remotes/deb/apt/ name='bar' url='http://some.url/somewhere/'``
+Use the remote object to kick off a synchronize task by specifying the repository to sync with.
```
mdellweg, line 75: It's the other way around. The remote is the parameter to the sync endpoint on the repository.

docs/workflows/upload.rst
```diff
@@ -1,94 +1,344 @@
...
-Create a repository
--------------------
+.. image:: upload.svg
+  :alt: Upload a package
 
-If you don't already have a repository, create one::
+This requires you to already have a :ref:`repository created <create_a_repository>`.
 
-    $ http POST $BASE_ADDR/pulp/api/v3/repositories/deb/apt/ name=foo
+If you don't have any repositories yet, create one:
```
mdellweg, line 9: Can we move all of this to a `create_a_repository` section and just state: We assume you have a repository named `vim`.
This section should really focus on the package upload.

docs/workflows/upload.rst
```diff
@@ -1,94 +1,344 @@
...
+      ],
+      "started_at": "2020-06-29T07:40:53.218540Z",
+      "state": "completed",
+      "task_group": null,
+      "worker": "/pulp/api/v3/workers/50a13e76-fe27-4e3e-8cee-ae5ec41d272a/"
+  }
+
+.. note::
+
+  A common pool location will be generated if you do not specify a ``relative_path``.
```
mdellweg, line 193: Read that before...
Would it be enough to say something like: "Instead of specifying an existing artifact, you can upload a file with the following command. The rest is the same."

quba42:
I am in favour of creating a sub task of https://pulp.plan.io/issues/6398 for this PR.
Something along the lines of "Update the workflow documentation", since that accurately reflects the scope of this PR.

maximiliankolb:
Thanks for your initial review @mdellweg. Would you please have a look at it again as I incorporated all your suggestions and fixed the format of the PR.
Any feedback is welcome.

I would very much like to move forward with the documentation and possibly add and extend it in the future.

CC @quba42

quba42:
I will have a look at it (as soon as my pulplift boxes are up and I have had a chance to look at some of the other pulp_deb PRs).

CHANGES/6398.doc
```diff
@@ -0,0 +1 @@
+Reworked workflow and added flow charts
```
quba42, line 1: I created a new ticket that more accurately reflects the scope of this change: https://pulp.plan.io/issues/7148
(I tried to create it as a sub task of 6389, but that did not work!?)
I would also suggest changing the text to something like:
```
Reworked the workflow documentation and added flow charts.
```
I find this clearer even if it is already implied by the `.doc` extension.

PR #184, opened 2020-07-13
Fix APT client installing patches

Manisha15:
fixes APT client installing same packages again and again due to missing installed_size.

fixes #6982
https://pulp.plan.io/issues/6982

### DIFF
--- a/pulp_deb/app/serializers/content_serializers.py
+++ b/pulp_deb/app/serializers/content_serializers.py
@@ -202,7 +202,7 @@ class BasePackage822Serializer(SingleArtifactContentSerializer):
         "bugs": "Bugs",
         "essential": "Essential",
         "build_essential": "Build_essential",
-        "installed_size": "Installed_size",
+        "installed_size": "Installed-Size",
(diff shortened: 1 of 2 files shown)


pulp_deb/app/serializers/content_serializers.py
```diff
@@ -202,7 +202,7 @@ class BasePackage822Serializer(SingleArtifactContentSerializer):
         "bugs": "Bugs",
         "essential": "Essential",
         "build_essential": "Build_essential",
-        "installed_size": "Installed_size",
+        "installed_size": "Installed-Size",
```
mdellweg, line 205: I'm just comparing the lines above and below here... Are you sure it's going to be a dash instead of that underscore? I can see that the capital "S" might have been the problem.

Manisha15: Yes, I double checked Package file of remote repositories and its dash and not underscore.

mdellweg: Fine for me.

PR #185, opened 2020-07-14
Do not sync repository with bad remote repository

Manisha15:
Do not sync repository with bad remote repository that has non-existing components or architectures.

fixes #6948
https://pulp.plan.io/issues/6948,

### DIFF
--- a/pulp_deb/app/tasks/synchronizing.py
+++ b/pulp_deb/app/tasks/synchronizing.py
@@ -81,6 +81,20 @@ class NoPackageIndexFile(Exception):
     pass
 
 
+class MissingAttribute(Exception):
+    """
+    Exception to signal, that no atrribute representing a valid attritute is present.
+    """
+
+    def __init__(self, distribution, *args, **kwargs):
+        """
+        Exception to signal, that no atrribute representing a valid attritute is present.
+        """
+        super().__init__("No valid attribute found in '{}'.".format(distribution), *args, **kwargs)
+
+    pass
+
+
 def synchronize(remote_pk, repository_pk, mirror):
     """
     Sync content from the remote repository.
@@ -181,6 +195,8 @@ class DebUpdateReleaseFileAttributes(Stage):
         super().__init__(*args, **kwargs)
         self.remote = remote
         self.gpgkey = remote.gpgkey
+        self.components = remote.components
+        self.architectures = remote.architectures
         if self.gpgkey:
             gnupghome = os.path.join(os.getcwd(), "gpg-home")
             os.makedirs(gnupghome)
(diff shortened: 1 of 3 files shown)


quba42:
It looks like their is a failed test because the new exceptions are thrown before a different exception we are testing for are thrown.
We will need to fix that test (and possibly add one for this exception...)

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -263,6 +279,16 @@ async def run(self):
                     log.debug("Codename: {}".format(release_file.codename))
                     log.debug("Components: {}".format(release_file.components))
                     log.debug("Architectures: {}".format(release_file.architectures))
+
+                    if "Release" or "InRelease" in da_names:
```
mdellweg, line 283: This seems a little bit inconsistent. If we call for a release and don't find it, we ignore it, but if one of the releases we found lacks a component or architecture, we fail hard.

Also i see a real problem in that architectures get added and removed between Debian releases. And this change would prevent to specify a subset of architectures when synching multiple releases, and one of the arch's is missing from one of the releases.

quba42: So I guess these really are two separate points:

1. We should be consistent. If we are going to "validate" `components` and `architectures`, we should also validate `distributions` (releases) in the same way.
2. There is a conflict between preserving the valid (but perhaps not recommended) use case of wanting to recreate multiple upstream `distributions` (within a single pulp repository), while at the same time filtering/restricting to for example, a subset of the available architectures (if present) on the one hand, and wanting to catch bad user input on the other hand.

I do think that being maximally permissive with user input (and simply silently publishing a bunch of empty `distributions`, `compnents`, or `architectures`) whenever users make mistakes, opens the door to a lot of hard to diagnoze problems for most users to preserve a pretty rare specialist use case.

We could try to mitigate this with documentation (but most users wont read it).

I will change the status of this PR to draft until we have had a chance to discuss this further.

quba42: I guess it comes down to: Do we want it to be a valid use case to publish many distributions in a single pulp repository, while filtering the components and/or architectures, even if not all of the upstream distributions have all of the components/architectures we are filtering for.

mdellweg: And i think the answer is yes.
BTW: "<...> even if not all of the upstream distributions have all of the components/architectures we are filtering for." that is the part we have no control over.
Another, much more involved solution might be to explicitly specify the components to sync from per release requested.

If we want to keep the current, easy to grasp, configuration model, we could fail hard only on not found releases, to turn the inconsistency in the opposite direction. I guess that would catch most configuration errors, while preserving all the filtering features for arch/comp.

quba42:
I have been thinking about this a bit more. For reasons of both usage and implementation I do not think asking for structured user input is practical.

The only solution I can come up with to satisfy all use cases is making the behaviour configurable (globally and/or per remote).
Lets say a remote could have `"strict": true` or `"strict": false`.

A strict remote would throw a hard error whenever a remote repository is missing a distribution known to the remote, or when one of the distributions is missing a component or architecture known to the remote.
A non strict remote would log a warning instead.

Personally I would vote that strict remotes are the default behaviour (since I think most users should keep their pulp repos simple (for example by keeping it one distribution per repo) and would benefit from sanity checking on the remotes they create). However, that is debatable since the non strict remote is more "liberal" in what use cases and input it accepts (https://en.wikipedia.org/wiki/Robustness_principle).

Any thoughts on the general approach?

PR #187, opened 2020-07-21
Add test for known cases where distribution is not codename or suite

Manisha15:
fixes #6051
https://pulp.plan.io/issues/6051

### DIFF
--- a/CHANGES/6051.misc
+++ b/CHANGES/6051.misc
@@ -1,0 +1,1 @@
+Added tests for repos with distribution paths that are not equal to the codename.
--- a/pulp_deb/tests/functional/api/test_publish_complex_dists.py
+++ b/pulp_deb/tests/functional/api/test_publish_complex_dists.py
@@ -1,0 +1,175 @@
+# coding=utf-8
+"""Tests that sync deb plugin repositories."""
+import unittest
+import os
+
+from pulp_smash import config
+
+from pulp_smash.pulp3.utils import (
+    gen_repo,
+    get_content,
+    delete_orphans,
+    get_versions,
+    gen_distribution,
+    download_content_unit,
+)
+
+from pulp_deb.tests.functional.constants import (
+    DEB_COMPLEX_DISTS_FIXTURE_URL,
+    DEB_RELEASE_FILE_NAME,
+    DEB_RELEASE_NAME,
+    DEB_RELEASE_COMPONENT_NAME,
+    DEB_PACKAGE_INDEX_NAME,
+)
+from pulp_deb.tests.functional.utils import set_up_module as setUpModule  # noqa:F401
+from pulp_deb.tests.functional.utils import (
+    gen_deb_remote,
+    monitor_task,
+    deb_remote_api,
+    deb_repository_api,
+    deb_apt_publication_api,
+    deb_distribution_api,
+)
+
+from pulpcore.client.pulp_deb import (
+    RepositorySyncURL,
+    DebAptPublication,
+)
+
+
+class ComplexDistSyncTestCase(unittest.TestCase):
+    """Sync a repository with the deb plugin."""
+
+    def _publication_extra_args(self):
+        return {"structured": True}
+
+    def setUp(self):
+        """Cleanup."""
+        delete_orphans()
+
+    def test_publish_complex_dist_ubuntu_backports(self):
+        """Test publishing repository with distrbution ragnarok-backports."""
+        expected_values = {
+            "distribution": "ragnarok-backports",
+            "codename": "ragnarok",
+            "suite": "ragnarok-backports",
+            "components": ["asgard", "jotunheimr"],
+            "release_file_folder": "dists/ragnarok-backports/",
+            "package_index_paths": [
+                "dists/ragnarok-backports/asgard/binary-ppc64/Packages",
+                "dists/ragnarok-backports/asgard/binary-armeb/Packages",
+                "dists/ragnarok-backports/jotunheimr/binary-ppc64/Packages",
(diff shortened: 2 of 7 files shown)


pulp_deb/tests/functional/api/test_publish_complex_dists.py
```diff
@@ -0,0 +1,124 @@
...
+            if (
+                component["release"] == release["pulp_href"]
+                and release["distribution"] == "ragnarok/updates"
+            ):
+                self.assertIn(component["component"], ["updates/asgard", "updates/jotunheimr"])
+            elif (
+                component["release"] == release["pulp_href"]
+                and release["distribution"] == "ragnarok-backports"
+            ):
+                self.assertIn(component["component"], ["asgard", "jotunheimr"])
```
quba42, line 124:
```suggestion
            if component["release"] == release["pulp_href"]:
                if release["distribution"] == "ragnarok/updates":
                    self.assertIn(component["component"], ["updates/asgard", "updates/jotunheimr"])
                elif release["distribution"] == "ragnarok-backports":
                    self.assertIn(component["component"], ["asgard", "jotunheimr"])
```
Would this be slightly more readable?

The other thing I learned the hard way when writing tests that loop over a list to assert something about one or more specific entries, is to make sure that the test does not pass if no entries are found. If the release-component combination we want to test is missing the test will turn green without ever checking any assertions. Or is this already ensured by earlier assertions?

CHANGES/6051.misc
```diff
@@ -0,0 +1 @@
+Added tests for known use cases of complex distributions where distribution is not the codename or suite
```
quba42, line 1:
```suggestion
Added tests for known use cases of complex distributions where distribution is not the codename or suite.
```
I think we usually make it a sentence with puntuation. :wink:

pulp_deb/tests/functional/api/test_publish_complex_dists.py
Manisha15: Thank you for the suggestions.

No, this asertion case is not covered yet. I will fix it.

pulp_deb/tests/functional/api/test_publish_complex_dists.py
```diff
@@ -0,0 +1,129 @@
...
+    DebAptPublication,
+)
+
+
+class ComplexDistSyncTestCase(unittest.TestCase):
+    """Sync a repository with the deb plugin."""
+
+    class Meta:
+        publication_api = deb_apt_publication_api
+        Publication = DebAptPublication
```
mdellweg, line 39: This mechanism is not used by this test. Is something planned or can we remove it?

quba42: Probably an artifact caused by copying from another test :wink:

Manisha15: @mdellweg The mechanism is used in line #60 and then from line #80 on to publish the repository.

mdellweg: The definitions are used. Yes. But the mechanism i am referring to is that you can subclass that class change it's Meta member class and you have a second test case (i.e. for verbatim publishing). But there is no use for this here afaik.

quba42: @mdellweg, so you would just eliminate the "Meta" class and call things directly?

pulp_deb/tests/functional/api/test_publish_complex_dists.py
```diff
@@ -0,0 +1,124 @@
...
+    RepositorySyncURL,
+    DebAptPublication,
+)
+
+
+class ComplexDistSyncTestCase(unittest.TestCase):
+    """Sync a repository with the deb plugin."""
+
+    def _publication_extra_args(self):
+        return {"simple": True}
```
quba42, line 38: We need to use `"structured": True` as well. (This is one of the main cases we want to test). I am not sure if we should leave the `"simple": True` or get rid of it. I will need to think about this one more time.

pulp_deb/tests/functional/constants.py
```diff
@@ -92,6 +92,9 @@ def _clean_dict(d):
 
 DEB_INVALID_FIXTURE_URL = urljoin(PULP_FIXTURES_BASE_URL, "debian-invalid/")
 
+DEB_COMPLEX_DISTS_FIXTURE_URL = urljoin(PULP_FIXTURES_BASE_URL, "debian-complex-dists/")
+DEB_COMPLEX_DISTS_FIXTURE_DISTRIBUTIONS = "ragnarok/updates ragnarok-backports"
```
quba42, line 96: My apologies for not thinking of this sooner, but I think it would be better if we did not sync both these distributions within the single test.
I think we should have two separate tests that both inherit from the same parent class in such a way that they are identical, except that one of them uses `distribution=ragnarok/updates` and the other uses `distribution=ragnarok-backports`.

The reason is, that when these tests fail, they currently keep running into collisions between the two distributions instead of into our assertions, which makes it very difficult to understand what went wrong.

quba42:
I expanded and refactored the test, until it was actually able to catch the tiny remaining bug I found.
I then also fixed said bug (along with some renaming of things, hopefully for greater clarity...)

The tests are currently red, because I have a fixup commit (and have not added any `[noissue]` tags.
I am fairly confident the tests will pass (based on my local testing).

If we want to be fully comprehensive with the changelog, then I guess I should still open a separate issue for the bug I fixed, and only enable the new tests with the bugfix. I will leave this and further review until Monday...

PR #188, opened 2020-07-22
Rework the README file

quba42:
[noissue]

I was thinking the two extra links can't hurt. :wink:


mdellweg:
Is this from the Plugin Template? Maybe you want to do a similar thing there.

quba42:
Will have a look at the plugin template.

quba42:
I made some more changes, based on what I saw in the current plugin template. (Most notably converting from `.rst` back to `.md`.)

I don't think it makes a lot of sense to also change the plugin template, since it only provides only a very generic README to start with, that will probably need to be manually adjusted for most new plugins as they transition from a skeleton to a real fleshed out plugin.

PR #190, opened 2020-07-23
Fix gnupg metadata verification

quba42:
It looks like a newer version of python2-gnupg now expects the gnupg
home folder to exist before any keys are imported (whereas the older
version presumably just created it if it was not there already).

### DIFF
--- a/plugins/pulp_deb/plugins/importers/sync.py
+++ b/plugins/pulp_deb/plugins/importers/sync.py
@@ -184,7 +184,10 @@ class ParseReleaseStep(publish_step.PluginStep):
         # check signature
         if not self.get_config().get_boolean(constants.CONFIG_REQUIRE_SIGNATURE, False):
             return
-        gpg = self.gnupg_factory(homedir=os.path.join(self.get_working_dir(), 'gpg-home'))
+        worker_gpg_homedir = os.path.join(self.get_working_dir(), 'gpg-home')
+        if not os.path.exists(worker_gpg_homedir):
+            os.mkdir(worker_gpg_homedir, 0o700)
+        gpg = self.gnupg_factory(homedir=worker_gpg_homedir)
         shared_gpg = self.gnupg_factory(homedir=os.path.join('/', 'var', 'lib', 'pulp', 'gpg-home'))
 
         if self.get_config().get(constants.CONFIG_GPG_KEYS):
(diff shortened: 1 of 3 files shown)


plugins/pulp_deb/plugins/importers/sync.py
```diff
@@ -184,7 +184,9 @@ def verify_release_file(self, distribution):
         # check signature
         if not self.get_config().get_boolean(constants.CONFIG_REQUIRE_SIGNATURE, False):
             return
-        gpg = self.gnupg_factory(homedir=os.path.join(self.get_working_dir(), 'gpg-home'))
+        worker_gpg_homedir = os.path.join(self.get_working_dir(), 'gpg-home')
+        os.mkdir(worker_gpg_homedir, 0o700)
```
m-bucher, line 188: Can we guarantee that the directory cannot already be there?
Because if the directory might be there, the pyhton2-docu states the following for `os.mkdir()`:
>  If the directory already exists, OSError is raised.

quba42: I was also thinking about putting a guard on it. I am not sure how the worker directories are handled. If they are cleaned up after each task (which is at least naively plausible), then a guard would be entirely superfluous.

quba42: If they are not cleaned up, then it might almost make sense to perform the entire thing in a temp dir. (I guess the whole point is to individually import the one GPG key we want into a fresh gnupghome for each publish.)

quba42: This would be the alternative version of the fix:
```diff
-        gpg = self.gnupg_factory(homedir=os.path.join(self.get_working_dir(), 'gpg-home'))`
+        worker_gpg_homedir = os.path.join(self.get_working_dir(), 'gpg-home')
+        if not os.path.exists(worker_gpg_homedir):
+            os.mkdir(worker_gpg_homedir, 0o700)
+        gpg = self.gnupg_factory(homedir=worker_gpg_homedir)
```

quba42: I decided to go with the guarded version, because it does not hurt and is safer.
At this point there is not much incentive to trying to keep the pulp2 code base super lean...

PR #192, opened 2020-07-23
Fix API documentation

quba42:
fixes #6506
https://pulp.plan.io/issues/6506


quba42:
Since it is all green, and merely fixes some URLs in the docs, I will just go ahead and merge this on my own authority...

PR #195, opened 2020-08-03
Modify data type of fields in content model

Manisha15:
fixes #7257
https://pulp.plan.io/issues/7257


CHANGES/7257.bugfix
```diff
@@ -0,0 +1 @@
+Fixed several fields data type in the content model.
```
quba42, line 1:
```suggestion
Fixed content creation for fields containing more than 255 characters by using ``TextField`` instead of ``CharField`` for all package models. 
```

mdellweg:
It would be interesting whether you can still filter for say architecture=i386 if it is a textfield.

quba42:
> It would be interesting whether you can still filter for say architecture=i386 if it is a textfield.

Tested using `http get :/pulp/api/v3/content/deb/packages/?architecture=i386` and `http get :/pulp/api/v3/content/deb/packages/?architecture=amd64` with results as one would expect.

PR #197, opened 2020-08-11
Fix relative path of packages

Manisha15:
fixes #7153
https://pulp.plan.io/issues/7153

### DIFF
--- a/CHANGES/7153.bugfix
+++ b/CHANGES/7153.bugfix
@@ -1,0 +1,1 @@
+Optional version strings are now stripped from the sourcename before using it for package file paths.
--- a/pulp_deb/app/models/content.py
+++ b/pulp_deb/app/models/content.py
@@ -190,12 +190,13 @@ class BasePackage(Content):
     def filename(self, component=""):
         """Assemble filename in pool directory."""
         sourcename = self.source or self.package
+        sourcename = sourcename.split("(", 1)[0].rstrip()
         if sourcename.startswith("lib"):
             prefix = sourcename[0:4]
         else:
             prefix = sourcename[0]
         return os.path.join(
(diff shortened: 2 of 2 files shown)


pulp_deb/app/models/content.py
```diff
@@ -195,7 +195,7 @@ def filename(self, component=""):
         else:
             prefix = sourcename[0]
         return os.path.join(
-            "pool", component, prefix, sourcename, "{}.{}".format(self.name, self.SUFFIX)
+            "pool", component, prefix, sourcename.split()[0], "{}.{}".format(self.name, self.SUFFIX)
```
quba42, line 198: I went and double checked the Debian source name specification here:
https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-source

The Key bits:
> In a binary package control file or a .changes file, the source package name may be followed by a version number in parentheses. [4]
>
> [4] It is customary to leave a space after the package name if a version number is specified.

This implies that the current fix will only work so long as packages adhere to what is "customary".
I think we should also make sure that we strip any version that does not leave a space, that is, that we remove anything past the first open parentheses `(`.

Note that:
> Package names (both source and binary) must consist only of lower case letters (a-z), digits (0-9), plus (+) and minus (-) signs, and periods (.). They must be at least two characters long and must start with an alphanumeric character.

CHANGES/7153.bugfix
```diff
@@ -0,0 +1 @@
+Fixed relative_path of packages so that packages have relative path same as of remote repository packages.
```
quba42, line 1:
```suggestion
Optional version strings are now stripped from the sourcename before using it for package file paths.
```

pulp_deb/app/models/content.py
quba42: We could also consider stripping things right after line 192, rather than when we use it in line 198.

PR #198, opened 2020-08-13
Add high level feature overview documentation

quba42:
fixes #7318
https://pulp.plan.io/issues/7318


docs/feature_overview.rst
```diff
@@ -0,0 +1,134 @@
...
+
+Repository Synchronization
+--------------------------------------------------------------------------------
+
+Synchronizing upstream repositories is one of two ways to obtain content for the ``pulp_deb`` plugin.
+(The other one being :ref:`package uploads <package_uploads>`.)
+
+The aim is for the plugin to be able to synchronize (and usably publish) arbitrary (valid) APT repositories.
+Most of the testing is based on the structure of the official Debian and Ubuntu repositories.
+As of this writing all testing happens with regular binary package repositories only.
```
mdellweg, line 24: Isn't there at least a udeb and maybe one installer_file in the fixtures?

mdellweg: Also what is the point of talking about tests in a "feature overview"?

quba42: I guess I was not very clear on supported package types.
Perhaps I should make that a completely separate heading.
The point is to give some indication what can really be expected to work for sure.

We do have a separate model for udebs (`InstallerPackage` I think).
However, if I have understood things correctly (and please correct me if I am wrong), those are only usable in conjunction with the verbatim publisher right now?

I don't think any of us have ever tried to upload source packages? (If they work, then by accident?)

Linux Mint repos (for example) should work, but I don't think I have ever tested one.

That being said, I will rework this paragraph (once I have cleared up some of these points for myself).

mdellweg: Publishing udebs without the 'other' installer files doesn't bear any value, so it is at least only useful in conjunction with the verbatim publisher atm.

quba42:
In the interest of reviewability I have split out https://github.com/pulp/pulp_deb/pull/202 into a separate "spin off PR".
I still think both are required for a complete "feature overview" documentation.

quba42:
Added many suggestion by @maximiliankolb

PR #199, opened 2020-08-19
Add tests for publishing empty repository

Manisha15:
fixes #7344
https://pulp.plan.io/issues/7344

### DIFF
--- a/pulp_deb/tests/functional/api/test_publish_empty_repository.py
+++ b/pulp_deb/tests/functional/api/test_publish_empty_repository.py
@@ -1,0 +1,115 @@
+# coding=utf-8
+"""Tests that sync deb plugin repositories."""
+import unittest
+
+from pulp_smash import config
+
+from pulp_smash.pulp3.utils import (
+    gen_repo,
+    get_content,
+    delete_orphans,
+    get_versions,
+    gen_distribution,
+    download_content_unit,
+)
+
+from pulp_deb.tests.functional.constants import (
+    DEB_FIXTURE_URL,
+    DEB_PACKAGE_INDEX_NAME,
+    DEB_PACKAGE_NAME,
+)
+from pulp_deb.tests.functional.utils import set_up_module as setUpModule  # noqa:F401
+from pulp_deb.tests.functional.utils import (
+    gen_deb_remote,
+    monitor_task,
+    deb_remote_api,
+    deb_repository_api,
+    deb_apt_publication_api,
+    deb_distribution_api,
+)
+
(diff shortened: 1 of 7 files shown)


pulp_deb/tests/functional/api/test_publish_empty_repository.py
```diff
@@ -0,0 +1,88 @@
...
+
+        # Test the publication:
+        self.assertEqual(publication.repository_version, version_hrefs[-1])
+
+        release = get_content(
+            repo=publication.to_dict(), version_href=publication.repository_version
+        )
+
+        self.assertFalse(release[DEB_PACKAGE_NAME])
+        self.assertTrue(release[DEB_PACKAGE_INDEX_NAME])
```
quba42, line 88: I think we are nearly done with this test, but it needs one more extension.

Currently we are testing for the presence of package index files as synced for use with the verbatim publisher.

We also need to test for the presence of package index files as created by the APT publisher.
These are only visible via the content app.

Therefore we need to create a distribution and download the package indexes we want to test for.
Fortunately we have a working example for how to do this in the lines past `161` here:
https://github.com/pulp/pulp_deb/blob/master/pulp_deb/tests/functional/api/test_publish_complex_dists.py

quba42 (changes requested):
Hopefully the documentation I added here (https://github.com/pulp/pulp_deb/pull/198) and here (https://github.com/pulp/pulp_deb/pull/202) can demystify why we have multiple sets of package indices within pulp (those synced from upstream, as well as those created by the APT publisher).

pulp_deb/tests/functional/api/test_publish_empty_repository.py
```diff
@@ -0,0 +1,118 @@
...
+
+        release = get_content(
+            repo=publication.to_dict(), version_href=publication.repository_version
+        )
+
+        package_index_paths = [
+            "dists/ginnungagap/asgard/binary-ppc64/Packages",
+            "dists/ginnungagap/jotunheimr/binary-armeb/Packages",
+            "dists/ginnungagap/asgard/binary-armeb/Packages",
+            "dists/ginnungagap/jotunheimr/binary-ppc64/Packages",
```
quba42, line 97: If you rebase to current master, we can add `"dists/default/all/binary-all/Packages",` to this list, and also cover the simple publisher.

Note: you will also need to use `len(package_index_paths)-1` in line 102 for the test to turn green.

quba42:
Apart from my final comment this is ready to be merged (as far as I can tell).

quba42 (approved):

PR #200, opened 2020-08-19
Add architecture=all consistent for simple and structured mode

Manisha15:


pulp_deb/app/tasks/publishing.py
```diff
@@ -172,23 +171,29 @@ def __init__(self, parent, component):
                 open(package_index_path, "wb"),
                 package_index_path,
             )
+        if "all" not in self.parent.architectures:
```
quba42, line 174: Rather than catching this special case here, I think we should simply ensure `all` is always present in the list passed to the `_ReleaseHelper` for both simple (see line 96) and structured (see line 122) mode.

This would have several benefits:

1. The `all` string would also be added to the `Architectures` field in our generated `Release` files, which is a requirement if `architecture=all` type packages are not added to the architecture specific `Packages` files.
1. We can remove this `if` block, since `all` will always be included in the `for` loop from line 161.
1. It would actually somewhat mitigate the problem we identified with the simple publisher in relation to empty repositories (https://github.com/pulp/pulp_deb/pull/199), because the simple publisher would at least include an empty `all` architecture.

If we do it this way, we would still publish (empty) `binary-all` package indices when there are no `Architecture: all` type packages. However, this is entirely legal, and indeed appears to be how upstream Debian does things (e.g.: http://ftp.de.debian.org/debian/dists/bullseye-updates/contrib/binary-all/).

pulp_deb/app/tasks/publishing.py
```diff
@@ -158,7 +158,6 @@ def __init__(self, parent, component):
         self.component = component
         self.plain_component = os.path.basename(component)
         self.package_index_files = {}
-
```
quba42, line 161: We should not change white space if we don't have to.

pulp_deb/app/tasks/publishing.py
```diff
@@ -172,23 +171,29 @@ def __init__(self, parent, component):
...
+                "dists", self.parent.distribution, self.plain_component, "binary-all", "Packages",
+            )
+            os.makedirs(os.path.dirname(package_index_path), exist_ok=True)
+            self.package_index_files["all"] = (
+                open(package_index_path, "wb"),
+                package_index_path,
+            )
 
     def add_package(self, package):
+
```
quba42, line 185: We should not change white space if we don't have to. (Unless this came from black I guess.)

quba42 (changes requested):
The current version seems to work as expected.
However, if I understand (https://wiki.debian.org/DebianRepository/Format#Architectures) correctly, we also need to make sure the string `all` is included in the `Architectures` field of our `Release` files.

See my comment below on how best to achieve that.

quba42 (approved):

quba42:
Looking good!

PR #202, opened 2020-08-24
Add meaningful REST API endpoint descriptions

quba42:
fixes #7355
https://pulp.plan.io/issues/7355

Motivated by https://github.com/pulp/pulp_deb/pull/198

The existing REST API docs (for reference): https://pulp-deb.readthedocs.io/en/latest/restapi.html

### DIFF
--- a/pulp_deb/app/viewsets/content.py
+++ b/pulp_deb/app/viewsets/content.py
@@ -20,8 +20,14 @@ class GenericContentFilter(ContentFilter):
 
 
 class GenericContentViewSet(SingleArtifactContentUploadViewSet):
+    # The doc string is a top level element of the user facing REST API documentation:
     """
-    A ViewSet for GenericContent.
+    GenericContent is a catch all category for storing files not covered by any other type.
+
+    Associated artifacts: Exactly one arbitrary file that does not match any other type.
+
+    This is needed to store arbitrary files for use with the verbatim publisher. If you are not
+    using the verbatim publisher, you may ignore this type.
     """
 
     endpoint_name = "generic_contents"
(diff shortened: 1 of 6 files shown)


quba42:
`flake8` is demanding both that doc strings start with a single line ending in a period, followed by a blank line, AND that lines are no longer than 100 characters.

These two rules combined essentially make it impossible, to have a meaningful first line in doc strings for classes involving complex concepts.

I think at least one of these `flake8` rules should be dropped, but I am unsure what a good style convention would be.
Does anyone have any suggestions?

quba42:
`flake8` is demanding both that doc strings start with a single line ending in a period, followed by a blank line, AND that lines are no longer than 100 characters.

These two rules combined essentially make it impossible, to have a meaningful first line in doc strings for classes involving complex concepts.

I think at least one of these `flake8` rules should be dropped, but I am unsure what a good style convention would be.
Does anyone have any suggestions?

pulp_deb/app/viewsets/content.py
```diff
@@ -60,8 +66,11 @@ class Meta:
 
 
 class PackageViewSet(SingleArtifactContentUploadViewSet):
+    # The doc string is a top level element of the user facing REST API documentation:
```
mdellweg, line 69: Nice catch!

quba42: First, I wanted to write about each content type in the documentations "Feature Overview" chapter. Then I noticed that I was basically writing a second REST API documentation, so I went and had a look at the actual REST API documentation, and noticed all the descriptions were "A ViewSet for <class_name>.".

So it did take a couple of steps to get  there. :stuck_out_tongue_winking_eye:

PR #208, opened 2020-09-02
Disable daily client builds

quba42:
[noissue]

Unless somebody can provide me with a convincing use case, I would like to disable daily client builds.


quba42:
So apparently Katello likes to have the frequent client builds, so i guess I will close this.

PR #210, opened 2020-09-02
Deprecation issues

quba42:
I am opening this as a draft pull request, because I need some help with evaluating the changes to `pyproject.toml` (which came from the plugin template).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,22 +1,21 @@
-[tool.towncrier]
-    package = "pulp_deb"
-    filename = "CHANGES.rst"
-    directory = "CHANGES/"
-    title_format = "{version} ({project_date})"
-    template = "CHANGES/.TEMPLATE.rst"
-    issue_format = "`#{issue} <https://pulp.plan.io/issues/{issue}>`_"
+[build-system]
+requires = [
+    "setuptools >= 40.8.0",
+    "wheel >= 0.29.0",
+]
+build-backend = 'setuptools.build_meta'
 
-    [[tool.towncrier.section]]
-        path = ""
+[tool.towncrier]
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -1,24 +1,23 @@
...
-    directory = "CHANGES/"
-    title_format = "{version} ({project_date})"
-    template = "CHANGES/.TEMPLATE.rst"
-    issue_format = "`#{issue} <https://pulp.plan.io/issues/{issue}>`_"
+[build-system]
+requires = [
+    "setuptools >= 40.8.0",
+    "wheel >= 0.29.0",
+]
+build-backend = 'setuptools.build_meta'
```
quba42, line 6: Is this new section useful? needed? what does it do?

pyproject.toml
```diff
@@ -29,10 +28,21 @@
...
+    [[tool.towncrier.type]]
+        directory = "deprecation"
+        name = "Deprecations"
+        showcontent = true
+
     [[tool.towncrier.type]]
         directory = "misc"
         name = "Misc"
-        showcontent = true
+        showcontent = false
```
quba42, line 44:
```suggestion
        showcontent = true
```
I think I'd rather keep this `true`.

quba42:
Other than my two comments only the order seems to have changed.

pyproject.toml
quba42: https://github.com/pulp/plugin_template/commit/ce246bda2e2fbde76c64228950f7d6197703eaf2

quba42:
It looks like travis got stuck somehow. The individual jobs say passed, but the overview page still says "created" (yellow).

PR #214, opened 2020-09-08
Update docs/_static/api.json

quba42:
[noissue]

This will become unnecessary once the work on https://pulp.plan.io/issues/7229
(and related issues) is done (for all plugins)!
This PR will manually check in the current state one final time (hopefully)!

Generated from an instance running pulpcore 3.6.3 and pulp_deb 2.6.1!
When using pulp_deb master instead, the same output is generated.

### DIFF
--- a/docs/_static/api.json
+++ b/docs/_static/api.json
@@ -18,294 +18,11 @@
         }
     },
     "paths": {
-        "/pulp/api/v3/artifacts/": {
-            "get": {
-                "operationId": "pulp_api_v3_artifacts_list",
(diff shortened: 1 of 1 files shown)


docs/_static/api.json
```diff
@@ -18,294 +18,11 @@
         }
     },
     "paths": {
-        "/pulp/api/v3/artifacts/": {
```
quba42, line 21: Apparently, the previously committed `api.json` contained a lot of pulpcore endpoints that are now removed. I am not sure why this was ever the case, but it strikes me as "correct" to drop it now.

@dkliban Perhaps you can enlighten me?

mdellweg: If it's about endpoints not specified in the plugin, that's expected.

Enlightenment: https://github.com/pulp/pulpcore/pull/840

mdellweg (approved):

PR #215, opened 2020-09-24
Update plugin template

quba42:
Latest changes from the 2020.09.23 tag in the plugin template.

### DIFF
--- a/template_config.yml
+++ b/template_config.yml
@@ -22,9 +22,11 @@ plugin_dash: pulp-deb
 plugin_dash_short: deb
 plugin_name: pulp_deb
 plugin_snake: pulp_deb
+publish_docs_to_pulpprojectdotorg: false
 pulp_settings: null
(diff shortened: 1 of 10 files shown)


template_config.yml
```diff
@@ -22,9 +22,11 @@ plugin_dash: pulp-deb
 plugin_dash_short: deb
 plugin_name: pulp_deb
 plugin_snake: pulp_deb
+publish_docs_to_pulpprojectdotorg: false
```
mdellweg, line 25: Talk to Dennis if you are interested in this.

mdellweg (approved):

mdellweg:
Have you considered activating S3 tests. That might become interesting for the sync pipeline.

template_config.yml
quba42: I absolutely plan to, but I will probably do that in a follow up PR.

PR #218, opened 2020-09-28
Release script with pulpcore requirement

fao89:
[noissue]

### DIFF
--- a/.travis/release.py
+++ b/.travis/release.py
@@ -15,6 +15,7 @@ from pathlib import Path
 from git import Repo
 from redminelib import Redmine
 from redminelib.exceptions import ResourceAttrError
+from sh import sed
 
 
 REDMINE_URL = "https://pulp.plan.io"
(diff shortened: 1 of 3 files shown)


.travis/release.py
```diff
@@ -136,21 +137,30 @@ def validate_redmine_data(redmine_query_url, redmine_issues):
 git.checkout("HEAD", b=f"release_{release_version}")
 
 # First commit: changelog
+if "pulpcore" not in release_path:
+    pulpcore_requirement = f"pulpcore>={lower_pulpcore_version},<{upper_pulpcore_version}"
```
quba42, line 141: From a ReStructuredText formatting point of view, I would prefer to place this string in double backticks. (Not sure if backticks need to be escaped somehow in the context of the `sed()` function.)

See https://github.com/pulp/pulp_deb/pull/219#discussion_r496774407 for an example of the end result I am aming for.

.travis/release.py
```diff
@@ -136,21 +137,30 @@ def validate_redmine_data(redmine_query_url, redmine_issues):
...
-with open(f"{plugin_path}/requirements.txt", "wt") as setup_file:
-    for line in setup_lines:
-        if "pulpcore" in line and "pulpcore" not in release_path:
-            line = f"pulpcore>={lower_pulpcore_version},<{upper_pulpcore_version}\n"
-
-        setup_file.write(line)
+if "pulpcore" not in release_path:
+    sed(
+        "-i",
+        f"s/pulpcore.*/pulpcore>={lower_pulpcore_version},<{upper_pulpcore_version}/",
```
quba42, line 161: Why can't this reuse the `pulpcore_requirement` variable from line 141?

quba42 (commented):
I can confirm that the script (in its current form) works for Y releases without any issues/changelog entries.

According to my local test, it does not yet add the same string for Y releases that do have issues (and hence changelog entries).

I may have another look at this soon. :wink:

.travis/release.py
fao89: actually, I forgot to add it here

.travis/release.py
fao89: I've addressed the requested changes and tested it:
https://github.com/pulp/pulp_deb/compare/master...fao89:release_2.7.0

PR #219, opened 2020-09-29
Release 2.7.0

quba42:
This is a pure compatibility release.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,15 @@ Changelog
 
 .. towncrier release notes start
 
+2.7.0 (2020-09-29)
+==================
+
+Compatible with: ``pulpcore>=3.7,<3.9``
+
+
+----
+
+
 2.6.1 (2020-09-03)
 ==================
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
-pulpcore>=3.6
+pulpcore>=3.7,<3.9
 python-debian>=0.1.36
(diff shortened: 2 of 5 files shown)


requirements.txt
```diff
@@ -1,2 +1,2 @@
-pulpcore>=3.6
+pulpcore>=3.7,<3.9
```
mdellweg: If you didn't introduce a change that needs to push the lower bound here, you can make that ">=3.6,<3.9".
If there was such a change, imho that change should have bumped the requirement...

mdellweg (approved):

requirements.txt
quba42: Thanks for the hint.
I have just failed repeatedly to get the `pulp_deb` tests (with this branch checked out) to work when the pulpcore `3.6` branch is checked out.
As a result, I think I will raise the floor to 3.7 for this release.

That being said, I also plan to mail the mailing list about your general point, so that I may learn and improve for future releases. :wink:

mdellweg (approved):

CHANGES.rst
```diff
@@ -11,6 +11,15 @@ Changelog
 
 .. towncrier release notes start
 
+2.7.0 (2020-09-29)
+==================
+
+Compatible with: ``pulpcore>=3.7,<3.9``
```
quba42, line 17: I manually added the backticks to the result from using the script from https://github.com/pulp/pulp_deb/pull/218.

PR #220, opened 2020-09-29
Update the release documentation

quba42:
[noissue]

This reflects the lessons learned from the 2.7.0 release.
See: https://github.com/pulp/pulp_deb/pull/219

### DIFF
--- a/docs/plugin_maintenance.rst
+++ b/docs/plugin_maintenance.rst
@@ -126,11 +126,18 @@ For a ``Y`` release, perform the following steps (the example assumes we are rel
    .. code-block:: none
 
       workon pulp_release
-      python .travis/release.py minor --lower 3.6 --upper 3.7
+      python .travis/release.py minor --lower 3.6 --upper 3.8
 
    The ``--lower`` and ``--upper`` parameters give the pulpcore version range that the release should be compatible with.
-   Currently, each release is pegged to exactly one pulpcore-``Y`` release.
-   This is due to change with the pulpcore ``3.7`` release.
+   The example would result in a declared compatibility range of ``pulpcore>=3.6,<3.8``.
+   That is, for all pulpcore ``3.6.*`` and ``3.7.*`` releases.
+
+   .. note::
+      pulpcore will introduce breaking changes to the plugin API over a cycle of two ``Y`` releases.
+      Affected functions will be deprecated in some ``Y`` release, but only be removed with the next ``Y`` release after that.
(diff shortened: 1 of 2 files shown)


docs/plugin_maintenance.rst
```diff
@@ -126,11 +126,17 @@ For a ``Y`` release, perform the following steps (the example assumes we are rel
...
 
       workon pulp_release
-      python .travis/release.py minor --lower 3.6 --upper 3.7
+      python .travis/release.py minor --lower 3.6 --upper 3.8
 
    The ``--lower`` and ``--upper`` parameters give the pulpcore version range that the release should be compatible with.
-   Currently, each release is pegged to exactly one pulpcore-``Y`` release.
-   This is due to change with the pulpcore ``3.7`` release.
+   The example would result in a declared compatibility range of ``pulpcore>=3.6,<3.8``.
+   That is, for all pulpcore ``3.6.*`` releases.
```
quba42, line 133:
```suggestion
   That is, for all pulpcore ``3.6.*`` and ``3.7.*`` releases.
```

docs/plugin_maintenance.rst
```diff
@@ -126,11 +126,17 @@ For a ``Y`` release, perform the following steps (the example assumes we are rel
...
+
+   .. note::
+      pulpcore has now adopted a release cycle were backwards incompatible changes (or "removals") may be performed every second ``Y`` release.
```
mdellweg, line 136: Trying to read this in half a year...
```suggestion
      pulpcore has a release cycle were backwards incompatible changes (or "removals") may be performed every second ``Y`` release.
```

docs/plugin_maintenance.rst
```diff
@@ -126,11 +126,17 @@ For a ``Y`` release, perform the following steps (the example assumes we are rel
...
+      That is, a breaking change will be announced as a "deprecation" in the next pulpcore ``Y`` release, but only be included with the ``Y`` release after that one.
```
mdellweg, line 137: Not sure if it makes this easier to understand, but the way i see it is that a breaking change needs to be split into pure additions to happen now and pure removals (that are pre announced by deprecations) happening with the next Y.

quba42: Thanks, I am definitely still seeking clarifications on how the process will work exactly (which may be why my formulation here is still a bit vague).

quba42: I have reformulated the text some more, and linked the pulpcore plugin API deprecation policy: https://docs.pulpproject.org/pulpcore/plugins/plugin-writer/concepts/index.html#plugin-api-stability-and-deprecation-policy

PR #222, opened 2020-09-30
Adds a stage for publishing docs to docs.pulpproject.org/pulp_deb/

dkliban:
[noissue]


mdellweg:
Can `docs/_static/api.json` be removed with this?

quba42:
> Can `docs/_static/api.json` be removed with this?

My understanding is that once the new docs are live and everything (like URLs) has been updated to reflect that, the deployment mechanism for the old docs, along with `docs/_static/api.json` can be removed.

quba42:
@dkliban Is there a better way of reviewing this than just merging it and checking if everything worked after the next release?

It looks sane enough to me, but I don't know exactly what it should look like.

mdellweg:
It looks to me like there are only files generated by the pulp plugin_template in this PR. So maybe generating yourself and seeing no change, is all we can do now.

dkliban:
@quba42 You will see it work whenever the cron job runs. The nightly docs should get published to https://docs.pulpproject.org/pulp_deb/en/master/nightly/

docs/Makefile
```diff
@@ -42,6 +44,17 @@ help:
 clean:
 	-rm -rf $(BUILDDIR)/*
 
+diagrams:
```
quba42, line 47: I am not sure I fully understand what the `diagrams` target is good for.
If I understand it correctly it will only do things if there is a folder called `diagrams_src`, which is currently not the case (at least not in source).
I have the following questions:

1. Are we anticipating that there will be such a folder? Or is it perhaps a folder that only exists for the Travis pipeline?
2. If the make target where to "build" something, it would end up in the `_diagrams` folder, so should this folder be added to the `clean` target?
3. The Makefile change, does not appear to come from the plugin Template, so should it be added there?

quba42 (commented):
I was locally able to generate the files from the plugin template the same way.
So once my final three questions on the Makefile changes are answered I am looking forward to merging this soon!

docs/Makefile
dkliban: 1. Pulpcore is currently the only project that has diagrams, but I anticipate other projects will want to make use of plantuml to add workflow diagrams.
2. We should add _diagrams directory to the clean target. 
3. This change is there, but is only available when bootstrapping a plugin for the first time. https://github.com/pulp/plugin_template/blob/master/templates/docs/docs/Makefile.j2

PR #225, opened 2020-10-15
Support flat repo format

ThikaXer:
fixes #7502
https://pulp.plan.io/issues/7502

### DIFF
--- a/CHANGES/7502.feature
+++ b/CHANGES/7502.feature
@@ -1,0 +1,1 @@
+Added the capability to synchronize repositories using "flat repository format".
--- a/pulp_deb/app/tasks/publishing.py
+++ b/pulp_deb/app/tasks/publishing.py
@@ -172,7 +172,7 @@ class _ComponentHelper:
         for architecture in self.parent.architectures:
             package_index_path = os.path.join(
                 "dists",
-                self.parent.distribution,
+                self.parent.distribution.strip("/"),
                 self.plain_component,
                 "binary-{}".format(architecture),
                 "Packages",
@@ -276,7 +276,7 @@ class _ReleaseHelper:
             component.finish()
         # Publish Release file
         self.release["Components"] = " ".join(self.components.keys())
-        release_dir = os.path.join("dists", self.distribution)
+        release_dir = os.path.join("dists", self.distribution.strip("/"))
         release_path = os.path.join(release_dir, "Release")
         os.makedirs(os.path.dirname(release_path), exist_ok=True)
         with open(release_path, "wb") as release_file:
--- a/pulp_deb/app/tasks/synchronizing.py
+++ b/pulp_deb/app/tasks/synchronizing.py
@@ -409,10 +409,14 @@ class DebFirstStage(Stage):
     async def _handle_distribution(self, distribution):
         log.info('Downloading Release file for distribution: "{}"'.format(distribution))
         # Create release_file
+        if distribution[-1] == "/":
+            release_file_dir = distribution.strip("/")
+        else:
+            release_file_dir = os.path.join("dists", distribution)
         release_file_dc = DeclarativeContent(
             content=ReleaseFile(distribution=distribution),
             d_artifacts=[
-                self._to_d_artifact(os.path.join("dists", distribution, filename))
+                self._to_d_artifact(os.path.join(release_file_dir, filename))
                 for filename in ["Release", "InRelease", "Release.gpg"]
             ],
         )
@@ -432,7 +436,11 @@ class DebFirstStage(Stage):
             )
             await self.put(release_architecture_dc)
         # Parse release file
-        log.info('Parsing Release file for release: "{}"'.format(release_file.codename))
+        log.info(
+            'Parsing Release file for release with codename="{}" and distribution="{}"'.format(
+                release_file.codename, distribution
+            )
+        )
         release_file_dict = deb822.Release(release_file.main_artifact.file)
         # collect file references in new dict
         file_references = defaultdict(deb822.Deb822Dict)
@@ -501,9 +509,13 @@ class DebFirstStage(Stage):
     ):
         # Create package_index
         release_base_path = os.path.dirname(release_file.relative_path)
-        package_index_dir = os.path.join(
-            release_component.plain_component, infix, "binary-{}".format(architecture)
-        )
+        if release_file.distribution[-1] == "/":
+            # Flat repo format
+            package_index_dir = ""
+        else:
+            package_index_dir = os.path.join(
+                release_component.plain_component, infix, "binary-{}".format(architecture)
+            )
         d_artifacts = []
         for filename in ["Packages", "Packages.gz", "Packages.xz", "Release"]:
             path = os.path.join(package_index_dir, filename)
@@ -537,7 +549,7 @@ class DebFirstStage(Stage):
         package_futures = []
         for package_paragraph in deb822.Packages.iter_paragraphs(package_index.main_artifact.file):
             try:
-                package_relpath = package_paragraph["Filename"]
+                package_relpath = os.path.normpath(package_paragraph["Filename"])
                 package_sha256 = package_paragraph["sha256"]
                 if package_relpath.endswith(".deb"):
                     package_class = Package
--- a/pulp_deb/tests/functional/api/test_publish_flat_repo_format.py
+++ b/pulp_deb/tests/functional/api/test_publish_flat_repo_format.py
@@ -1,0 +1,236 @@
+# coding=utf-8
+"""Tests that sync deb plugin repositories."""
+import unittest
+import os
+
+from pulp_smash import config
+
+from pulp_smash.pulp3.utils import (
+    gen_repo,
+    get_content,
+    delete_orphans,
+    get_versions,
+    gen_distribution,
+    download_content_unit,
+)
+
+from pulp_deb.tests.functional.constants import (
+    DEB_FLAT_REPO_FIXTURE_URL,
+    DEB_RELEASE_FILE_NAME,
+    DEB_RELEASE_NAME,
+    DEB_RELEASE_COMPONENT_NAME,
+    DEB_PACKAGE_INDEX_NAME,
+)
+from pulp_deb.tests.functional.utils import set_up_module as setUpModule  # noqa:F401
+from pulp_deb.tests.functional.utils import (
+    gen_deb_remote,
+    monitor_task,
+    deb_remote_api,
+    deb_repository_api,
+    deb_apt_publication_api,
+    deb_verbatim_publication_api,
+    deb_distribution_api,
+)
+
+from pulpcore.client.pulp_deb import (
+    RepositorySyncURL,
+    DebAptPublication,
+    DebVerbatimPublication,
+)
+
+
+class FlatRepoSyncTestCase(unittest.TestCase):
+    """Sync a repository with the deb plugin."""
+
+    def _publication_extra_args(self, modus):
+        if modus == "verbatim":
+            return {}
+        else:
+            return {modus: True}
+
+    def setUp(self):
+        """Cleanup."""
+        delete_orphans()
+
+    def test_publish_flat_repo_structured(self):
+        """Test publishing flat repository with distribution /."""
+        expected_values = {
+            "distribution": "/",
+            "codename": "ragnarok",
+            "suite": "mythology",
+            "components": ["asgard"],
+            "release_file_folder_sync": "",
+            "release_file_folder_dist": "dists",
+            "package_index_paths_sync": ["Packages"],
+            "package_index_paths_dist": ["dists/asgard/binary-ppc64/Packages"],
+        }
+        self.do_publish(expected_values, "structured")
+
+    def test_publish_flat_repo_simple(self):
+        """Test publishing flat repository with distribution /."""
+        expected_values = {
+            "distribution": "/",
+            "codename": "ragnarok",
+            "suite": "mythology",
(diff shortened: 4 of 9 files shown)


quba42:
The following is a check list for what needs to happen before this PR is fully ready:

- [ ] Provide the requested functionality
- [ ] Add a flat repository format fixture to https://github.com/pulp/pulp-fixtures
- [ ] Use the new fixture for a full sync + publish + distribute test (there should be existing tests that do this for the existing fixtures that can be emulated).
- [ ] Add this new feature to the feature overview documentation.
- [ ] Update the documentation for how users should set the `distribution` field in their remotes (both API and workflow docs). We should explicitly mention that this field uses the same syntax as `/etc/apt/sources.list` entries.

digger18:
Any update on this PR?
Thanks in advance!

snirkatriel:
+1 👆

ThikaXer:
Fixtures have been created in https://github.com/pulp/pulp-fixtures/pull/199 but have not yet been merged (that's why Travis not yet running through I think). Quba42 is on holiday this week and thus not reviewing.
Test has already been included in test_publish_flat_repo_format.py
I'm not that sure where in the documentation I need to mention this feature...

ThikaXer:
I would appreciate any reccomendation on or reviews of the code, as I'm still very new to this! Thanks in advance :)

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -431,7 +439,7 @@ async def _handle_distribution(self, distribution):
             )
             await self.put(release_architecture_dc)
         # Parse release file
-        log.info('Parsing Release file for release: "{}"'.format(release_file.codename))
+        log.info('Parsing Release file for distribution: "{}"'.format(distribution))
```
mdellweg, line 442:
```suggestion
        log.info('Parsing Release file for release: "{}" at distribution: "{}"'.format(release_file.codename, distribution))
```
You never know which information might be really needed in the logfiles...

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -500,9 +508,12 @@ async def _handle_package_index(
     ):
         # Create package_index
         release_base_path = os.path.dirname(release_file.relative_path)
-        package_index_dir = os.path.join(
-            release_component.plain_component, infix, "binary-{}".format(architecture)
-        )
+        if self.distributions[0][-1] == "/":
```
mdellweg, line 511: Maybe add the "flat repo" comment here too.

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -500,9 +508,12 @@ async def _handle_package_index(
...
+            package_index_dir = ""
```
mdellweg, line 512: I guess, those "flat repos" only support "*.deb" and not "*.udeb"?

pulp_deb/tests/functional/api/test_publish_flat_repo_format.py
```diff
@@ -0,0 +1,194 @@
...
+
+    def test_publish_flat_repo_structured(self):
+        """Test publishing flat repository with distribution debian-flat/."""
+        expected_values = {
+            "distribution": "debian-flat/",
+            "codename": "ragnarok-flat",
+            "suite": "mythology",
+            "components": ["asgard-flat"],
+            "release_file_folder_sync": "debian-flat/",
+            "release_file_folder_dist": "dists/debian-flat/",
```
mdellweg, line 63: This is because we republish them "not flat"?

pulp_deb/app/tasks/synchronizing.py
ThikaXer: Hm, I'm not sure, it's not specified anywhere (here the definition from debian flat repos https://wiki.debian.org/DebianRepository/Format#Flat_Repository_Format ). I would assume that *.udebs should be ssupported as well, but still they will be just in the same directory as everything else... What do you think?

pulp_deb/tests/functional/api/test_publish_flat_repo_format.py
ThikaXer: Here, I was not sure how to proceed. Maybe you have some suggestions or ideas.
the problem is, that on one side, structured publishing should look more or less like the original (as I understand it). Here, we get a conflict with the style of where Release file and Packages files are: in structured we want to introduce the architecture all if not yet there. So for flat repositories how should one do that? 
On the other side, having some structured repository (ie folder structure in dists/ and pool/) is quite nice, too, maybe even nicer than just putting everything in one directory as the flat repo does. 
So in the end, I left the structured publisher to do a "structured" but not flat publishing. If one really wants to keep the flat structure, the verbatim publisher can be used.

What do you think about this problem?

pulp_deb/app/tasks/synchronizing.py
ThikaXer: Done.

pulp_deb/app/tasks/synchronizing.py
ThikaXer: Done.

pulp_deb/tests/functional/api/test_publish_flat_repo_format.py
mdellweg: I was asking to understand. If you use the structured publishing, it should be structured (as in looking as if it were an official upstream debian repository with different packages).
There is no need to try to mimic the structure of how the repository was synced if you need to republish anyway.
Verbatim is really for creating a bit by bit identical mirror that is preserving valid signatures on the metadata.

ThikaXer: Ah, then this should be ok, shouldn't it?

mdellweg: Yes. Sorry if i confused you.

ThikaXer: No prob, thanks for the input ;)

digger18:
Can this be merged? we need it :)

quba42:
> Can this be merged? we need it :)

I will start final review now. I think there is a good chance this will be merged this week. I plan to create a release as soon as this is merged. The next release will have version 2.8.0 and is needed anyway before pulpcore 3.9.0 is released at the end of the month.

pulp_deb/app/tasks/synchronizing.py
quba42: A Debian repo in flat repo format supports at most a single `Packages` and a single `Sources` index. Unless it is possible to reference a mix of both `.deb` and `.udeb` packages from a single `Packages` file (I don't think so?!) it is therefore not possible to have `.udeb` packages? Or maybe it is possible to have a flat repo format repo containing exclusively `.udeb` packages?

Anyway, from a practical point of view, I think we can shelve this discussion. If anyone ever finds a flat repo with `.udeb` packages in the wild, they can open a new issue then. For now, I say the `pulp_deb` plugin supports `.deb` packages only for flat repos. Most people who are expert enough to create installer packages, are probably also expert enough to use full repo structure...

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -537,7 +554,7 @@ async def _handle_package_index(
         package_futures = []
         for package_paragraph in deb822.Packages.iter_paragraphs(package_index.main_artifact.file):
             try:
-                package_relpath = package_paragraph["Filename"]
+                package_relpath = os.path.normpath(package_paragraph["Filename"])
```
quba42, line 557: Just tested out this function on various examples.
This is a pretty useful function that we might use in place of some of our `strip` functions as well. (Although it might not do exactly what we want in all cases).

quba42 (changes requested):
So, I am unsure if we want to permit users giving a distribution starting with `./`.
I am also not entirely sure if it makes sense for us to create release component content units during the sync of the flat repo case. (I guess we could argue that flat repos implicitly have a single component, and this would have the advantage that the structured publisher will be able to publish flat repos.)

I would also like to add some more documentation. (If you don't mind I can do that, and then you can review it.)

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -501,9 +513,14 @@ async def _handle_package_index(
     ):
         # Create package_index
         release_base_path = os.path.dirname(release_file.relative_path)
-        package_index_dir = os.path.join(
-            release_component.plain_component, infix, "binary-{}".format(architecture)
-        )
+        if self.distributions[0][-1] == "/":
+            # in case te repo is in flat repo format leave out the dir
+            # "dists/infix/binary-architecture"
+            package_index_dir = ""
```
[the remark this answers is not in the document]
ThikaXer: Here, I think  it really should be just the empty string as the Packages file is referenced relative to the Release file, isn't it? I just tested with the `package_index_dir = release_file.distribution.strip("/")` but it could not locate the Packages file. I tested it locally with the `nest/fjalar/` distribution but removed the Packages files in the root.

CHANGES/7502.feature
```diff
@@ -0,0 +1 @@
+Support flat repository format
```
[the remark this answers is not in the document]
ThikaXer: Done

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -409,10 +409,18 @@ def _to_d_artifact(self, relative_path, data=None):
     async def _handle_distribution(self, distribution):
         log.info('Downloading Release file for distribution: "{}"'.format(distribution))
         # Create release_file
+        if distribution == "/" or distribution == "./":
```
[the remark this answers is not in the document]
ThikaXer: Done

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -432,7 +440,11 @@ async def _handle_distribution(self, distribution):
             )
             await self.put(release_architecture_dc)
         # Parse release file
-        log.info('Parsing Release file for release: "{}"'.format(release_file.codename))
+        log.info(
+            'Parsing Release file for release: "{}" at distribution: "{}"'.format(
```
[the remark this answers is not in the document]
ThikaXer: Done

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -501,9 +513,14 @@ async def _handle_package_index(
     ):
         # Create package_index
         release_base_path = os.path.dirname(release_file.relative_path)
-        package_index_dir = os.path.join(
-            release_component.plain_component, infix, "binary-{}".format(architecture)
-        )
+        if self.distributions[0][-1] == "/":
```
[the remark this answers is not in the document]
ThikaXer: Done

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -501,9 +513,14 @@ async def _handle_package_index(
...
+            # in case te repo is in flat repo format leave out the dir
```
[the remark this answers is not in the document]
ThikaXer: Done

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -501,9 +513,14 @@ async def _handle_package_index(
...
+            # "dists/infix/binary-architecture"
```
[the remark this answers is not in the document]
ThikaXer: Done

pulp_deb/app/tasks/synchronizing.py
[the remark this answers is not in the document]
quba42: You are right, these are the paths as found in the Release file, not the paths relative to the repo root. My bad!

pulp_deb/app/tasks/publishing.py
```diff
@@ -172,7 +172,7 @@ def __init__(self, parent, component):
         for architecture in self.parent.architectures:
             package_index_path = os.path.join(
                 "dists",
-                self.parent.distribution,
+                self.parent.distribution.strip("./"),
```
quba42, line 175:
```suggestion
                self.parent.distribution.strip("/"),
```

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -409,10 +409,18 @@ def _to_d_artifact(self, relative_path, data=None):
     async def _handle_distribution(self, distribution):
         log.info('Downloading Release file for distribution: "{}"'.format(distribution))
         # Create release_file
+        if distribution == "/":
+            # if distribution = "/" problem with os.path.join
+            release_file_dir = ""
+        elif distribution[-1] == "/":
+            # in case te repo is in flat repo format leave out the dir "dists"
+            release_file_dir = distribution.strip(".")
```
quba42, line 417:
```suggestion
            release_file_dir = distribution.strip("/")
```

quba42 (changes requested):
Two more strip corrections and then we should be able to merge. (At least I think those should be changed?)

pulp_deb/app/tasks/publishing.py
ThikaXer: Done

pulp_deb/app/tasks/synchronizing.py
ThikaXer: Done

quba42 (approved):

pulp_deb/app/tasks/synchronizing.py
ThikaXer: I think we can even reduce the code a bit more. With strip() we have the possibility to combine both cases of flat repo format

PR #226, opened 2020-10-30
Ignore architectures with missing Packages file

ThikaXer:
fixes #7467


CHANGES/7467.feature
```diff
@@ -0,0 +1 @@
+Allows pulp_deb to synchronise a repository with missing architectures, ie in the Release file architectures are specified but the corresponding Packages files are missing.
```
mdellweg, line 1: Changes should be formulated in past tense.
This line is what will be compiled into the changelog.

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -14,6 +14,8 @@
 from debian import deb822
 
 from urllib.parse import urlparse, urlunparse
+from urllib.error import URLError, HTTPError
+from urllib.request import urlopen
```
mdellweg, line 18: you should use aio* libraries in the sync code.
Especially for things that may stall the event loop.

mdellweg (changes requested):

CHANGES/7467.feature
ThikaXer: Fixed.

pulp_deb/app/tasks/synchronizing.py
ThikaXer: I tried to understand the async and the aiohttp library, but somehow I seem to miss something (or everything). How can I know in which ClientSession I'm in for example? I need a ClientSession to do a request to a website, don't I?
Or is it ok to say the following?
```
async def download():
   Do the urllib.request here

Later in synchronize:
task = asyncio.create_task(download())
await task
´´´
Maybe this just pushing the problem along?

Comment (mdellweg) [reply pulp_deb/app/tasks/synchronizing.py]: Yes, this is still the same problem. This whole sync pipeline is written in async python [0]. That means the code should not incorporate any blocking call ever. Downloader that are safe to use as a coroutine are described in [1].

[0] https://docs.python.org/3/library/asyncio.html
[1] https://docs.pulpproject.org/pulpcore/plugins/api-reference/download.html#parallel-downloading

Comment (ThikaXer) [reply pulp_deb/app/tasks/synchronizing.py]: Thanks! I tried again with asyncio... it took me quite a while to understand where the problem is. Is the new version better?

Comment (mdellweg) [reply pulp_deb/app/tasks/synchronizing.py]: With respect to not using requests, definitely. Somehow i have the impression this should be a lot less complex to implement. Trying to understand that piece of code once more:
You see there is the `NoPackageIndexFile` exception that is emitted in two places in `DebUpdatePackageIndexAttributes`. This is the places the code already knows about the condition you are after. And given the asynchronous nature of that code i'd suggest you try to recover from there by catching the exception, resolving the `d_content' to `None`, and then (back in the first stage) ignoring the futures that resolved to `None`.

Comment (mdellweg) [reply pulp_deb/app/tasks/synchronizing.py]: Maybe you even want to make the reaction to those missing files configurable.

File: pulp_deb/app/tasks/synchronizing.py [hunk 3]
```diff
@@ -283,29 +285,46 @@ async def run(self):
...
-                        raise NoPackageIndexFile(relative_dir=relative_dir)
-
-                    content = d_content.content
-                    if not [
-                        da for da in d_content.d_artifacts if da.artifact.sha256 == content.sha256
-                    ]:
-                        # No main_artifact found, uncompress one
-                        filename = _uncompress_artifact(d_content.d_artifacts, relative_dir)
+                        # no packages artifact found
+                        with NamedTemporaryFile(delete=False) as temp_file:
```
mdellweg, line 289: I don't think you should replace it with an empty file. Rather just resolve the d_content to None and don't even send it down to the next pipeline step.

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -283,31 +285,34 @@ async def run(self):
...
-                            content.relative_path,
-                            d_content.d_artifacts[0].remote,
-                        )
-                        d_content.d_artifacts.append(da)
-                        da.artifact.save()
-                        log.info(
-                            "*** Expected: {} *** Uncompressed: {} ***".format(
-                                content.sha256, da.artifact.sha256
+                        # no packages artifact found
+                        d_content = None
```
ThikaXer, line 289: When setting d_content to None and later in DebFirstStage (533-535)
```
        if package_index.main_artifact.file.size == 0: #no package_index
            log.info("PackageIndex created but was None")
            return
```
the program seems to run in a non ending loop. How can I find out where the problem is this time?

mdellweg: if think it is 
```
d_content.content = None
d_content.resolve()
```

The pipeline is heavily using this pattern:
https://docs.pulpproject.org/pulpcore/plugins/api-reference/stages.html#pulpcore.plugin.stages.ResolveContentFutures

mdellweg: So i think the best way to handle this is
1. start from scratch
2. replace the `raise NoPackageIndexFile(relative_dir=relative_dir)` with
```
d_content.content = None
d_content.resolve()
continue
```
3. handle the None case in the first stage
4. consider an option to decide about failing or ignoring

ThikaXer: Thanks! Here is the next try ;)
Right now, I only included steps 1-3. Concerning step 4, I would have to think a bit more about where to let the user set the handle of failing or ignoring.

CHANGES/7467.feature
```diff
@@ -0,0 +1 @@
+Allowed pulp_deb to synchronise a repository with missing architectures, ie in the Release file architectures are specified but the corresponding Packages files are missing.
```
mdellweg, line 1: I think, a missing Packages file is indicating, that a combination of component-architecture is missing.

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -283,8 +283,9 @@ async def run(self):
                 if isinstance(d_content.content, PackageIndex):
                     relative_dir = os.path.dirname(d_content.content.relative_path)
                     if not d_content.d_artifacts:
-                        raise NoPackageIndexFile(relative_dir=relative_dir)
-
+                        d_content.content = None
+                        d_content.resolve()
+                        continue
```
mdellweg, line 288: :+1:
Sorry, it should have been more obvious to me in the first place.

ThikaXer: No problem. I missed the d_content.resolve() and got confused :)

pulp_deb/app/tasks/synchronizing.py
ThikaXer: Done.

pulp_deb/app/tasks/synchronizing.py
mdellweg: Yeah, i just saw that https://docs.pulpproject.org/pulpcore/plugins/plugin-writer/concepts/sync_pipeline/sync_pipeline.html#multiple-level-discovery and https://docs.pulpproject.org/pulpcore/plugins/api-reference/stages.html#pulpcore.plugin.stages.ResolveContentFutures may not be enough to under stand that pattern.

Given your recent insight into the "looping pattern" of the stages api. Would you be willing to improve the documentation there? Add some cross references and start a paragraph with "If you need to drop DeclarativeContent prematurely from the pipeline...".

pulp_deb/tests/functional/api/test_sync.py
```diff
@@ -130,31 +130,6 @@ def test_invalid_distribution(self):
...
-        The needed Package indices associated with the relevant content unit are missing.
-        The Release file next to the missing Package indices is retained.
-        Assert that the relevant exception is thrown.
-        """
-        with self.assertRaises(PulpTaskError) as exc:
-            self.do_test(url=DEB_INVALID_FIXTURE_URL, architectures="armeb")
-        error = exc.exception.task.error
-        self.assertIn("No suitable Package index file", error["description"])
-        self.assertIn("armeb", error["description"])
-
```
mdellweg, line 157: I am still not entirely convinced that this is covering all usecases. If i specify, that i want to sync a set of components with a set of architectures, and one of the combinations is missing i might consider that an error.
Should we have an option on the remote to switch this between warning and failing?
@quba42 Any thoughts?

pulp_deb/tests/functional/api/test_sync_missing_architecture.py
```diff
@@ -0,0 +1,173 @@
...
+    DEB_MISSING_ARCH_DISTS_FIXTURE_URL,
+    DEB_RELEASE_FILE_NAME,
+    DEB_RELEASE_NAME,
+    DEB_RELEASE_COMPONENT_NAME,
+    DEB_PACKAGE_INDEX_NAME,
+)
+from pulp_deb.tests.functional.utils import set_up_module as setUpModule  # noqa:F401
+from pulp_deb.tests.functional.utils import (
+    gen_deb_remote,
+    monitor_task,
```
mdellweg, line 27: There is a (maybe better) version of this tool in pulp-smash.pulp3.bindings now.
You should probably migrate that eventually. (Not this PR)

pulp_deb/tests/functional/api/test_sync_missing_architecture.py
```diff
@@ -0,0 +1,173 @@
...
+        self.addCleanup(deb_remote_api.delete, remote.pulp_href)
+
+        # Sync the repository:
+        repository_sync_data = RepositorySyncURL(remote=remote.pulp_href)
+        sync_response = deb_repository_api.sync(repo.pulp_href, repository_sync_data)
+        monitor_task(sync_response.task)
+        repo = deb_repository_api.read(repo.pulp_href)
+        version_hrefs = tuple(ver["pulp_href"] for ver in get_versions(repo.to_dict()))
+
+        self.assertIsNotNone(repo.latest_version_href)
```
mdellweg, line 96: latest version will never be none, be cause an empty repository will still have version 0.
So this assertion does not help anything.

pulp_deb/tests/functional/api/test_sync_missing_architecture.py
```diff
@@ -0,0 +1,173 @@
...
+        self.assertEqual(expected_values["codename"], release["codename"])
+        self.assertEqual(expected_values["suite"], release["suite"])
+
+        components = get_content(
+            repo=publication.to_dict(), version_href=publication.repository_version
+        )[DEB_RELEASE_COMPONENT_NAME]
+
+        self.assertEqual(len(expected_values["components"]), len(components))
+        for component in components:
+            self.assertIn(component["component"], expected_values["components"])
```
mdellweg, line 135: This looks like a set comparison (`set(...) == set(...)`)

pulp_deb/tests/functional/api/test_sync_missing_architecture.py
```diff
@@ -0,0 +1,173 @@
...
+
+        package_indecies = get_content(
```
mdellweg, line 137: Typo: indices

CHANGES/7467.feature
ThikaXer: Done.

CHANGES/7467.feature
```diff
@@ -0,0 +1 @@
+Allowed pulp_deb to synchronize a repository with missing Packages file for a architecture-component combination, even though it is specified in the Release file.
```
quba42, line 1: Since we are now planning to make it configurable per remote flag, I would suggest something like:
```suggestion
Users may set the ``ignore_missing_package_indices`` flag in their remotes, to allow the synchronization of repositories missing declared package indices.
```

pulp_deb/app/tasks/synchronizing.py
ThikaXer: Done.

pulp_deb/tests/functional/api/test_sync_missing_architecture.py
ThikaXer: Done.

pulp_deb/tests/functional/api/test_sync_missing_architecture.py
ThikaXer: It is now handled in a better way, I hope ;) Done

pulp_deb/tests/functional/api/test_sync_missing_architecture.py
ThikaXer: Done

CHANGES/7467.feature
ThikaXer: Done

CHANGES/7467.feature
```diff
@@ -0,0 +1 @@
+Users may set the ``ignore_missing_package_indices`` flag in their remotes, to allow for the synchronization of repositories missing declared package indices.
```
mdellweg, line 1: This should still be worded as a change.
Maybe: "Added an ``ignore_missing_package_indices`` flag to remotes, so users may..."

ThikaXer: Done. I rewrote the changes sentence :)

pulp_deb/tests/functional/api/test_sync.py
ThikaXer: Reverted this.

pulp_deb/tests/functional/api/test_sync_missing_architecture.py
quba42: Should I open an issue for that? Something like "Replace pulp_deb.tests.functional.utils.monitor_task with tool from pulp-smash.pulp3.bindings for tests"?

pulp_deb/app/serializers/remote_serializers.py
```diff
@@ -40,6 +40,17 @@ class AptRemoteSerializer(RemoteSerializer):
         allow_null=True,
     )
 
+    ignore_missing_package_indices = BooleanField(
+        help_text="By default, upstream repositories that declare architectures "
+        "and corresponding package indices in their Release files without "
+        "actually publishing them, will fail to synchronize.\n"
+        'Set this flag to allow the synchronization of such "partial mirrors" '
```
quba42, line 47: Should this be "set this flag to True", or is it obvious that "setting a flag" means "setting it to True"?

pulp_deb/app/serializers/remote_serializers.py
```diff
@@ -40,6 +40,17 @@ class AptRemoteSerializer(RemoteSerializer):
         allow_null=True,
     )
 
+    ignore_missing_package_indices = BooleanField(
+        help_text="By default, upstream repositories that declare architectures "
```
quba42, line 44: I will double check and report back whether literal newlines in the `help_text` will result in the desired output for the auto generated API docs.

We will also need to remember to update `docs/_static/api.json` (hopefully for the last time) with the next release.

quba42 (changes requested):
Looks very nearly done to me. I will merge as soon as the last few open threads are resolved.

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -283,8 +283,9 @@ async def run(self):
                 if isinstance(d_content.content, PackageIndex):
                     relative_dir = os.path.dirname(d_content.content.relative_path)
```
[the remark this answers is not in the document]
ThikaXer: Done.

pulp_deb/app/serializers/remote_serializers.py
ThikaXer: Done.

pulp_deb/app/serializers/remote_serializers.py
quba42: Ok, I just remembered local builds of the API docs don't actually work...
I guess we will just have to double check this post release. (At least I don't know of any better way to go about it.)

quba42 (approved):

pulp_deb/app/serializers/remote_serializers.py
mdellweg: Yeah it's kind of obvious. If it weren't a boolean it would be called a parameter, not a flag.

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -523,6 +524,14 @@ async def _handle_package_index(
         package_index = await self._create_unit(
             DeclarativeContent(content=content_unit, d_artifacts=d_artifacts)
         )
+        # Skip if empty packages files
```
mdellweg, line 527: This comment should probably move one line down.

quba42: Or maybe just removed? Not sure it still adds a lot of information in addition to:
```
https://github.com/pulp/pulp_deb/pull/226#discussion_r524229981
    return
```

quba42: In fact I think the "*empty* packages file" still refers to an earlier version of the PR.

mdellweg: Yes, the code explains itself. (At least to the interpreter, and comments in general tend to be out of sync with the actual code, so don't add unnecessary ones.)

ThikaXer: Comment removed :)

pulp_deb/app/tasks/synchronizing.py
ThikaXer: Maybe something like this?
https://github.com/pulp/pulpcore/pull/1028

mdellweg: Thanks for following up on this.

PR #227, opened 2020-11-19
Set default for DJANGO_SETTINGS_MODULE

ekohl:
This sets a default for DJANGO_SETTINGS_MODULE which means the user doesn't have to specify it anymore. Since it's a default, the user can still override it. This follows recent commits to pulpcore itself.

### DIFF
--- a/pulp_deb/tests/functional/setup_signing_service.py
+++ b/pulp_deb/tests/functional/setup_signing_service.py
@@ -14,6 +14,8 @@ if __name__ == "__main__":
         print("Usage: {} <path_to_signing_script>".format(sys.argv[0]))
         sys.exit(1)
 
+    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "pulpcore.app.settings")
+
     import django
 
     django.setup()


quba42:
Is this dependent on a particular pulpcore commit/release?
The commit message needs a `[noissue]` in the second line for the tests to run.

It would also be great if you could briefly bring me up to speed when and where the need for this arose/how you came across this? (Just so I can better understand the change.)

ekohl:
> Is this dependent on a particular pulpcore commit/release?

No

>  The commit message needs a `[noissue]` in the second line for the tests to run.

I can do that or open an issue. Which one do you prefer?

> It would also be great if you could briefly bring me up to speed when and where the need for this arose/how you came across this? (Just so I can better understand the change.)

The short summary is that right now you call the script as `DJANGO_SETTINGS_MODULE=pulpcore.app.settings /path/to/script.py` and it'll break if the user doesn't specify the env var. This sets a default for the env var so the user can call it as `/path/to/script.py`.

Related in pulpcore are https://pulp.plan.io/issues/7720 (https://github.com/pulp/pulpcore/commit/bfd0aa5058c18ad1eb6ba1a32e502d0a00309be7) and https://pulp.plan.io/issues/7179 (https://github.com/pulp/pulpcore/commit/6a2e7f304c31adcbf12d964753f639c5e3ea4cce).

Given the repetition, perhaps something that wraps `env.environ.setdefault(..) ; django.setup()` does make sense.

Btw, I would strongly recommend using the alternative: [custom management commands](https://docs.djangoproject.com/en/2.2/howto/custom-management-commands/). This means you can run `pulpcore-manager setup-signing-service`. It's easy to discover and nicely integrated in the ecosystem.

quba42:
> > The commit message needs a `[noissue]` in the second line for the tests to run.
> 
> I can do that or open an issue. Which one do you prefer?

Since this is not a core part of the plugin, I think a `[noissue]` would be fine. Unless you would like to announce this change to users in the release notes, in which case you would need to open a pulp_deb issue (https://pulp.plan.io/projects/pulp_deb/issues) and reference it. Since it is meant to help users more easily create signing services, that might be worthwhile... I leave it up to you, I am fine with it either way.

In general, I am fine with merging this change as soon as I see that the tests are still running. However, we will need to wait for (https://github.com/pulp/pulp_deb/pull/230) in order to run the tests, since testing on Travis is disabled.

quba42:
Right, the tests (using GitHub actions) have now been fixed in `master`.
If you add `[noissue]`/an issue reference in the commit message, and rebase onto current master everything should be good to go.

ekohl:
Rebased and added `[noissue]`.

quba42:
Thanks for this change, should make it easier to use this script for Katello testing (https://community.theforeman.org/t/pulp-deb-for-pulp3-in-katello/21331).

PR #228, opened 2020-11-23
Make codename an optional field

quba42:
fixes #7886
https://pulp.plan.io/issues/7886

### DIFF
--- a/pulp_deb/app/tasks/publishing.py
+++ b/pulp_deb/app/tasks/publishing.py
@@ -237,7 +237,7 @@ class _ReleaseHelper:
         if suite:
             self.release["Suite"] = suite
         self.release["Version"] = version
-        self.release["Codename"] = codename
+        self.release["Codename"] = codename or distribution.split("/")[0]
         self.release["Date"] = datetime.now(tz=timezone.utc).strftime("%a, %d %b %Y %H:%M:%S %z")
         self.release["Architectures"] = " ".join(architectures)
         self.release["Components"] = ""  # Will be set later
--- a/pulp_deb/app/tasks/synchronizing.py
+++ b/pulp_deb/app/tasks/synchronizing.py
@@ -257,7 +257,6 @@ class DebUpdateReleaseFileAttributes(Stage):
                     release_file.codename = release_file_dict["Codename"]
                     if "suite" in release_file_dict:
                         release_file.suite = release_file_dict.get("Suite")
-                    # TODO split of extra stuff e.g. : 'updates/main' -> 'main'
                     release_file.components = release_file_dict["Components"]
                     release_file.architectures = release_file_dict["Architectures"]
                     log.debug("Codename: {}".format(release_file.codename))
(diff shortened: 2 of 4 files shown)


pulp_deb/app/tasks/publishing.py
```diff
@@ -237,7 +237,10 @@ def __init__(
         if suite:
             self.release["Suite"] = suite
         self.release["Version"] = version
-        self.release["Codename"] = codename
+        if codename:
+            self.release["Codename"] = codename
+        else:
+            self.release["Codename"] = "fallback"
```
mdellweg, line 243:
```suggestion
        self.release["Codename"] = codename or "fallback"
```
Also why not fallback to the distribution?

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -254,9 +254,10 @@ async def run(self):
 
                     release_file.sha256 = release_file_artifact.sha256
                     release_file_dict = deb822.Release(release_file_artifact.file)
-                    release_file.codename = release_file_dict["Codename"]
+                    if "codename" in release_file_dict:
+                        release_file.codename = release_file_dict["Codename"]
                     if "suite" in release_file_dict:
-                        release_file.suite = release_file_dict.get("Suite")
+                        release_file.suite = release_file_dict["Suite"]
                     # TODO split of extra stuff e.g. : 'updates/main' -> 'main'
```
mdellweg, line 261: Is this todo still open?

pulp_deb/app/tasks/publishing.py
```diff
@@ -237,7 +237,10 @@ def __init__(
         if suite:
             self.release["Suite"] = suite
         self.release["Version"] = version
-        self.release["Codename"] = codename
+        if codename:
+            self.release["Codename"] = codename
+        else:
+            self.release["Codename"] = distirbution.split("/")[0]
```
quba42, line 243: I feel like our own publisher should set a "Codename" even if missing from the upstream (as a matter of good practice).
The natural fallback would be the first part of the distribution (so if the distribution is "stable/updates/extralong", it would be "stable").
My implementation does presume that people are not adding a slash at the front of the distribution (which they should not be doing, but we are not ensuring it...)

pulp_deb/app/tasks/synchronizing.py
quba42: My opinion is that we no longer need this TODO.
If the upstream has extra stuff in its components, we do still record it as is (i.e. with the extra stuff), but when we then retrieve it for use, we use the `palin_component()` property function where appropriate.

This is how we should keep it.

pulp_deb/app/tasks/publishing.py
quba42: Allready changed it from "fallback" to the base of the distribution (`distribution.split("/")[0]`), but I did not see your more concise  syntax in time, will amend.

quba42:
@mdellweg Do you have any idea why the tests are not running on this branch?

quba42:
Rebased to check if tests will run again (now using GitHub actions).

mdellweg:
check-manifest is missing. You might need to regenerate some `*_requirements.txt`

quba42:
@mdellweg See https://github.com/pulp/pulp_deb/pull/231

PR #232, opened 2020-12-01
Add latest plugin template changes

quba42:
[noissue]

Generated usign:

./plugin-template --github pulp_deb

Plugin template commit:

f64162216fe57f9f573d1bb59f74f2bd85a70e4a


fao89 (commented):
Looks good to me, please verify if the needed environment variables are declared on 
https://github.com/pulp/pulp_deb/settings/secrets/actions

quba42:
> Looks good to me, please verify if the needed environment variables are declared on
> https://github.com/pulp/pulp_deb/settings/secrets/actions

As far as I can tell the only secret variable `pulp_deb` has set is `RUBYGEMS_API_KEY`, which appears to be inherited from the organization somehow. From a quick `git grep secrets`, `pulp_deb` is missing at least the following: `PYPI_PASSWORD`, `PULP_DOCS_KEY`, `REDMINE_API_KEY`. Without the ability to compare with any other plugins, I can't be sure if that is everything.

PR #236, opened 2020-12-02
Fix unit test artifact mocking

quba42:
[noissue]


quba42:
Apparently pulpcore now checks if all checksum fields are there before saving artifacts.

quba42:
Should also be cherry-picked to the `2.8` release branch!

PR #237, opened 2020-12-02
Add latest plugin template changes

quba42:
This should hopefully fix docs publishing.

[noissue]

Generated usign:

./plugin-template --github pulp_deb

Plugin template commit:

f02e5898fdcccf502127277b4ff7c27d3620f8fa

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -132,9 +132,8 @@ jobs:
       
       - name: Publish nightly docs
         env:
-          BRANCH: ${GITHUB_REF##*/}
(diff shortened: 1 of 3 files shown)


quba42:
@daviddavis In your estimation, should I also cherry-pick this to the `2.8` release branch?

Also, are docs currently released in the release workflow only? (I think they used to be updated as soon as something was merged to master.)

daviddavis:
I would cherry pick this to the 2.8 branch. If you do a release from it, I believe it will publish the docs to wrong place again.

The docs are publishing at release time and nightly. There was some discussion about whether to publish the docs again when we push (see https://pulp.plan.io/issues/7910). Not sure if we'll do it again.

PR #238, opened 2020-12-07
Update usage of monitor_task() test utility function

dralley:


quba42:
Since I don't have a clear understanding of what this change is about, It would be great if someone else could also have a quick look at it as well.
The secret handling changes to the CI just look like the latest plugin template changes, (when I apply the template from the `2020.12.07` tag, the changes look very similar). I think we can count those as reviewed.

Like I said, I don't know exactly what the test changes are about, but they do not look like anything major and the tests are still green, so I guess that is a review of sorts... Should I just merge this?

dralley:
@quba42 Not yet.

The changes are basically that:

* We wanted to go from every plugin having its own version of monitor_task() to just one version provided by pulp-smash
* I didn't think it made much sense to return the created resources list when a user might want the rest of the info, and accessing created resources is really easy

The template update is because the "Required PR" part of the automation was broken for pulp-smash, so I fixed it and copied the change over.

@mdellweg could you add your review as well?

mdellweg (approved):
I am ok with the change. However since this plugin does not *yet* use the pulp-smash provided version of `monitor_task`, we can safely wait for it to be merged.

mdellweg:
Oh, i see, it was merged, so this can go ahead.

PR #240, opened 2020-12-15
Moves version to PulpPluginAppConfig

mdellweg:
This redeploys the plugin template and updates a few files that should
be updated to match but are not plugin template managed.  Specifically:

* Removing `__version__` from `pulp_deb/__init__.py`
* Adding the version string directly to `docs/conf.py`
* Adding the `version` attribute to `PulpPluginAppConfig` subclass in
`pulp_deb/app/__init__.py`.
* Updating .bumpversion.cfg

[noissue]


ipanova (approved):

quba42:
Just for reference, I understand this addresses: https://www.redhat.com/archives/pulp-dev/2020-December/msg00011.html

pyproject.toml
```diff
@@ -48,6 +48,7 @@ ignore = [
     ".bumpversion.cfg",
     "CHANGES/**",
     "CONTRIBUTING.rst",
+    "HISTORY.rst",
```
quba42, line 51:
```suggestion
```

docs/changes.rst
```diff
@@ -1 +1,5 @@
+.. _pulp_deb-changes:
+
 .. include:: ../CHANGES.rst
+
+.. include:: ../HISTORY.rst
```
quba42, line 5:
```suggestion
```
I have only one request, can we not include this empty `HISTORY.rst` file? (Note I can't leave a comment on the file itself, since it is empty).
I guess it comes from the plugin template, but what is it even intended for?
Are any plugins using it? (If not I could open a PR to remove it from the template as well...)

quba42 (changes requested):
Since I don't know the exact point in the plugin template that was applied, I am having a little trouble sorting out manual changes from the plugin template. However, comparing with https://github.com/pulp/pulp_file/pull/451/files, and given @ipanova's approval and the green tests I am happy to merge. Thanks for taking this off my hands!

My only requested change is that we not add the empty `HISTORY.rst` file from the plugin template.

mdellweg:
@quba42 removed HISTORY.rst.
I have no idea what its detailed purpose was.

quba42 (approved):

PR #243, opened 2021-01-14
Doc improvements

quba42:
This PR adds several small documentation improvements:

- A minor change to the README
- Updated plugin template usage instructions
- A updating API docs how to
- Also updates the `docs/_static/api.json` file

### DIFF
--- a/docs/_static/api.json
+++ b/docs/_static/api.json
@@ -15,6 +15,14 @@
         },
         "x-logo": {
             "url": "https://pulp.plan.io/attachments/download/517478/pulp_logo_word_rectangle.svg"
+        },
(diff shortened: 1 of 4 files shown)


docs/_static/api.json
```diff
@@ -15,6 +15,14 @@
         },
```
mdellweg, line 15: I thought, with publishing the docs on pulpproject.org, this file wasn't needed anymore.

quba42: If so, I will get rid of it and everything relating to it, in a new PR once the docs have actually made it to: https://docs.pulpproject.org/pulp_deb/

PR #245, opened 2021-01-19
Change all plugin documentation URLs

quba42:
fixes #8113
https://pulp.plan.io/issues/8113


mdellweg (approved):

.gitignore
```diff
@@ -66,6 +66,7 @@ instance/
 
 # Sphinx documentation
 docs/_build/
+docs/_static/
```
quba42, line 69: I may need a `.keep` file to ensure the folder is there. (Need to double check if the build will work without the folder existing).

quba42:
I will wait for final confirmation by @dkliban that `docs/_static/ap.json` is not needed in source.
However, from what I see in `.ci/scripts/docs-builder.py` it looks like this is generated on the fly by the build pipeline.

.gitignore
quba42: Never mind, it already contains something named `.this_makes_it_so_git_keeps_the__static_folder`.

docs/plugin_maintenance.rst
```diff
@@ -93,28 +93,25 @@ In order to apply a full plugin skeleton from the plugin template use:
...
    make html
 
-You now have any changes to ``docs/_static/api.json`` in your local ``pulp_deb`` repository.
-Commit, push, and merge these changes as needed.
+You can now find the built documentation at ``docs/_build/html/index.html`` within your local ``pulp_deb`` repository.
+You can also find the API doc contents within ``docs/_static/api.json``.
 
-You will also have a built version of this documentation at ``docs/_build/html/index.html``.
-You can open this locally built documentation in a browser, but you will not be able to view the API docs, since those make use of an external service, that obviously has no access to your local build.
+You can open the locally built documentation in a browser, but you will not be able to view the API docs, since those make use of an external service, that obviously has no access to your local build.
```
mdellweg, line 114: How so?

mdellweg (approved):

docs/plugin_maintenance.rst
quba42: My analysis of the reason why this fails could be wrong, but I get the following trying to view locally built API docs:
local_api_docs

mdellweg: right

PR #247, opened 2021-02-18
Update plugin template

mdellweg:
This accommodates for a change in pulpcore that will effect the CI.

https://github.com/pulp/pulpcore/pull/1102

[noissue]


quba42:
Closing since an even newer version of the plugin template was applied by https://github.com/pulp/pulp_deb/pull/251.

PR #251, opened 2021-03-09
Fix CI pipeline

quba42:
Since there are several independent things currently broken within the test pipeline, I am collecting them all on this branch.


quba42:
Now the tests are green, but only since the vast majority is being skipped. :disappointed:

pulp_deb/tests/functional/setup_signing_service.py
```diff
@@ -22,7 +36,12 @@
...
     from pulp_deb.app.models import AptReleaseSigningService
 
+    with open(public_key_file, "rb") as key:
+        public_key = key.read()
+
     AptReleaseSigningService.objects.create(
         name="sign_deb_release",
         script=script_path,
+        public_key=public_key,
+        pubkey_fingerprint="6EDF301256480B9B801EBA3D05A5E6DA269D9D98",
```
mdellweg, line 46: This is just to get the signing service setup again.
The validation step still needs to be adjusted in a followup work.

mdellweg (approved):

pulp_deb/tests/functional/setup_signing_service.py
quba42: yup.

PR #252, opened 2021-03-11
Update AptReleaseSigningService validation

quba42:
This is a direct consequence of https://github.com/pulp/pulpcore/pull/1012

fixes #8307
https://pulp.plan.io/issues/8307

### DIFF
--- a/pulp_deb/app/models/signing_service.py
+++ b/pulp_deb/app/models/signing_service.py
@@ -24,7 +24,6 @@ class AptReleaseSigningService(SigningService):
             "inline": "<relative_path>/InRelease",
             "detached": "<relative_path>/Release.gpg",
           }
-          "public_key" : "<relative_path>/public.key"
         }
 
         It will also ensure that the so returned files do indeed provide valid signatures as
@@ -41,12 +40,7 @@ class AptReleaseSigningService(SigningService):
                 test_file.flush()
                 return_value = self.sign(test_release_path)
 
-                public_key_path = return_value.get("public_key")
                 signatures = return_value.get("signatures")
-
-                if not public_key_path:
-                    message = "The signing service script must report a 'public_key' field!"
-                    raise RuntimeError(message)
 
                 if not signatures:
                     message = "The signing service script must report a 'signatures' field!"
@@ -76,8 +70,18 @@ class AptReleaseSigningService(SigningService):
 
                 # Prepare GPG:
                 gpg = gnupg.GPG(gnupghome=temp_directory_name)
-                with open(public_key_path, "rb") as key:
-                    gpg.import_keys(key.read())
+                gpg.import_keys(self.public_key)
+                imported_keys = gpg.list_keys()
+
+                if len(imported_keys) != 1:
+                    message = "We have imported more than one key! Aborting validation!"
+                    raise RuntimeError(message)
+
+                if imported_keys[0]["fingerprint"] != self.pubkey_fingerprint:
+                    message = (
+                        "The signing service fingerprint does not appear to match its public key!"
+                    )
+                    raise RuntimeError(message)
 
                 # Verify InRelease file
                 inline_path = signatures.get("inline")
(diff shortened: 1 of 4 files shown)


pulp_deb/app/models/signing_service.py
```diff
@@ -98,6 +91,12 @@ def validate(self):
                             message = "GPG Verification of the inline file '{}' failed!"
                             raise RuntimeError(message.format(inline_path))
 
+                        if verified.pubkey_fingerprint != self.pubkey_fingerprint:
+                            message = (
+                                "The fingerprint does not appear to match the public key provided!"
+                            )
```
mdellweg, line 97: :100:

pulp_deb/app/models/signing_service.py
```diff
@@ -80,8 +74,7 @@ def validate(self):
 
                 # Prepare GPG:
                 gpg = gnupg.GPG(gnupghome=temp_directory_name)
-                with open(public_key_path, "rb") as key:
-                    gpg.import_keys(key.read())
+                gpg.import_keys(self.public_key)
```
mdellweg, line 77: I think, you should either check the fingerprint of the imported key against the provided value, or maybe use the validate method to populate the fingerprint field.

quba42: I am not sure I understand what the intended workflow is. Should the person creating the `SigningService` provide both the `public_key` and the `pubkey_fingerprint`, or can `pubkey_fingerprint` simply be extracted from the `public_key` provided?
If it is the latter, shouldn't this be handled by pulpcore code, rather than having each class inheriting from `SigningService` do so?

mdellweg: That is exaclty the question I'm not entirely sure about. In any case, the validate should make sure they match.

cc @bmbouter

mdellweg: Maybe this is hard to access, since there is no interface for the "add a singing service" task (yet) other than blank django shell.

lubosmj: @quba42, yes, the intended workflow is to provide both the public key and its fingerprint. See https://docs.pulpproject.org/pulpcore/workflows/signed-metadata.html for more details.

quba42: Then I think the current version of the `validate` function which checks that the fingerprint of the key that was used to verify the signature on the test data is the same as the fingerprint contained in the signing service (after importing only the key contained in the signing service into a temporary gpghome folder), should be sufficient.

(GPG being the clunky unhelpful piece of software it is, I can't think of a way to do a more direct validation between the publik key and fingerprint contained in the signing service. Suggestions welcome...)

quba42: I have added an explicit check comparing the fingerprint of the imported public key to the fingerprint contained in the signing service.

pulp_deb/app/models/signing_service.py
quba42: I have changed the text to report more accurately what is actually checked, and also added a more direct check of the fingerprint.

pulp_deb/app/models/signing_service.py
```diff
@@ -94,6 +98,10 @@ def validate(self):
                             message = "GPG Verification of the inline file '{}' failed!"
                             raise RuntimeError(message.format(inline_path))
 
+                        if verified.pubkey_fingerprint != self.pubkey_fingerprint:
+                            message = "We appear to have validated '{}' using the wrong key!"
```
mdellweg, line 102: It's more like the signing has been performed with the wrong key. IMHO.

pulp_deb/tests/functional/sign_deb_release.sh
```diff
@@ -30,7 +24,6 @@ COMMON_GPG_OPTS="--batch --armor --digest-algo SHA256"
   "${RELEASE_FILE}"
 
 echo { \
-       \"public_key\": \"${PUBLIC_KEY_PATH}\", \
```
mdellweg, line 33: :+1:

pulp_deb/app/models/signing_service.py
quba42: You are right, I changed the formulation.
Technically it should not really be possible to run into this condition since we have already checked that the signing service key is the only public key we have available. However, in this case verified is very much better than assumed!

mdellweg (approved):

PR #253, opened 2021-03-15
Handle ALLOWED_CONTENT_CHECKSUMS parameter

quba42:
fixes #8388
https://pulp.plan.io/issues/8388

TODO:
- [ ] test both with md5, sha1 enabled as well as disabled
- [x] Add WARNINGS when md5 is disabled plus ability to disable the warning
- [x] documentation


quba42:
As far as I am concerned the current state is functionally complete.
To what extent the tests can and should be adjusted could be done as a follow on task.

mdellweg (approved):

PR #254, opened 2021-03-16
Cleanup logging

quba42:
These cosmetic issues were bothering me while working on other things. :wink:

[noissue]

* Remove unused loggers
* Replace deprecated warn() calls with warning()
* Use gettext function for all log messages
* Name all loggers "log"

### DIFF
--- a/pulp_deb/app/tasks/synchronizing.py
+++ b/pulp_deb/app/tasks/synchronizing.py
@@ -1,5 +1,3 @@
-from gettext import gettext as _
-import logging
 import asyncio
 import aiohttp
 import os
@@ -8,6 +6,7 @@ import bz2
 import gzip
 import lzma
 import gnupg
+
 from collections import defaultdict
 from tempfile import NamedTemporaryFile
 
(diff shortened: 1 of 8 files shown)


pulp_deb/app/tasks/synchronizing.py
```diff
@@ -303,7 +305,7 @@ async def run(self):
                         d_content.d_artifacts.append(da)
                         da.artifact.save()
                         log.info(
-                            "*** Expected: {} *** Uncompressed: {} ***".format(
+                            _("*** Expected: {} *** Uncompressed: {} ***").format(
```
mdellweg, line 308: Hmm. This should never have been in the code it guess. Or changed into a warning/error when they mismatch.

mdellweg (approved):

pulp_deb/app/tasks/synchronizing.py
quba42: Dropped it as a debug log message that was forgotten.

PR #255, opened 2021-03-17
Update CI files

mdellweg:
[noissue]

### DIFF
--- a/.ci/scripts/release_requirements.txt
+++ b/.ci/scripts/release_requirements.txt
@@ -1,4 +1,0 @@
-bump2version
-gitpython
-python-redmine
-towncrier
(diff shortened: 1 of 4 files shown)


.ci/scripts/release_requirements.txt
```diff
@@ -1,4 +0,0 @@
-bump2version
-gitpython
-python-redmine
-towncrier
```
quba42, line 4: This file is still referenced from the release documentation, so that would need to be updated when we drop it.

mdellweg: Done.
Actually it has been moved to `.ci/assets`

quba42 (approved):
LGTM!

PR #260, opened 2021-04-01
Replace workingdir with Python stdlib

dralley:
[noissue]

### DIFF
--- a/pulp_deb/app/tasks/publishing.py
+++ b/pulp_deb/app/tasks/publishing.py
@@ -4,6 +4,7 @@ import shutil
 from datetime import datetime, timezone
 from debian import deb822
 from gzip import GzipFile
+import tempfile
 
 from django.conf import settings
 from django.core.files import File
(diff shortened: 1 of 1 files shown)


mdellweg (approved):

mdellweg:
It is hanging on the S3 tests again. And i cannot see that this PR is causing it.

quba42:
> It is hanging on the S3 tests again. And i cannot see that this PR is causing it.

The nightly s3 job is successful and the error message is not very informative: 
```
pulp_deb/tests/functional/api/test_crud_content_unit.py::GenericContentUnitUploadTestCase::test_03_duplicate_content_unit PASSED [ 23%]
Error: The operation was canceled.
```

I have no ideas.

mdellweg:
It is hitting the timeout (between 5 h and 6 h) of gh workers.
I just don't understand why it is hanging. What i want to say is, that it is probably ok to merge.

quba42 (approved):

PR #287, opened 2021-05-26
Consider that release file components may not be plain components

quba42:
Closes #8828
https://pulp.plan.io/issues/8828

When creating an intersection between the release file components and
the components from the remote, we need to consider the special case,
where the release file components have a path prefix, but those from the 
remote do not!

### DIFF
--- a/CHANGES/8828.bugfix
+++ b/CHANGES/8828.bugfix
@@ -1,0 +1,1 @@
+Fixed a bug where components from the remote were being ignored when specified as the plain component.
--- a/pulp_deb/app/tasks/synchronizing.py
+++ b/pulp_deb/app/tasks/synchronizing.py
@@ -177,12 +177,23 @@ class DebDeclarativeVersion(DeclarativeVersion):
         return pipeline
 
 
-def _filter_split(values, filter_list):
-    """Filter space separated list and return iterable."""
-    value_set = set(values.split())
-    if filter_list:
-        value_set &= set(filter_list)
-    return sorted(value_set)
+def _filter_split(values, filter_values):
+    """
+    Returns the intersection of two strings of whitespace separated elements as a sorted set.
+    If an element of values.split() has a path prefix, it is still considered to be equal to an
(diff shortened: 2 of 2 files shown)


quba42:
Looks like I need to amend the test as well.

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -378,20 +389,6 @@ def __init__(self, remote, *args, **kwargs):
...
         self.remote = remote
         self.parsed_url = urlparse(remote.url)
-        self.distributions = [
-            distribution.strip() for distribution in self.remote.distributions.split()
-        ]
-        self.num_distributions = len(self.distributions)
-        self.components = (
-            None
-            if self.remote.components is None
-            else [component.strip() for component in self.remote.components.split()]
```
quba42, line 388: I am fairly confident that `strip()` on the elements of a `split()` is redundant, since split (without argument) already splits on "any and all whitespace", stripping it all away in the process.

@mdellweg Or am I missing some strange corner case where the additional `strip()` is important?

mdellweg: How do you think I could still remember?
I would try with multiple spaces between the entries.

CHANGES/8828.bugfix
```diff
@@ -0,0 +1 @@
+Fixed a bug were components from the remote were being ignored when specified as the plain component.
```
quba42, line 1:
```suggestion
Fixed a bug where components from the remote were being ignored when specified as the plain component.
```

pulp_deb/app/tasks/synchronizing.py
quba42: I guess I did not really, I just thought asking can't hurt.
That being said I have tested the behavior with strings with multiple spaces, tabs, newlines etc. and it all seems to work fine.
It could be that the strip was copied from pulp 2 where we had comma separated instead of space separated so the additional strip makes sense there...

Manisha15 (approved):

PR #288, opened 2021-05-27
Do not explicitly cast primary keys to str

quba42:
[noissue]

The need to do this was undeprecated as it is now performed in pulpcore.


quba42:
Does it make sense to actively undo this?

mdellweg (approved):

quba42:
Looks like the pulp test got stuck. Given I ran them locally, and the s3 tests also run the same tests, and I have seen this happen from time to time recently, I am merging anyway.

PR #294, opened 2021-06-09
Add custom fields to packages

m-bucher:


m-bucher:
I feel I should add some tests, but I am unsure where.
Basically this could be a unittest for the Serializer, but that one is disabled currently :thinking:

quba42:
This feature should only be merged after we have a fix for the missing `Size` field. This mandatory field should not fall into the catch all custom fields backup mechanism from this PR.

m-bucher:
> This feature should only be merged after we have a fix for the missing `Size` field. This mandatory field should not fall into the catch all custom fields backup mechanism from this PR.

See #300

quba42:
Size fix is now merged.

CHANGES/8232.bugfix
```diff
@@ -0,0 +1 @@
+Add custom_fields to hold non-standard PackageIndex fields
```
quba42, line 1: I can't comment on the filename `CHANGES/8232.bugfix`, but I feel like with the Size field fix handled separately, this is more of a feature than a bugfix. It adds the ability to retain rare optional fields, which should not actually be required for any core APT functionality, that is, "not a bug".

quba42 (approved):
Seems to work well. A get on a package unit does not show the new field, but that is probably for the best. We can always add that functionality if the need ever arises.

PR #295, opened 2021-06-15
Initial source file and source indices support

masselstine:
fixes #8775
https://pulp.plan.io/issues/8775

Allows the upload of source files (orig.tar, debian.tar) and Debian
Source Control files (dsc files), as well as synchronizing with a
remote's source indices (sync_sources=True).

The source_files referenced in a dsc_file must be uploaded before the
dsc_file so a typical workflow might be

--
apt-get source bc

http --form post $BASE/pulp/api/v3/content/deb/source_files/ \
  file@"bc_1.07.1.orig.tar.gz" name=bc_1.07.1.orig.tar.gz
http --form post $BASE/pulp/api/v3/content/deb/source_files/ \
  file@"bc_1.07.1-2build1.debian.tar.xz" name=bc_1.07.1-2build1.debian.tar.xz
http --form post $BASE/pulp/api/v3/content/deb/dsc_files/ \
  file@"bc/bc_1.07.1-2build1.dsc"
--

Attempting to upload the dsc file before its source files will result
validation errors.

Once uploaded source_files and dsc_files can be inspected via their
pulp_href or using the respective endpoints

http get $BASE/pulp/api/v3/content/deb/dsc_files/
http get $BASE/pulp/api/v3/content/deb/source_files/

Synchronizing with a remote with "sync_sources=True" will synchronize
the Sources files as well as download all associated source_files and
dsc_files, performing a high level of data validation and creating all
required associations. As with uploading inspecting contents after a
sync is complete can be done with the same endpoints as above.

To remain spec compliant with https://wiki.debian.org/DebianRepository
"md5" must be present in the ALLOWED_CONTENT_CHECKSUMS. In all cases
the use of md5 is supplemental so security concerns around this
addition, at least for use with this feature, should be minimal.

TODO: publishing - this work has been started and will be completed
over the coming days.


quba42:
I had to approve running the tests...

It looks like the linter wants you to add a file named:
```
CHANGES/8775.feature
```
containing a short changelog entry, perhaps something like:
```
Added support for the synchronization and upload of source packages.
```
(Feel free to amend the wording since you most likely know more about Debian source packages than I do. :wink:)

mdellweg:
I always figured, the dsc and the tarball to be part of a single two artifact content unit. Have you considered that?

masselstine:
> I always figured, the dsc and the tarball to be part of a single two artifact content unit. Have you considered that?

AFAIK the orig.tar can be used by multiple revisions of a DSC. For example upstream releases foo version 1.2.3 and a package maintainer produces a Dsc foo_1.2.3-1.dsc, the package maintainer then refines their package and produces foo_1.2.3-2.dsc but the upstream project content doesn't change, so the orig.tar remains the same. As such we definitely can't include the orig.tar in the MultiArtifact. NOTE: (native)  packages do not require an orig.tar and since 3.0 (quilt) there can be an arbitrary number of orig.tar.

This leaves just the debian.tar (format v3) or diff.tar (format v1). I agree that this does appear to be ripe for handling as a MultiArtifact. I believe that like the orig.tar that this file is optional but I can't actually dig up this information, would this cause any issue with using MultiArtifact? I admittedly had no experience with Django going into this and it could be this inexperience but I fail to see an way to implement this. The artifact handling should be easy enough, my difficulties are more about the additional meta-data handling that a simple artifact simply doesn't have.  I have to go through my code today and cleanup the linter warning and such but I am 100% open to some assistance with taking what I have and making adjustments.

evgeni:
> > I always figured, the dsc and the tarball to be part of a single two artifact content unit. Have you considered that?
> 
> AFAIK the orig.tar can be used by multiple revisions of a DSC. For example upstream releases foo version 1.2.3 and a package maintainer produces a Dsc foo_1.2.3-1.dsc, the package maintainer then refines their package and produces foo_1.2.3-2.dsc but the upstream project content doesn't change, so the orig.tar remains the same. As such we definitely can't include the orig.tar in the MultiArtifact. NOTE: (native)  packages do not require an orig.tar and since 3.0 (quilt) there can be an arbitrary number of orig.tar.

Isn't it even harder? I remember that foo-1.2.3-1 and foo-1.2.3-2 can't have different orig tar gzs? Or was that a different tool, not the archive enforcing that?

masselstine:
> > > I always figured, the dsc and the tarball to be part of a single two artifact content unit. Have you considered that?
> > 
> > 
> > AFAIK the orig.tar can be used by multiple revisions of a DSC. For example upstream releases foo version 1.2.3 and a package maintainer produces a Dsc foo_1.2.3-1.dsc, the package maintainer then refines their package and produces foo_1.2.3-2.dsc but the upstream project content doesn't change, so the orig.tar remains the same. As such we definitely can't include the orig.tar in the MultiArtifact. NOTE: (native)  packages do not require an orig.tar and since 3.0 (quilt) there can be an arbitrary number of orig.tar.
> 
> Isn't it even harder? I remember that foo-1.2.3-1 and foo-1.2.3-2 can't have different orig tar gzs? Or was that a different tool, not the archive enforcing that?

It makes sense but I have failed in my search of the documentation and may have to revert to reading code. For as good as the Debian docs can be, they talk around the topic of required/optional/enforced when it comes to the files included in the Debian Source Control file. At any rate, I am near 100% sure the only file that could be included in a MultiArtifact approach would be the debian.tar/diff.tar. Any orig.tar need to exist on their own.

masselstine:
> > > > I always figured, the dsc and the tarball to be part of a single two artifact content unit. Have you considered that?
> > > 
> > > 
> > > AFAIK the orig.tar can be used by multiple revisions of a DSC. For example upstream releases foo version 1.2.3 and a package maintainer produces a Dsc foo_1.2.3-1.dsc, the package maintainer then refines their package and produces foo_1.2.3-2.dsc but the upstream project content doesn't change, so the orig.tar remains the same. As such we definitely can't include the orig.tar in the MultiArtifact. NOTE: (native)  packages do not require an orig.tar and since 3.0 (quilt) there can be an arbitrary number of orig.tar.
> > 
> > 
> > Isn't it even harder? I remember that foo-1.2.3-1 and foo-1.2.3-2 can't have different orig tar gzs? Or was that a different tool, not the archive enforcing that?
> 
> It makes sense but I have failed in my search of the documentation and may have to revert to reading code. For as good as the Debian docs can be, they talk around the topic of required/optional/enforced when it comes to the files included in the Debian Source Control file. At any rate, I am near 100% sure the only file that could be included in a MultiArtifact approach would be the debian.tar/diff.tar. Any orig.tar need to exist on their own.

BTW I found an example of a package which does not have a debian.tar, 'apt'
http://ftp.ca.debian.org/debian/pool/main/a/apt/

So based on documentation and examples there can be 0, 1 or more orig.tar and these can be shared by multiple revisions of .dsc and .debian.tar/diff.tar and the debian.tar/diff.tar are optional.

I am about to push another update to the WIP with the final TODO items taken care of, code formatted and sanitized. Anyone can feel free to apply the work to their repo and give it a whirl and provide feedback. I will review the MultiArtifact over the weekend given the above constraints to hopefully be able to better answer this question.

mdellweg:
Not sure anymore, how useful it is, but you can first upload all the files as artifacts, and then upload a DSC file (which contains checksums for the others) and then the content unit is built containing all the artifacts.
Artifacts (aka tar-files) will be shared by multiple content units just fine.
The association which source package belongs to a repo version would imho be easier.

masselstine:
> Not sure anymore, how useful it is, but you can first upload all the files as artifacts, and then upload a DSC file (which contains checksums for the others) and then the content unit is built containing all the artifacts.
> Artifacts (aka tar-files) will be shared by multiple content units just fine.
> The association which source package belongs to a repo version would imho be easier.

I am looking into this over the weekend and have it as a TODO on the latest commit log, so I am definitely exploring this. I can see how the  associations would be easier so I agree it is worth the exploration.

masselstine:
> > Not sure anymore, how useful it is, but you can first upload all the files as artifacts, and then upload a DSC file (which contains checksums for the others) and then the content unit is built containing all the artifacts.
> > Artifacts (aka tar-files) will be shared by multiple content units just fine.
> > The association which source package belongs to a repo version would imho be easier.
> 
> I am looking into this over the weekend and have it as a TODO on the latest commit log, so I am definitely exploring this. I can see how the associations would be easier so I agree it is worth the exploration.

I didn't get the time over the weekend to look into this as much as I wanted, but did have some time today. In the process I discovered a few flaws in my approach and some ways to get us closer to using MulitpleArtifact if we choose to go that route. I have dropped the 'source_files' endpoint and now upload non-DSC source files as standard artifacts using the /pulp/api/v3/artifacts/ endpoint. I still have DscFile model and endpoint which gets associated with the artifacts based on the sha256 and size.

I am still not convinced MultipleArtifact is the way to go, but again this takes us closer to it so it can be explored. I will push an update tomorrow.

quba42:
The linting can be somewhat nit-picky:
```
Run flake8 --config flake8.cfg
./pulp_deb/app/models/content.py:481:1: D107 Missing docstring in __init__
./pulp_deb/app/serializers/content_serializers.py:950:1: D102 Missing docstring in public method
```

masselstine:
> The linting can be somewhat nit-picky:
> 
> ```
> Run flake8 --config flake8.cfg
> ./pulp_deb/app/models/content.py:481:1: D107 Missing docstring in __init__
> ./pulp_deb/app/serializers/content_serializers.py:950:1: D102 Missing docstring in public method
> ```

Yes it is. I have flake8 setup locally and it didn't catch this. I have corrected it but I have also changed a lot of code. I will push an update up tonight and see what comes up once you release the linter on it :)

masselstine:
> > > Not sure anymore, how useful it is, but you can first upload all the files as artifacts, and then upload a DSC file (which contains checksums for the others) and then the content unit is built containing all the artifacts.
> > > Artifacts (aka tar-files) will be shared by multiple content units just fine.
> > > The association which source package belongs to a repo version would imho be easier.
> > 
> > 
> > I am looking into this over the weekend and have it as a TODO on the latest commit log, so I am definitely exploring this. I can see how the associations would be easier so I agree it is worth the exploration.
> 
> I didn't get the time over the weekend to look into this as much as I wanted, but did have some time today. In the process I discovered a few flaws in my approach and some ways to get us closer to using MulitpleArtifact if we choose to go that route. I have dropped the 'source_files' endpoint and now upload non-DSC source files as standard artifacts using the /pulp/api/v3/artifacts/ endpoint. I still have DscFile model and endpoint which gets associated with the artifacts based on the sha256 and size.
> 
> I am still not convinced MultipleArtifact is the way to go, but again this takes us closer to it so it can be explored. I will push an update tomorrow.

OK. I have dug deeper into this than I had planned. The MultipleArtifact, as far as I can tell, is for handling one or more Artifact+Content where the Content is homogeneous. We could treat all source files (.dsc, *tar) the same, ie. with one Content type, but we would still require a second Content type to host the .dsc file/paragraph meta-data. During synchronization or when associating multiple artifacts to the Content we would have to review all the Artifacts to figure out which one is the .dsc. On top of all this I see no way to reassemble to things when it comes time to publish a DSC paragraph. I will be the first to admit that I may be missing the forest for the trees, but I see no way to treat this in a single Content type. Just review the MultipleArtifactContentSerializer create() method

`
for relative_path, artifact in artifacts.items():
            models.ContentArtifact.objects.create(
                artifact=artifact, content=content, relative_path=relative_path
            )
`

What I have done is made it so you can upload non-DSC files as an artifact and when you do upload the DSC file/or associate it via the dsc_file endpoint, the serialization code validates that the required artifacts are present and during create() it automatically associates the artifacts to the SourceFile content type (via ContentArtifact). This ensure the Artifact can't be deleted and I am working through the release and publishing workflows to ensure they can be handled correctly. Once I am happy with the changes I will post the code.

quba42:
> OK. I have dug deeper into this than I had planned. The MultipleArtifact, as far as I can tell, is for handling one or more Artifact+Content where the Content is homogeneous.

I am not sure if this is sufficient for the DSC file case, but we do have something similar for the `ReleaseFile` content unit, which is associated with multiple artifacts (any number of `Release`, `InRelease`, and `Release.gpg`), but one of them is retrievable as the "main artifact": https://github.com/pulp/pulp_deb/blob/784cf71d1d2aeeadb53c9f17a995c1b10d0722cc/pulp_deb/app/models/content.py#L62

It is then up to the sync/upload code that creates or updates this content unit to make sure that the correct artifact from the the list of artifacts can be retrieved as the "main artifact". Without having looked at it in too much detail, I could imagine something similar to always keep the DSC artifact easily accessible from the content unit. If that does not work, then it is probably better to create separate content units and associate them using foreign key relations.

mdellweg:
FWIW you will subclass the serializer anyway, and you can use the validate step to lookup and add all the other source file artifacts there and then ’super(). validate’ the rest. I have a strong suspicion this can work.
Btw I wouldn't call InRelease and Release homogenous.

masselstine:
As with many things I got busy with other work. I just wanted to drop a quick not to let folks know that this has not be forgotten. I should have time later this week to get back to this. Sorry for the delay.

masselstine:
> FWIW you will subclass the serializer anyway, and you can use the validate step to lookup and add all the other source file artifacts there and then ’super(). validate’ the rest. I have a strong suspicion this can work.
> Btw I wouldn't call InRelease and Release homogenous.

AFAICT I can overwrite the create() and deferred_validate() but I can't overwrite the field level validation in ContentArtifactsField.run_validation(). So an artifacts dict needs to be passed and can't be delayed until the dsc file contents are serialised. This provides a horrible UX as no only are all artifact URLs required but relative paths given instead of being derived. I can possibly kill the per-field validation but yet again that is defeating the point of using MultipleArtifact in the first place. I would really like to have someone take a stab at this as this might be chalked up to my lack of experience with Django.

mdellweg:
OK, i think, i know now where the confusion kicks in.
What you might want to do is something like this: https://github.com/pulp/pulp_gem/blob/master/pulp_gem/app/serializers.py#L48
The difference there being that the "spec_artifact" is produced in the serialitzation process, while for the dsc case it can rather be an artifact lookup.

And just to be sure, the serializers we are talking of here are really DRF (https://www.django-rest-framework.org/api-guide/serializers/), not plain django.

masselstine:
> OK, i think, i know now where the confusion kicks in.
> What you might want to do is something like this: https://github.com/pulp/pulp_gem/blob/master/pulp_gem/app/serializers.py#L48
> The difference there being that the "spec_artifact" is produced in the serialitzation process, while for the dsc case it can rather be an artifact lookup.
> 
> And just to be sure, the serializers we are talking of here are really DRF (https://www.django-rest-framework.org/api-guide/serializers/), not plain django.

Thanks for pointing me at this. It actually comes very close to what I have attempted but there are a few differences that I will review to see if they resolve the complications I have hit. I had tried to find other examples in some of the other plugins but came up blank.

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -652,6 +659,85 @@ async def _handle_package_index(
...
+            # No reference here, skip this component
+            return
+        log.info(_("Downloading: {}/Sources").format(source_index_dir))
+        content_unit = SourceIndex(
+            release=release_file,
+            component=release_component.component,
+            sha256=d_artifacts[0].artifact.sha256,
+            relative_path=os.path.join(release_base_path, source_index_dir, "Sources"),
+        )
+        source_index = await self._create_unit(
```
quba42, line 686: I found a small bug during unit creation: In line 361 of `synchronizing.py` you should add the following change:
```diff
-                if isinstance(d_content.content, PackageIndex):
+                if isinstance(d_content.content, PackageIndex) or isinstance(
+                    d_content.content, SourceIndex
+                ):
```
This will ensure every SourceIndex has a main artifact in Pulp, even if the upstream repo has only archived versions of the SourceIndex. I had to add this locally in order to synchronize successfully from official Debian repos.

quba42 (changes requested):

quba42:
I am also currently getting an error when trying to create APT publications (verbatim publications work fine):
```
pulp [13863e9a237b4c1ab567f7fb3ef2eba4]: pulpcore.tasking.pulpcore_worker:INFO:
  File "/home/vagrant/devel/pulpcore/pulpcore/tasking/pulpcore_worker.py", line 272, in _perform_task
    result = func(*args, **kwargs)
  File "/home/vagrant/devel/pulp_deb/pulp_deb/app/tasks/publishing.py", line 195, in publish
    release_helper.finish()
  File "/home/vagrant/devel/pulpcore/pulpcore/app/models/publication.py", line 177, in __exit__
    self.delete()
  File "/home/vagrant/devel/pulpcore/pulpcore/app/models/publication.py", line 135, in delete
    publication = pubs.latest("repository_version", "pulp_created")
  File "/usr/local/lib/pulp/lib64/python3.9/site-packages/django/db/models/query.py", line 649, in latest
    return self.reverse()._earliest(*fields, field_name=field_name)
  File "/usr/local/lib/pulp/lib64/python3.9/site-packages/django/db/models/query.py", line 643, in _earliest
    return obj.get()
  File "/usr/local/lib/pulp/lib64/python3.9/site-packages/django/db/models/query.py", line 406, in get
    raise self.model.DoesNotExist(
```
I am finding it difficult to follow the flow of execution here, it looks a bit like pulpcore is swallowing the error message for the actual root cause. I will look into this further next week.

In case anyone else wants to try and reproduce this, I have been testing with the following remote:
```
REMOTE_HREF=$(http post "${BASE_ADDR}"/pulp/api/v3/remotes/deb/apt/ name="${ENTITIES_NAME}" distributions="buster" components="contrib" architectures="amd64" url="http://ftp.de.debian.org/debian/" policy="on_demand" sync_sources=true | jq -r '.pulp_href')
```

quba42:
Some things we will need to do (possibly as follow on tasks once this is merged):

- [ ] Update the plugin docs
- [ ] Add some basic test coverate (will require new fixtures)
- [ ] Is there a use case for synchronizing only source, but not binary packages from a remote repo? (Currently not possible)

I am just leaving these here for now, so I don't forget about it later. :wink:

pulp_deb/app/tasks/synchronizing.py
mdellweg: `isinstance` can check multiple types in one call. It should be `isinstance(d_content.content, (PackageIndex, SourceIndex))`

https://docs.python.org/3/library/functions.html#isinstance

pulp_deb/app/tasks/publishing.py
```diff
@@ -169,6 +180,18 @@ def publish(repository_version_pk, simple=False, structured=False, signing_servi
...
+                        release_component__in=components,
+                    ):
+                        try:
+                            release_helper.components[drc.release_component.component].add_dsc_file(
+                                drc.dsc_file
+                            )
+                        except IntegrityError:
+                            continue
+
                     release_helper.finish()
```
quba42, line 195: I am getting the following error from this line when trying to create APT publications (verbatim publications work fine):
```
pulp [13863e9a237b4c1ab567f7fb3ef2eba4]: pulpcore.tasking.pulpcore_worker:INFO:
  File "/home/vagrant/devel/pulpcore/pulpcore/tasking/pulpcore_worker.py", line 272, in _perform_task
    result = func(*args, **kwargs)
  File "/home/vagrant/devel/pulp_deb/pulp_deb/app/tasks/publishing.py", line 195, in publish
    release_helper.finish()
  File "/home/vagrant/devel/pulpcore/pulpcore/app/models/publication.py", line 177, in __exit__
    self.delete()
  File "/home/vagrant/devel/pulpcore/pulpcore/app/models/publication.py", line 135, in delete
    publication = pubs.latest("repository_version", "pulp_created")
  File "/usr/local/lib/pulp/lib64/python3.9/site-packages/django/db/models/query.py", line 649, in latest
    return self.reverse()._earliest(*fields, field_name=field_name)
  File "/usr/local/lib/pulp/lib64/python3.9/site-packages/django/db/models/query.py", line 643, in _earliest
    return obj.get()
  File "/usr/local/lib/pulp/lib64/python3.9/site-packages/django/db/models/query.py", line 406, in get
    raise self.model.DoesNotExist(
```
I have confirmed the error only occurs for remotes with `sync_sources=true`. I have also rebased the entire change to the latest pulp_deb main branch without any change in error. Bizzarly, I am getting the same error with the same line, even when I publish using `structured=false` (I would expect this line is never executed for that code path).

I don't yet have any good ideas what might be causing this. I think the error is saying that some pulpcore query of the DB unexpectedly returned empty (not sure though).

The following remote should let anyone rwho wants to reproduce the problem:
```
REMOTE_HREF=$(http post "${BASE_ADDR}"/pulp/api/v3/remotes/deb/apt/ name="${ENTITIES_NAME}" distributions="buster" components="contrib" architectures="amd64" url="http://ftp.de.debian.org/debian/" policy="on_demand" sync_sources=true | jq -r '.pulp_href')
```

quba42 (changes requested):
The below issue needs to be fixed!

pulp_deb/app/tasks/publishing.py
masselstine: Thanks for giving the changes a spin. I am back at it this week and hope to get the updates done shortly.

pulp_deb/app/tasks/synchronizing.py
masselstine: Should I 'generisize' the DebUpdatePackageIndexAttributes() function name, progress report message... while I am at it?

quba42: As the TODO notes, the name was never that good to begin with. Something like `DebEnsureMainIndexArtifact()` would make more sense to me. Feel free to change it to that (or something similar if you have a better idea).

pulp_deb/app/tasks/publishing.py
masselstine: Continuing to look at this, sorry for the slow progress but summer holidays mean time with the kids, away from the computer. Immediate download works fine, on_demand is failing so I am trying to figure out why the declarative artifacts are not working, like they are for packages.

quba42: I have also been busy with other things. I think there is a good chance I can spend some serious time on it this week (or next week at the latest). I want to try out some changes on top of your branch as part of a proper review of this large feature. Just to gain some confidence I really understand what is happening. :wink:

masselstine:
@quba42 I didn't touch DebUpdatePackageIndexAttributes() rename/cleanup but I have made the necessary changes to support the remote on_demand policy. You can now retest your usecase of syncing with

`REMOTE_HREF=$(http post "${BASE_ADDR}"/pulp/api/v3/remotes/deb/apt/ name="${ENTITIES_NAME}" distributions="buster" components="contrib" architectures="amd64" url="http://ftp.de.debian.org/debian/" policy="on_demand" sync_sources=true | jq -r '.pulp_href')`

I tried to keep things as clear as possible with checksums as there can be several influences including ALLOWED_CONTENT_CHECKSUMS as well as the checksum types offered by the remote (since we have to rely on the contents of the source index on the remote as we don't have a copy of the artifacts since the download might be deferred.

Sorry again for the delay but I hope we are getting close to a satisfactory starting point with source support. Please let me know if you see any more issues with your use of the changes.

quba42:
I performed another round of testing, and can confirm that the basic functionality seems to work!
As far as I am concerned this PR is now "Ready for reveiw"
The failed test looks like a random fluke to me.

There are a few more TODOs for myself before merging:

- [ ] Experiment some more on top of the branch with respect to multi artifact types.
- [ ] Normal code review
- [ ] Perhaps start preparing some documentation to go with the new feature in a separate PR

I did notice one small thing (but it is not critical for merging this initial implementation): It looks like Source indices are currently loosing the `Priority` and `Section` fields relative to the upstream repo.

pulp_deb/app/tasks/publishing.py
quba42: on_demand policy now works.

masselstine:
> I performed another round of testing, and can confirm that the basic functionality seems to work! As far as I am concerned this PR is now "Ready for reveiw" The failed test looks like a random fluke to me.
> 
> There are a few more TODOs for myself before merging:
> 
> * [ ]  Experiment some more on top of the branch with respect to multi artifact types.
> * [ ]  Normal code review
> * [ ]  Perhaps start preparing some documentation to go with the new feature in a separate PR
> 
> I did notice one small thing (but it is not critical for merging this initial implementation): It looks like Source indices are currently loosing the `Priority` and `Section` fields relative to the upstream repo.

Hmmm. 'Priority' and 'Section' are optional (https://wiki.debian.org/DebianRepository/Format#A.22Sources.22_Indices) I was using Aptly as my source and I guess it never included these optional elements, so I simply didn't code them. These can be added easily enough as there should be no special handling.

quba42:
@masselstine I have been told, that if you re-base this PR to latest main branch (should not cause any conflicts), the s3 test can also turn green. (Apparently some version of something in the test setup had to be pinned).

I would not worry about the `Priority` and `Section` thing for now, like you said they are not required and can always be added later.

CHANGES/8775.feature
```diff
@@ -0,0 +1 @@
+Add support for uploading, synchronizing and publishing source files and the 'Sources' indices
```
quba42, line 1:
```suggestion
Add support for handling source packages.
```

I am not 100% sure if I am happy with this version, but I think it would be good to keep things clear and concise for the change-log.

pulp_deb/app/models/__init__.py
```diff
@@ -12,6 +12,10 @@
     ReleaseArchitecture,
     ReleaseComponent,
     ReleaseFile,
+    SourceIndex,
+    SourceContent,
```
quba42, line 16: As far as I can tell the `SourceContent` type is from an earlier version of this PR and no longer required.
I tested with the following changes, which seemed to work fine: https://github.com/quba42/pulp_deb/commit/b7d6168d2bb3b977a0b0746377fbe306028c2120 (of course I had to re-generate the migrations).

pulp_deb/app/models/content.py
```diff
@@ -327,3 +357,211 @@ class PackageReleaseComponent(Content):
...
+    sha1 = models.CharField(max_length=40, null=True)
+    sha256 = models.CharField(max_length=64, null=False)
+    sha512 = models.CharField(max_length=128, null=True)
+
+    class Meta:
+        default_related_name = "%(app_label)s_%(model_name)s"
+        unique_together = (("size", "sha256"),)
+
+
+class DscFile(Content):
```
quba42, line 383: Since this is now a multi artifact type, that does not just contain .dsc files, I would consider renaming `DscFile` to `SourcePackage`. Of course this will require quite some renaming.

pulp_deb/app/models/content.py
```diff
@@ -327,3 +357,211 @@ class PackageReleaseComponent(Content):
...
+                )
+        return contents
+
+    repo_key_fields = ("source", "version")
+
+    class Meta:
+        default_related_name = "%(app_label)s_%(model_name)s"
+
+
+class DscFileReleaseComponent(Content):
```
quba42, line 552:
```suggestion
class SourcePackageReleaseComponent(Content):
```
I would also rename this here.

quba42 (changes requested):
I found two more things I would change.
I have not yet gone through everything, so I may find more, but I am posting these now.

pulp_deb/app/models/__init__.py
masselstine: SourceContent is still used when uploading content
---
# Add artifacts
pulp_href=$($HTTP_CMD --form post $BASE/pulp/api/v3/artifacts/ file@"/tmp/bc_1.07.1-2build1.debian.tar.xz" | jq -r .pulp_href)
echo "Added /tmp/bc_1.07.1-2build1.debian.tar.xz as $pulp_href"
pulp_href=$($HTTP_CMD --form post $BASE/pulp/api/v3/artifacts/ file@"/tmp/bc_1.07.1.orig.tar.gz" | jq -r .pulp_href)
echo "Added /tmp/bc_1.07.1.orig.tar.gz as $pulp_href"
pulp_href=$($HTTP_CMD --form post $BASE/pulp/api/v3/artifacts/ file@"/tmp/bc_1.07.1-2build1.dsc" | jq -r .pulp_href)
echo "Added /tmp/bc_1.07.1-2build1.dsc as $pulp_href"
dsc_artifact_href=$pulp_href

# Create DSC file
task=$($HTTP_CMD --form post $BASE/pulp/api/v3/content/deb/dsc_files/ artifact=$dsc_artifact_href | jq -r .task)
---
I will have a  closer look to see if there is another approach we can use, but we can't simply drop this.

pulp_deb/app/models/content.py
masselstine: This is what a good editor is for. I will look at renaming it.

pulp_deb/app/models/content.py
masselstine: Same here, I will look at renaming it.

masselstine:
@quba42 I have updated the commit addressing all your changes except the one related to SourceContent. As I indicated above it is still being used.

pulp_deb/app/models/__init__.py
quba42: We have a big conference today and tomorrow, but I will also try to test the upload scenario soon. Perhaps I can come up with something. My intuition is that it should be possible to avoid this content type, that is never really used _as a content type_. I would like to avoid it, if possible. And since it requires changes to the DB, I would like to avoid it from the get go. I will have a look.

quba42:
The issue has moved to GitHub: https://github.com/pulp/pulp_deb/issues/409

masselstine:
@quba42 I managed to sort through things and remove the dreaded SourceContent. I wasn't sure how to handle the move from redmine to github issue #s so I did some adlib in the commit log. Please let me know if I needs adjusting, my goal was to bring this back to life, details like the commit log are easy enough to get sorted.

quba42:
`CHANGES/8775.feature` needs to be renamed to `CHANGES/409.feature` as a result of the move to GitHub issues. I think that is the only thing still missing from the issue move. I manually linked this PR with the GitHub issue (not sure if the CI would have taken care of that or if it is strictly needed). I am also still new to the new issue tracker...

masselstine:
> `CHANGES/8775.feature` needs to be renamed to `CHANGES/409.feature` as a result of the move to GitHub issues. I think that is the only thing still missing from the issue move. I manually linked this PR with the GitHub issue (not sure if the CI would have taken care of that or if it is strictly needed). I am also still new to the new issue tracker...

I should have caught that, sorry. I made the update and ran the checks, so hopefully good to go now.

pulp_deb/app/models/__init__.py
masselstine: SourceContent is removed from the latest iteration of the change.

pulp_deb/app/models/content.py
masselstine: Rename has been done.

pulp_deb/app/models/content.py
masselstine: Rename has been done.

masselstine:
I will fix up the merge conflicts shortly.

masselstine:
@quba42 any clue why this review still says "Changes requested"?

pulp_deb/app/serializers/content_serializers.py
```diff
@@ -674,3 +704,410 @@ class PackageReleaseComponentSerializer(NoArtifactContentSerializer):
...
+
+class SourcePackageReleaseComponentSerializer(NoArtifactContentSerializer):
+    """
+    A Serializer for SourcePackageReleaseComponent.
+    """
+
+    source_package = DetailRelatedField(
+        help_text="Source package that is contained in release_comonent.",
+        many=False,
+        queryset=ReleaseComponent.objects.all(),
```
Moustafa-Moustafa, line 1101: SourcePackages?

masselstine: I do see a typo above "release_comonent", but I am not understanding your comment, maybe you can expand some. NOTE SourcePackage is a Debian terminology (https://wiki.debian.org/Packaging/SourcePackage), am I missing a pluralization or something else?

daviddavis: This line should be:

```suggestion
        queryset=SourcePackage.objects.all(),
```

Moustafa-Moustafa: Sorry for the confusingly short comment. Thanks David, for the clarification.

Moustafa-Moustafa:
@masselstine, do you plan to add some testing to this PR? 
the pulp_deb team has added new fixtures that should help with writing tests, https://github.com/pulp/pulp_deb/pull/704

daviddavis:
@masselstine we're interested in getting this merged. Could we help collaborate on the tests? I see two tests that might make sense for this PR:

1. A sync test that syncs a repo with source package(s)
2. An upload test that uploads source package(s) into Pulp

We're much more interested in the second use case. Maybe that's a test we could contribute?

masselstine:
> @masselstine we're interested in getting this merged. Could we help collaborate on the tests? I see two tests that might make sense for this PR:
> 
> 1. A sync test that syncs a repo with source package(s)
> 2. An upload test that uploads source package(s) into Pulp
> 
> We're much more interested in the second use case. Maybe that's a test we could contribute?

I will not turn away any help. I was planning on spending some time on this this afternoon and over the weekend so I could have an update next week which would include some testing. I have been doing (1) and (2) in my own manual tests, sync'ing with a small local Debian repository (funnily enough using Aptly to present the repo), and using the source package for 'bc' for (2). For the sake of collaboration would you like to do (1) or (2), I will take whatever remains.

Moustafa-Moustafa:
> > @masselstine we're interested in getting this merged. Could we help collaborate on the tests? I see two tests that might make sense for this PR:
> > 
> > 1. A sync test that syncs a repo with source package(s)
> > 2. An upload test that uploads source package(s) into Pulp
> > 
> > We're much more interested in the second use case. Maybe that's a test we could contribute?
> 
> I will not turn away any help. I was planning on spending some time on this this afternoon and over the weekend so I could have an update next week which would include some testing. I have been doing (1) and (2) in my own manual tests, sync'ing with a small local Debian repository (funnily enough using Aptly to present the repo), and using the source package for 'bc' for (2). For the sake of collaboration would you like to do (1) or (2), I will take whatever remains.

We have done some testing for the upload scenario, So I guess we can help with (2).

masselstine:
@daviddavis @Moustafa-Moustafa @quba42 I have updated the commit to resolve the merge conflict, address the s/ReleaseComponent/SourcePackage/ in SourcePackageReleaseComponentSerializer() and fix the comment typo.

I have started to work on adding some testing to sync with a repo that contains source packages, as discussed earlier. I am thinking of making the tests in their own commit, but if you folks feel otherwise please let me know.

masselstine:
I can reproduce the CI errors and will send an update to ensure they continue to function as before.

hstct:
Hi there!

> I have started to work on adding some testing to sync with a repo that contains source packages, as discussed earlier. I am thinking of making the tests in their own commit, but if you folks feel otherwise please let me know.

Great work so far, just wanted to let you know if you want to write sync tests please use the `test_sync_optimize.py` functional tests as a reference. Since they are using the pytest framework and work with the new local test data (#704). 

Just mentioning it because this feature seems like it absolutely needs its own test data repository. If you need some assistance with it let me know.

I'm currently in the process of converting the remaining sync tests (#734). It should not be a problem really if your tests are using the new framework but I can hold off my PR for a while just to avoid some merge messes. It is not high priority.

hstct:
> CORRECTION: I can't reproduce this. My error was different and caused by a polluted database. After cleaning up the database both of the failed tests pass. I also see that historically these tests may be failing regularly (for instance https://github.com/pulp/pulp_deb/actions/runs/4522519027/jobs/7965046604#step:13:1995) so don't believe this is associated with my change.

Yea seems unrelated and CI issue will take a look at it tomorrow

masselstine:
@Moustafa-Moustafa @daviddavis I have started on the tests. I have put my work in progress at https://github.com/masselstine/pulp_deb/tree/src_tests_wip. For now I have mocked up the features but will eventually push to pulp-features once we have the right data.

Moustafa-Moustafa:
@quba42, Here's the set of pulp API requests I used to manually test this feature locally,
$REPO_HREF=repositories/deb/apt/b90dcbc2-7e09-444e-b0ce-4da77b68be11
http --form post $BASE_ADDR/pulp/api/v3/artifacts/ file@"../../hello_2.10.orig.tar.gz"
http --form post $BASE_ADDR/pulp/api/v3/artifacts/ file@"../../hello_sources/hello_2.10-2ubuntu2.debian.tar.xz"
http --form post $BASE_ADDR/pulp/api/v3/artifacts/ file@"../../hello_sources/hello_2.10-2ubuntu2.dsc"
PACKAGE_HREF=/pulp/api/v3/content/deb/source_packages/0536e1ee-8f03-405d-941b-e61c6332da65/
SOURCE_RELEASE_COMPONENT=$(http ${BASE_ADDR}/pulp/api/v3/content/deb/source_release_components/ source_package=$PACKAGE_HREF release_component=$COMP_HREF | jq -r .pulp_href)
http ${BASE_ADDR}/pulp/api/v3/$REPO_HREF/modify/ add_content_units:="[\"$PACKAGE_HREF\", \"$SOURCE_RELEASE_COMPONENT\"]"

masselstine:
Unfortunately I found an issue while creating tests. I was always using policy="on_demand" when setting up a remote in my tests. This has always worked well. 

While writing a testcase using the fixtures, which doesn't use  policy="on_demand",  I noticed that the downloader is always decompressing files. This doesn't cause the existing tests to fail as we are generating the Packages.gz and similar (though the logs are filled with mismatch checksum errors as the checksums are for the uncompressed files). But unlike the current tests which just continue to work OK, for a source repository *orig.tar.gz files will decompress and fail to download due to mismatch checksum.

I have been trying to figure out how to shutoff the auto_decompress for the source files, but so far I have failed. @Moustafa-Moustafa or @daviddavis if you know how to accomplish this it would be helpful.

If you download my WIP test branch you can do the following hack by adding 'sync_sources' to the _deb_remote_factory()

`    def _deb_remote_factory(repo_name=DEB_FIXTURE_STANDARD_REPOSITORY_NAME, **kwargs):
        url = deb_fixture_server.make_url(repo_name)
        return gen_object_with_cleanup(
            apt_remote_api, gen_local_deb_remote(url=str(url), sync_sources=True, **kwargs)
        )
`

and then running the following you will see the mismatch issue directly and in the logs
pytest -v -r sx --color=yes --pyargs pulp_deb.tests.functional -m 'parallel' -k test_download_content

daviddavis:
Ah interesting. Does this happen only when running the tests?

masselstine:
> Ah interesting. Does this happen only when running the tests?

Testing with similar repo as the 'debian' fixtures, served with Aptly and with the policy set to 'immediate' I see no checksum mismatch being reported in the logs. The orig.tar.gz is downloadable without issue. So barring any configs which I might not be capturing, this appears to only be seen in testing. It is Friday so I reserve the right to be corrected.

daviddavis:
What's the Content-Encoding header when the file is served by Aptly? I am guessing the test fixture server has a different Content-Encoding header since [it looks like aiohttp will decompress the file depending on the response's `Content-Encoding`](https://github.com/aio-libs/aiohttp/issues/4462).

masselstine:
> What's the Content-Encoding header when the file is served by Aptly? I am guessing the test fixture server has a different Content-Encoding header since [it looks like aiohttp will decompress the file depending on the response's `Content-Encoding`](https://github.com/aio-libs/aiohttp/issues/4462).

I set aptly aside and simply served the content with a ruby one-liner "ruby3.0 -run -e httpd pulp_deb/tests/functional/data -p 8081" and hardcoded the url in _deb_remote_factory to url = "http://my-machine:8081/debian/" and there are no longer signs of decompression, either with the test as-is or with the 'sync_sources=True' added to the call to gen_object_with_cleanup(). So this 100% appears to be in the way the fixture is served. I started looking into Content-Encoding set to 'identity' as the aiohttp version doesn't support setting auto_decompress. 

I will try to keep an eye out on IRC for coordination. I was hoping adding the testing was going to be fairly easy but should have expected some curveballs.

hstct:
@masselstine can you add the commit with your testing to this PR? This would make it easier to follow.

For now I assume its the same status as you have linked here (https://github.com/masselstine/pulp_deb/tree/src_tests_wip)

Since the `test_sync.py` on the main branch is still using the old test framework I would advise to put your test in the `test_sync_optimize.py` file for now (both files will be combined once #734 gets merged). `test_sync_optimize.py` uses pytest already and you can use it as a guideline on how to call the pytest fixtures as intended.

I say this because I noticed you did something like this:
`@pytest.mark.skip(reason="Requires https://fixtures.pulpproject.org update")`

With #704 we introduced a change that will allow pytest to serve its own testdata locally. We use [this fixture](https://github.com/pulp/pulp_deb/blob/main/pulp_deb/tests/functional/conftest.py#L305-L310) which is provided by pulpcore to serve the local data and the [deb_remote_factory()](https://github.com/pulp/pulp_deb/blob/main/pulp_deb/tests/functional/conftest.py#L130-L142) uses it to create build the url to the data with the default repository being `/debian/`.

I'm not certain if this will fix the checksum mismatch error. I noticed they exist with other tests as well but they didn't cause the tests to fail curious to see if the problem still persists afterwards.

masselstine:
> @masselstine can you add the commit with your testing to this PR? This would make it easier to follow.
> 
> For now I assume its the same status as you have linked here (https://github.com/masselstine/pulp_deb/tree/src_tests_wip)
> 
> Since the `test_sync.py` on the main branch is still using the old test framework I would advise to put your test in the `test_sync_optimize.py` file for now (both files will be combined once #734 gets merged). `test_sync_optimize.py` uses pytest already and you can use it as a guideline on how to call the pytest fixtures as intended.
> 
> I say this because I noticed you did something like this: `@pytest.mark.skip(reason="Requires https://fixtures.pulpproject.org update")`
> 
> With #704 we introduced a change that will allow pytest to serve its own testdata locally. We use [this fixture](https://github.com/pulp/pulp_deb/blob/main/pulp_deb/tests/functional/conftest.py#L305-L310) which is provided by pulpcore to serve the local data and the [deb_remote_factory()](https://github.com/pulp/pulp_deb/blob/main/pulp_deb/tests/functional/conftest.py#L130-L142) uses it to create build the url to the data with the default repository being `/debian/`.
> 
> I'm not certain if this will fix the checksum mismatch error. I noticed they exist with other tests as well but they didn't cause the tests to fail curious to see if the problem still persists afterwards.

@hstct I have created #742 to hopefully prevent the observed issue from being intertwined with this work. The checksum mismatch on *.gz files is encountered with none of the change for source support. It came up here as I was attempting to add a suite of tests for this work when I observed the issue, which is fatal to the source tests as deb sources will often include *.gz files, such as *.orig.gz.

I am aware of the distinction between the externally web hosted https://fixtures.pulpproject.org fixtures and the work done in #704 to serve test data locally. It is when I started to look at adding a test that would use the locally served test data that I observed the issue with .gz content being decompressed. I do not want to include my test WIP here as it is not consumable until we can solve #742, as I actually have not passed the hacking phase where the code is hardly presentable.

As I noted in the new Issue the current suite of tests pass only because the *.gz files that are sync'd are inconsequential to the test passing. It is only when *.gz files that are not dynamically re-creatable, such as source content that this because an issue beyond filling up the log files with checksum mismatch reports.

masselstine:
I have a working rebase to resolve the conflicts. With #742 now closed I can get the test added and issue a new PR in the coming days.

masselstine:
I am extremely sorry for the delay in getting the update sent along with the completed testing. I will be able to do this next week.

Moustafa-Moustafa:
@masselstine , I have my tests for uploading a source package ready [here](https://github.com/pulp/pulp_deb/commit/bf7a78aa8153fd47c676470a622865ded8ec7c94#diff-9f47182cf5bf2cc17f6f4faca4a577112d4dc77dc1d6fd9521abf7f85b8fd4d5). I'm planning to send out a PR against your fork of pulp_deb, and then you can update this PR once my tests are merged to your fork. My fork is apparently a few commits ahead of yours, so once you sync your fork, I will send out my PR against it.

masselstine:
@Moustafa-Moustafa I am finally back for 3 weeks of business travel and 5 minutes away from updating my PR. You should be able to grab it within minutes.

Moustafa-Moustafa:
@masselstine , I created this [PR](https://github.com/masselstine/pulp_deb/pull/1) against your fork with my upload tests. Once it's merged, please update the main PR with the new tests.

Moustafa-Moustafa:
> > @masselstine , I created this [PR](https://github.com/masselstine/pulp_deb/pull/1) against your fork with my upload tests. Once it's merged, please update the main PR with the new tests.
> 
> Ha, we seem to often be working at the same time. I just updated my PR as somehow the migrations were incomplete. The migrations check should pass now, along with the other CI tests.

Awesome, glad to see this moving forward. I resolved the conflicts on my PR and it should be ready to review/merge.

masselstine:
@quba42 the s3, azure and pulp tests are failing due to checksum mismatch. This is happening because my update to pulp_smash does not appear to be present, ie. .gz files are being decompressed by the http client. How do we get pulp_smash updated for the CI tests here? I will fixup the [noissue] in my commit log for the tests and review the failure for the lowerbounds test.

masselstine:
> @quba42 the s3, azure and pulp tests are failing due to checksum mismatch. This is happening because my update to pulp_smash does not appear to be present, ie. .gz files are being decompressed by the http client. How do we get pulp_smash updated for the CI tests here? I will fixup the [noissue] in my commit log for the tests and review the failure for the lowerbounds test.

Actually, on further review it does appear that the CI scripts are pulling the latest pulp-smash during the install which includes my change. This seems to be confirmed by many of the *.gz files being fetched without checksum mismatches, thus no unexpected deflation for Packages.gz and Sources.gz and such. Of course in my Vagrant machine (pulp_installer_pulp3-source-fedora36) the tests run fine. So I am going to have to attempt to replicate the Github CI scripts to figure out what is going on here. Yet another hurdle to getting this work merged :(

hstct:
> Actually, on further review it does appear that the CI scripts are pulling the latest pulp-smash during the install which includes my change. This seems to be confirmed by many of the *.gz files being fetched without checksum mismatches, thus no unexpected deflation for Packages.gz and Sources.gz and such. Of course in my Vagrant machine (pulp_installer_pulp3-source-fedora36) the tests run fine. So I am going to have to attempt to replicate the Github CI scripts to figure out what is going on here. Yet another hurdle to getting this work merged :(

@masselstine fwiw: I just successfully ran your tests on an https://github.com/pulp/oci_env dev environment which should be close to what the github CI actually runs. No idea why github is complaining.

pulp_deb/app/migrations/0023_sourcepackage_sourcepackagereleasecomponent_and_more.py
```diff
@@ -0,0 +1,126 @@
+# Generated by Django 4.2.1 on 2023-06-02 15:17
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("core", "0107_distribution_hidden"),
```
hstct, line 9: why do we depend on this core migration? I stumbled upon this while testing your branch and my dev environment and I hadn't pulled the latest release yet.

quba42:
> @masselstine fwiw: I just successfully ran your tests on an https://github.com/pulp/oci_env dev environment which should be close to what the github CI actually runs. No idea why github is complaining.

In that case I feel like we might put this on tomorrows [Pulp Open Floor meeting agenda](https://hackmd.io/SVCMjpwXTfOMqF2OeyyLRw) to get more eyes on it. Hopefully that will at the least result in some pointers how to keep debugging this.

pulp_deb/app/migrations/0023_sourcepackage_sourcepackagereleasecomponent_and_more.py
masselstine: Is this unusual? I see this isn't infrequent, similar instances include the recently added 022_alter_aptdistribution_distribution_ptr_and_more.py. I definitely don't see a direct relationship here but the new models do depend on 'release' and 'release' depends on 'distribution' which was modified by "0107_distribution_hidden".

hstct: My bad! You are of course right I didn't see the indirect relation with the `Release`!

masselstine:
> > @masselstine fwiw: I just successfully ran your tests on an https://github.com/pulp/oci_env dev environment which should be close to what the github CI actually runs. No idea why github is complaining.
> 
> In that case I feel like we might put this on tomorrows [Pulp Open Floor meeting agenda](https://hackmd.io/SVCMjpwXTfOMqF2OeyyLRw) to get more eyes on it. Hopefully that will at the least result in some pointers how to keep debugging this.

I have made some progress by adding a 'tmate' step to the CI workflow. As with the previous aiohttp issue encountered while working on this PR I have to do further investigation and reading to determine the proper fix. Stay tuned.

masselstine:
> > > @masselstine fwiw: I just successfully ran your tests on an https://github.com/pulp/oci_env dev environment which should be close to what the github CI actually runs. No idea why github is complaining.
> > 
> > 
> > In that case I feel like we might put this on tomorrows [Pulp Open Floor meeting agenda](https://hackmd.io/SVCMjpwXTfOMqF2OeyyLRw) to get more eyes on it. Hopefully that will at the least result in some pointers how to keep debugging this.
> 
> I have made some progress by adding a 'tmate' step to the CI workflow. As with the previous aiohttp issue encountered while working on this PR I have to do further investigation and reading to determine the proper fix. Stay tuned.

Sorry. I have been busy and lacked time to continue my investigation. Had some time tonight though.

This issue is related to aiohttp handling, yes, again. The GET request is being made with only "gzip" and "deflate" in the Accept-Encoding. We can observer this via

`sudo tcpdump -i lo -A -s 10240 '(((ip[2:2] - ((ip[0]&0xf)<<2)) - ((tcp[12]&0xf0)>>2)) != 0)' | egrep --line-buffered "^........(GET |HTTP\/|POST |HEAD )|^[A-Za-z0-9-]+: " | sed -r 's/^........(GET |HTTP\/|POST |HEAD )/\n\1/g' | tee outme`

this will limit the server's ability to decide that the content is already compressed and send the data using 'identity' encoding. (see "the server may choose not to compress the body of a response if the identity value is also acceptable" in https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding). As was discussed previously aiohttp.client will by default automatically decompress "gzip" content (see auto_decompress  in https://docs.aiohttp.org/en/stable/client_reference.html)

It is unclear how this is working in some cases, including in my vagrant development box. But in order to have consistent results there needs to be at least one change, and most likely two changes.

1.  Pulpcore DownloaderFactory::_make_aiohttp_session_from_remote() will allow the remote to define headers, so we can tack "Accept-Encoding": "identity, gzip, deflate" onto the headers in the pulp_deb remote setup. This will allow the server to choose to send via "identity" encoding, thus not triggering the client's auto_decompression.
2. If the server code is already smart enough to recognize already compressed files and send with "identity" encoding, then there is nothing more to do. If not pulp-smash will need to be enhanced to recognize already compressed files and avoid attempting to compress them a second time and use "gzip" encoding.

I am in the process of implementing (1). I am hoping (2) will not be needed based on the fix I already had merged into pulp-smash.

masselstine:
> > > > @masselstine fwiw: I just successfully ran your tests on an https://github.com/pulp/oci_env dev environment which should be close to what the github CI actually runs. No idea why github is complaining.
> > > 
> > > 
> > > In that case I feel like we might put this on tomorrows [Pulp Open Floor meeting agenda](https://hackmd.io/SVCMjpwXTfOMqF2OeyyLRw) to get more eyes on it. Hopefully that will at the least result in some pointers how to keep debugging this.
> > 
> > 
> > I have made some progress by adding a 'tmate' step to the CI workflow. As with the previous aiohttp issue encountered while working on this PR I have to do further investigation and reading to determine the proper fix. Stay tuned.
> 
> Sorry. I have been busy and lacked time to continue my investigation. Had some time tonight though.
> 
> This issue is related to aiohttp handling, yes, again. The GET request is being made with only "gzip" and "deflate" in the Accept-Encoding. We can observer this via
> 
> `sudo tcpdump -i lo -A -s 10240 '(((ip[2:2] - ((ip[0]&0xf)<<2)) - ((tcp[12]&0xf0)>>2)) != 0)' | egrep --line-buffered "^........(GET |HTTP\/|POST |HEAD )|^[A-Za-z0-9-]+: " | sed -r 's/^........(GET |HTTP\/|POST |HEAD )/\n\1/g' | tee outme`
> 
> this will limit the server's ability to decide that the content is already compressed and send the data using 'identity' encoding. (see "the server may choose not to compress the body of a response if the identity value is also acceptable" in https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding). As was discussed previously aiohttp.client will by default automatically decompress "gzip" content (see auto_decompress in https://docs.aiohttp.org/en/stable/client_reference.html)
> 
> It is unclear how this is working in some cases, including in my vagrant development box. But in order to have consistent results there needs to be at least one change, and most likely two changes.
> 
> 1. Pulpcore DownloaderFactory::_make_aiohttp_session_from_remote() will allow the remote to define headers, so we can tack "Accept-Encoding": "identity, gzip, deflate" onto the headers in the pulp_deb remote setup. This will allow the server to choose to send via "identity" encoding, thus not triggering the client's auto_decompression.
> 2. If the server code is already smart enough to recognize already compressed files and send with "identity" encoding, then there is nothing more to do. If not pulp-smash will need to be enhanced to recognize already compressed files and avoid attempting to compress them a second time and use "gzip" encoding.
> 
> I am in the process of implementing (1). I am hoping (2) will not be needed based on the fix I already had merged into pulp-smash.

Unfortunately pulp-smash continue to ignore "Accept-Encoding" as we can see when running tcpdump to review the traffic in the github CI test container

```
GET /debian/pool/asgard/h/hello/hello_2.10.orig.tar.gz HTTP/1.1
Host: 127.0.0.1:54035
User-Agent: pulpcore/3.28.0 (cpython 3.8.13-final0, Linux x86_64) (aiohttp 3.8.4)
Accept-Encoding: identity
Accept: */*

HTTP/1.1 200 OK
Content-Type: application/x-tar
Content-Encoding: gzip
Etag: "176bb4128427e200-b13ba"
Last-Modified: Sat, 24 Jun 2023 21:01:17 GMT
Content-Length: 725946
Accept-Ranges: bytes
Date: Sat, 24 Jun 2023 21:11:32 GMT
Server: Python/3.8 aiohttp/3.8.4
```
So this comes down to poor implementation on the server side of the testcase. I will have to raise another bug against pulp-smash once I have a better idea why this is happening in the GH CI container.

masselstine:
> > > > > @masselstine fwiw: I just successfully ran your tests on an https://github.com/pulp/oci_env dev environment which should be close to what the github CI actually runs. No idea why github is complaining.
> > > > 
> > > > 
> > > > In that case I feel like we might put this on tomorrows [Pulp Open Floor meeting agenda](https://hackmd.io/SVCMjpwXTfOMqF2OeyyLRw) to get more eyes on it. Hopefully that will at the least result in some pointers how to keep debugging this.
> > > 
> > > 
> > > I have made some progress by adding a 'tmate' step to the CI workflow. As with the previous aiohttp issue encountered while working on this PR I have to do further investigation and reading to determine the proper fix. Stay tuned.
> > 
> > 
> > Sorry. I have been busy and lacked time to continue my investigation. Had some time tonight though.
> > This issue is related to aiohttp handling, yes, again. The GET request is being made with only "gzip" and "deflate" in the Accept-Encoding. We can observer this via
> > `sudo tcpdump -i lo -A -s 10240 '(((ip[2:2] - ((ip[0]&0xf)<<2)) - ((tcp[12]&0xf0)>>2)) != 0)' | egrep --line-buffered "^........(GET |HTTP\/|POST |HEAD )|^[A-Za-z0-9-]+: " | sed -r 's/^........(GET |HTTP\/|POST |HEAD )/\n\1/g' | tee outme`
> > this will limit the server's ability to decide that the content is already compressed and send the data using 'identity' encoding. (see "the server may choose not to compress the body of a response if the identity value is also acceptable" in https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding). As was discussed previously aiohttp.client will by default automatically decompress "gzip" content (see auto_decompress in https://docs.aiohttp.org/en/stable/client_reference.html)
> > It is unclear how this is working in some cases, including in my vagrant development box. But in order to have consistent results there needs to be at least one change, and most likely two changes.
> > 
> > 1. Pulpcore DownloaderFactory::_make_aiohttp_session_from_remote() will allow the remote to define headers, so we can tack "Accept-Encoding": "identity, gzip, deflate" onto the headers in the pulp_deb remote setup. This will allow the server to choose to send via "identity" encoding, thus not triggering the client's auto_decompression.
> > 2. If the server code is already smart enough to recognize already compressed files and send with "identity" encoding, then there is nothing more to do. If not pulp-smash will need to be enhanced to recognize already compressed files and avoid attempting to compress them a second time and use "gzip" encoding.
> > 
> > I am in the process of implementing (1). I am hoping (2) will not be needed based on the fix I already had merged into pulp-smash.
> 
> Unfortunately pulp-smash continue to ignore "Accept-Encoding" as we can see when running tcpdump to review the traffic in the github CI test container
> 
> ```
> GET /debian/pool/asgard/h/hello/hello_2.10.orig.tar.gz HTTP/1.1
> Host: 127.0.0.1:54035
> User-Agent: pulpcore/3.28.0 (cpython 3.8.13-final0, Linux x86_64) (aiohttp 3.8.4)
> Accept-Encoding: identity
> Accept: */*
> 
> HTTP/1.1 200 OK
> Content-Type: application/x-tar
> Content-Encoding: gzip
> Etag: "176bb4128427e200-b13ba"
> Last-Modified: Sat, 24 Jun 2023 21:01:17 GMT
> Content-Length: 725946
> Accept-Ranges: bytes
> Date: Sat, 24 Jun 2023 21:11:32 GMT
> Server: Python/3.8 aiohttp/3.8.4
> ```
> 
> So this comes down to poor implementation on the server side of the testcase. I will have to raise another bug against pulp-smash once I have a better idea why this is happening in the GH CI container.

So the CI container is making use of the pytest fixtures in pulpcore, which don't have the fix for #742. Without the change aiohttp.server will ignore the Accept-Encoding as documented in my last comment. When running inside the Vagrant box we are using the pulp-smash fixtures, which have the fix. This explains why we are not always able to reproduce this issue, it depends on how things are configured that result in pulpcore fixtures being used (busted) or pulp-smash fixtures being used (fixed).

Before I continue to implement a similar fix to #742 on pulpcore I would like to understand more about why pulp-smash and pulpcore seem to duplicate fixtures functionality and which one is expected to be used. The situation seems to be just slightly confusing and maybe @quba42 or @daviddavis you can shed some light on the situation. In the meantime I will read some of the commit logs to see if some light can be shed.

daviddavis:
@masselstine I haven't really worked with the Pulp fixtures before so I am not sure why there are duplicate fixtures in pulp-smash and pulpcore. I think @quba42 @hstct or @mdellweg might know.

quba42:
I think others know the details better than me, but my understanding is that fixtures are moving into pulpcore, and we are working to get rid of all remaining dependencies to pulp-smash. => i.e. pulp-smash is deprecated.

mdellweg:
> > I think others know the details better than me, but my understanding is that fixtures are moving into pulpcore, and we are working to get rid of all remaining dependencies to pulp-smash. => i.e. pulp-smash is deprecated.
> 
> Alright. I am going to have to prepare and submit my fixtures changes to pulpcore in that case then, as well as figure out how to adapt my Vagrant development box to use the pulpcore fixtures instead of pulp-smash in order to ensure everything is in order. Please standby, again.

Yes, it's about stopping to use pulp-smash altogether. The issue is, until you do not even install smash anymore, there is no way to tell pytest which one to pick up. So we need these fixes in both, i believe.

mdellweg:
Removing pulp_smash from the plugin ci will need a bit more work than just not installing it, I'm sorry. But if we have a pulp-smash fixture with a bugfix, we may want the same in pulpcore for sure.

masselstine:
> Removing pulp_smash from the plugin ci will need a bit more work than just not installing it, I'm sorry. But if we have a pulp-smash fixture with a bugfix, we may want the same in pulpcore for sure.
> 
> Sorry for the inconvenience. When we planned to susset pulp-smash we didn't anticipate for it to take years.

Up to this point I was only able to work for 15 minutes at a time with the CI container on GH in a tmate session. Now that I know the root cause I will be able to reproduce this in my dev env and test changes to the pulpcore fixtures. Attempting to just set the Content-Type as I did in pulp-smash raised another issue, so further work needs to be done before I can submit a fix to pulpcore. I will go ahead and raise a defect and try to get a fix out soon, and link it to this PR.

masselstine:
> > Removing pulp_smash from the plugin ci will need a bit more work than just not installing it, I'm sorry. But if we have a pulp-smash fixture with a bugfix, we may want the same in pulpcore for sure.
> > Sorry for the inconvenience. When we planned to susset pulp-smash we didn't anticipate for it to take years.
> 
> Up to this point I was only able to work for 15 minutes at a time with the CI container on GH in a tmate session. Now that I know the root cause I will be able to reproduce this in my dev env and test changes to the pulpcore fixtures. Attempting to just set the Content-Type as I did in pulp-smash raised another issue, so further work needs to be done before I can submit a fix to pulpcore. I will go ahead and raise a defect and try to get a fix out soon, and link it to this PR.

For anyone following along. It is possible to reproduce this issue using the vagrant development box. It comes down to pytest plugin load order. In the vagrant box **pulp_smash** is coming after **pulpcore**, in the CI test runs **pulp_smash** is coming before **pulpcore**. The **last** plugin included gets the last word, or in this case the implementation of functions such as _add_recording_route()_.

This probably deserves a defect of its own as a somewhat random plugin load order such as this is going to result in extremely frustrating problem solving such as what has had to be done here. We should most likely attempt to enforce a known order.

Should you want to reproduce the issue being hit in CI run the following (note the **-p** option)

```
pytest -v -r sx --color=yes -p pulp_smash --suppress-no-test-exit-code pulp_deb/tests/functional/api/test_sync.py::test_sync -m parallel
```

The **-p**  will act as an early load mechanism for the **pulp_smash** plugin, an thus **pulpcore** will be forced to be loaded after **pulp_smash**. And since **pulpcore** gets the last word and has yet to have been patched to properly handle Accept-Encoding (see #742) we hit the checksum mismatch failure.

quba42:
@masselstine Can we simply add your fix for pulp-smash to pulpcore as well? Is there already a PR for that?

I suspect that the long term solution to the load order issue you describe will be to drop pulp-smash (which is planned, but we are not their yet)...

mdellweg:
> @masselstine Can we simply add your fix for pulp-smash to pulpcore as well? Is there already a PR for that?
> 
> I suspect that the long term solution to the load order issue you describe will be to drop pulp-smash (which is planned, but we are not their yet)...

I absolutely agree. Dropping pulp-smash was the plan all along and we of thought the transition would be easier if we copied the testing facilities over to pulpcore instead of giving them new names. Just the fact that the transition is taking much too long and the implementations started to diverge in the meantime is what troubles you. I am very sorry about that.
Not sure i really want to suggest that: Whatever is in the local conftest.py will win over every single pytest plugin...
So in theory reimplementing the `add_recording_route` fixture there may or may not solve your problem.

masselstine:
@quba42 yes. Now that I have a local reproducer I can prepare the same or similar fix for pulpcore. The change is as valid there as it was in pulp-smash, ie. no web server should be ignoring the Accept-Encoding of a client. With the reproducer available I can ensure my change is fully vetted before I send it for merge.

Look for me to raise the issue and send a PR over the next week or so to the pulpcore project.

masselstine:
Folks. Sorry for the silence. Summer vacations and other obligations. I will be back on this shortly.

masselstine:
> > Folks. Sorry for the silence. Summer vacations and other obligations. I will be back on this shortly.
> 
> @masselstine no worries, but just a heads up. We have two PRs I expect to get merged possibly next week. One that will remove the `pulp-smash` dependency (#851) and another one that will refactor the tests (#857). These will probably be the last changes to the tests for (hopefully) a while so you will need to adapt yours a little bit.
> 
> I actually thought on helping out and port your change from `pulp-smash` to `pulpcore` and I have the change locally but had to work on other things and couldn't quite test it yet. I can continue to work on this eventually (but will take some time). Let me know if I should or if you will continue working on it. However whoever manages to open up a PR for `pulpcore` it would great if we can convince the team to backport this change to at least `pulpcore 3.28` so we don't have to change the plugin requirements yet again 🙂

I am not fuss on who ports the change, so @hstct  if you want to go for it, feel free. I will most likely get to this next week should you not get something sent by then.

hstct:
@masselstine I finally got around to open the PR in pulpcore. https://github.com/pulp/pulpcore/pull/4494

I tested it locally and hopefully this will resolve the test issue.

I had to build my own test data with source packages since we updated the `/debian/` test data repository at some point. I got a workflow for creating a dummy source package so in that case we don't need to include a random ubuntu package to the repository.

masselstine:
> @masselstine I finally got around to open the PR in pulpcore. [pulp/pulpcore#4494](https://github.com/pulp/pulpcore/pull/4494)
> 
> I tested it locally and hopefully this will resolve the test issue.
> 
> I had to build my own test data with source packages since we updated the `/debian/` test data repository at some point. I got a workflow for creating a dummy source package so in that case we don't need to include a random ubuntu package to the repository.

Thank you so much for preparing that PR. I have been completely absent here. I wanted to get things done but simply had no time. At any rate, no excuse and I apologize for this.

I will find time to rebase my change and see about using your new dummy source package. Thanks for hopefully unsticking this for the last time (I will believe it when I see it). I assume once your PR is merged we might have to do some version updates to ensure the CI tests use the updates?

hstct:
> Thank you so much for preparing that PR. I have been completely absent here. I wanted to get things done but simply had no time. At any rate, no excuse and I apologize for this.

No worries :)

> I will find time to rebase my change and see about using your new dummy source package.

I will open a PR with updated test data probably tomorrow which will include a source package repo.

> Thanks for hopefully unsticking this for the last time (I will believe it when I see it). I assume once your PR is merged we might have to do some version updates to ensure the CI tests use the updates?

The change needs to be backported and released in pulpcore but that usually happens regularly nowadays. Otherwise we need to update the pulpcore requirements. I'm hoping for the first option but both will work.

daviddavis:
Awesome, looks like https://github.com/pulp/pulpcore/pull/4494 is merged and has been backported to 3.28. 🎉

hstct:
> Awesome, looks like [pulp/pulpcore#4494](https://github.com/pulp/pulpcore/pull/4494) is merged and has been backported to 3.28. 🎉

That's true but it needs some maintenance in pulp deb before the CIs are running again properly :)

As well I opened up a PR that adds source package test data https://github.com/pulp/pulp_deb/pull/904

This will add two source packages to the standard test repository (`/debian/`) so the test included here probably need slight adaption.

hstct:
@masselstine the test data PR #904 is merged. Let's hope the github CI will not find anything else to complain :crossed_fingers:

hstct (approved):
Seems everything is in order now 🎆

hstct:
Great job! 🎆

PR #297, opened 2021-06-22
Add unit tests to test component and architecture filtering

quba42:
[noissue]

Having recently reworked these filter functions several times in a row to accommodate all kinds of edge cases, I feel like they would really benefit from some simple unit tests!

### DIFF
--- a/pulp_deb/tests/unit/test_sync_helpers.py
+++ b/pulp_deb/tests/unit/test_sync_helpers.py
@@ -1,0 +1,191 @@
+from django.test import TestCase
+
+from pulp_deb.app.tasks.synchronizing import _filter_split_architectures, _filter_split_components
+
+
+class TestArchitectureFiltering(TestCase):
+    """
+    Tests common as well as edge cases handled by the _filter_split_architectures function.
+    """
+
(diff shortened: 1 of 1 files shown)


pulp_deb/tests/unit/test_sync_helpers.py
```diff
@@ -0,0 +1,96 @@
...
+
+        self.assertEqual(
+            _filter_split_architectures(release_file_architectures, "bogus", "stable"),
+            ["all"],
+        )
+
+        release_file_architectures = "arm64 armel armhf i386 mips64el mipsel amd64"
+
+        self.assertEqual(
+            _filter_split_architectures(release_file_architectures, "bogus", "stable"),
```
mdellweg, line 94: Is the `release` argument even doing something in this function?

quba42: It is only used by the log message. It may do some actual work once my flat repo format changes are added on top.

quba42:
Added tests for component filtering.

Still missing: Test for the log messages.

mdellweg (approved):

PR #298, opened 2021-06-23
Make flat repo syncs more robust

quba42:
Closes #7673
https://pulp.plan.io/issues/7673

With this change flat repo syncs will work even if the upstream Release
file contains no "Components" or "Architectures" fields. In addition we
are ensuring flat repo syncs do not attempt to sync installer files or
packages.

### DIFF
--- a/pulp_deb/app/models/content.py
+++ b/pulp_deb/app/models/content.py
@@ -72,10 +72,9 @@ class PackageIndex(Content):
     """
     The "PackageIndex" content type.
 
-    This model represents the Packages file for a specific
-    component - architecture combination.
-    It's artifacts should include all (non-)compressed versions
-    of the upstream Packages file.
+    This model represents the Packages file(s) for a specific component - architecture combination
+    (or an entire flat repo). The artifacts will always include the uncompressed Packages file, as
+    well as any compressed package indices using an archive format supported by pulp_deb.
     """
 
     TYPE = "package_index"
--- a/pulp_deb/app/tasks/synchronizing.py
+++ b/pulp_deb/app/tasks/synchronizing.py
@@ -101,6 +101,19 @@ class NoPackageIndexFile(Exception):
     pass
 
 
+class MissingReleaseFileField(Exception):
+    """
+    Exception signifying that the upstream release file is missing a required field.
+    """
+
+    def __init__(self, distribution, field, *args, **kwargs):
+        """
+        The upstream release file is missing a required field.
+        """
+        message = "The release file for distribution '{}' is missing the required field '{}'."
+        super().__init__(_(message).format(distribution, field), *args, **kwargs)
+
+
 def synchronize(remote_pk, repository_pk, mirror):
     """
     Sync content from the remote repository.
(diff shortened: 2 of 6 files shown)


pulp_deb/app/models/content.py
```diff
@@ -72,10 +72,9 @@ class PackageIndex(Content):
     """
     The "PackageIndex" content type.
 
-    This model represents the Packages file for a specific
-    component - architecture combination.
-    It's artifacts should include all (non-)compressed versions
-    of the upstream Packages file.
+    This model represents the Packages file(s) for a specific component - architecture combination
+    (or an entire falt repo). The artifacts will always include the uncompressed Packages file, as
```
mdellweg, line 76:
```suggestion
    (or an entire flat repo). The artifacts will always include the uncompressed Packages file, as
```

quba42: Thanks, fixed!

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -326,18 +342,33 @@ async def run(self):
...
                         release_file.codename = release_file_dict["Codename"]
                     if "suite" in release_file_dict:
                         release_file.suite = release_file_dict["Suite"]
-                    release_file.components = release_file_dict["Components"]
-                    release_file.architectures = release_file_dict["Architectures"]
+
+                    if "components" in release_file_dict:
+                        release_file.components = release_file_dict["Components"]
+                    elif distribution[-1] == "/":
+                        release_file.components = ""
```
quba42, line 361:
```suggestion
                        release_file.components = "flat-repo-component"
```
I am unsure about this one. How much effort should we perform, to respect upstream Release file fields like "Components", that don't actually have any well defined meaning in the flat repo case, and could contain wildly misleading nonsense? See also line 237 above. Similar question there.

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -326,18 +342,33 @@ async def run(self):
...
+                    else:
+                        raise MissingReleaseFileField(distribution, "Components")
+
+                    if "architectures" in release_file_dict:
+                        release_file.architectures = release_file_dict["Architectures"]
+                    elif distribution[-1] == "/":
+                        release_file.architectures = ""
```
quba42, line 368: Unlike "Components", "Architectures" still has a useful meaning in the flat repo case. However, upstream repos don't reliably provide it. The current implementation only provides the Architectures field if the upstream repo does as well.
Since we later generate the actual list of Architectures from the contained packages, we could add this information later, perhaps in a new pipeline stage?

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -485,31 +517,41 @@ async def _handle_distribution(self, distribution):
...
+
+        # Create release architectures (unless we are handling a flat repo)
+        # release_file.architectures is unreliale for flat repos, so we need to delay until we can
+        # extract this information from the package index.
+        architectures = (
+            _filter_split_architectures(
+                release_file.architectures, self.remote.architectures, distribution
+            )
+            if distribution[-1] != "/"
+            else []
```
quba42, line 536: More readable as follows?:
```
architectures = []
if distribution[-1] != "/":
    architectures = _filter_split_architectures( release_file.architectures, self.remote.architectures, distribution )
```

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -646,7 +703,9 @@ async def _handle_package_index(
                 await self.put(package_dc)
             except KeyError:
                 log.warning(_("Ignoring invalid package paragraph. {}").format(package_paragraph))
-        # Assign packages to this release_component
+
+        # Assign packages to the release component
+        package_architectures = set([])
```
quba42, line 708: For falt repos (where there is only ever one package index), we could use this package_architectures set to sanity check, or even repair the "Architectures" field from the Release file as well?

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -646,7 +703,9 @@ async def _handle_package_index(
...
         for package_future in package_futures:
             package = await package_future.resolution()
```
quba42, line 710: Is this enough to guarantee that the package_future we are considering will actually materialize? Or are we risking considering the "Architecture" of packages that will never materialize in the finished repo?

mdellweg: The resolution will be available when the package_dc passes the `ResolveContentFutures`. Since this happens after the saver stages, the content and all its artifacts will be readily saved to the database. Whether it actually lands in the new repo-version is a question of the later stages. And it should obviously not be dropped before when awaited somewhere.

quba42: I think this is good enough for now. The worst case scenario is that the Pulp repo ends up advertising an architecture it does not have any packages for. This is both totally allowed (i.e. not that bad), and pretty unlikely to happen.

pulp_deb/app/tasks/synchronizing.py
quba42: I should add some logging what I am doing here.

pulp_deb/app/tasks/synchronizing.py
quba42: Added logging here, the repair work itself needs to happen at discussion https://github.com/pulp/pulp_deb/pull/298#discussion_r699014388.

pulp_deb/app/tasks/synchronizing.py
quba42: We are reparing via the creation of ReleaseArchitecture content, but it makes no sense to mess around in the ReleaseFile content. Done.

quba42 (commented):
Successfully tested with the following flat repo remotes:
```
REMOTE_HREF=$(http post "${BASE_ADDR}"/pulp/api/v3/remotes/deb/apt/ name="${ENTITIES_NAME}" distributions="/" url="https://developer.download.nvidia.com/compute/cuda/repos/ubuntu1804/x86_64/" policy="on_demand" | jq -r '.pulp_href')
REMOTE_HREF=$(http post "${BASE_ADDR}"/pulp/api/v3/remotes/deb/apt/ name="${ENTITIES_NAME}" distributions="lmn7/" url="http://archive.linuxmuster.net/dists/" policy="on_demand" | jq -r '.pulp_href')
REMOTE_HREF=$(http post "${BASE_ADDR}"/pulp/api/v3/remotes/deb/apt/ name="${ENTITIES_NAME}" distributions="debian-stable/binary/" url="https://pkg.jenkins.io/" policy="on_demand" | jq -r '.pulp_href')
```

PR #299, opened 2021-07-01
Add continuation hint to NoPackageIndexFile error message

quba42:
[noissue]

A user posted the error on the pulp matrix channel, so I decided to add the hint I gave right in the error message.

### DIFF
--- a/pulp_deb/app/tasks/synchronizing.py
+++ b/pulp_deb/app/tasks/synchronizing.py
@@ -90,9 +90,11 @@ class NoPackageIndexFile(Exception):
         """
         Exception to signal, that no file representing a package index is present.
         """
-        super().__init__(
-            "No suitable Package index file found in '{}'.".format(relative_dir), *args, **kwargs
+        message = (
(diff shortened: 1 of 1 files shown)


pulp_deb/app/tasks/synchronizing.py
```diff
@@ -90,9 +90,11 @@ def __init__(self, relative_dir, *args, **kwargs):
         """
         Exception to signal, that no file representing a package index is present.
         """
-        super().__init__(
-            "No suitable Package index file found in '{}'.".format(relative_dir), *args, **kwargs
+        message = (
+            "No suitable Package index file found in '{}'. "
+            "Consider setting ignore_missing_package_indices to 'True' for your remote."
```
mdellweg, line 95: How about "If you are synching from an incomplete mirror, consider..."

mdellweg (approved):
Maybe even more verbose, but definitely a good improvement.

pulp_deb/app/tasks/synchronizing.py
quba42: Suggestion incorporated!

PR #300, opened 2021-07-07
Add missing "Size" field in publications

m-bucher:
Closes #8506
https://pulp.plan.io/issues/8506

### DIFF
--- a/pulp_deb/app/tasks/synchronizing.py
+++ b/pulp_deb/app/tasks/synchronizing.py
@@ -622,7 +622,9 @@ class DebFirstStage(Stage):
                 )
                 package_path = os.path.join(self.parsed_url.path, package_relpath)
                 package_da = DeclarativeArtifact(
-                    artifact=Artifact(**_get_checksums(package_paragraph)),
+                    artifact=Artifact(
+                        size=int(package_paragraph["Size"]), **_get_checksums(package_paragraph)
+                    ),
                     url=urlunparse(self.parsed_url._replace(path=package_path)),
                     relative_path=package_relpath,
                     remote=self.remote,
--- a/pulp_deb/tests/functional/api/test_publish_complex_dists.py
+++ b/pulp_deb/tests/functional/api/test_publish_complex_dists.py
@@ -3,6 +3,8 @@
 import unittest
 import os
 
+from debian import deb822
+
 from pulp_smash import config
 from pulp_smash.pulp3.bindings import monitor_task
 from pulp_smash.pulp3.utils import (
@@ -13,6 +15,7 @@ from pulp_smash.pulp3.utils import (
     gen_distribution,
     download_content_unit,
 )
+from pulp_smash.utils import http_get
 
 from pulp_deb.tests.functional.constants import (
     DEB_COMPLEX_DISTS_FIXTURE_URL,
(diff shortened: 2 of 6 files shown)


pulp_deb/tests/functional/api/test_publish_complex_dists.py
```diff
@@ -170,4 +173,33 @@ def do_publish(self, expected_values):
         download_content_unit(cfg, distribution.to_dict(), release_file_path)
 
         for package_index_path in expected_values["package_index_paths"]:
-            download_content_unit(cfg, distribution.to_dict(), package_index_path)
+            published = download_content_unit(cfg, distribution.to_dict(), package_index_path)
+            remote = http_get("/".join([remote_body["url"], package_index_path]))
```
quba42, line 177: I wonder if it would not make more sense to just use the `DEB_COMPLEX_DISTS_FIXTURE_URL` directly here, instead of going through the remote we created?

pulp_deb/tests/functional/api/test_publish_complex_dists.py
```diff
@@ -170,4 +173,33 @@ def do_publish(self, expected_values):
...
+                    pkg[k], orig_pkg[k], "Field '{}' of package '{}' does not match".format(k, name)
+                )
+
+    def parse_package_index(self, pkg_idx):
+        """Parses PackageIndex file-string.
+        Returns a dict of the packages by '<Package>-<Version>'.
+        """
+        packages = {}
+        for package in deb822.Packages.iter_paragraphs(pkg_idx):
+            packages["-".join([package["Package"], package["Version"]])] = package
```
quba42, line 204: Not sure if Package name and Version alone are enough to guarantee a unique dict key...

pulp_deb/tests/functional/api/test_publish_complex_dists.py
```diff
@@ -170,4 +173,33 @@ def do_publish(self, expected_values):
...
+            published = download_content_unit(cfg, distribution.to_dict(), package_index_path)
+            remote = http_get("/".join([remote_body["url"], package_index_path]))
+            self.assert_equal_package_index(remote, published)
+
+    def assert_equal_package_index(self, orig, new):
+        """In-detail check of two PackageIndex file-strings"""
+        parsed_orig = self.parse_package_index(orig)
+        parsed_new = self.parse_package_index(new)
+
+        self.assertEqual(len(parsed_orig), len(parsed_new))
```
quba42, line 185: Should we also make sure that `len` is greater 0? Just to make sure the test does not pass trivially because `parse_package_index` returned something empty?

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -622,7 +622,9 @@ async def _handle_package_index(
                 )
                 package_path = os.path.join(self.parsed_url.path, package_relpath)
                 package_da = DeclarativeArtifact(
-                    artifact=Artifact(**_get_checksums(package_paragraph)),
+                    artifact=Artifact(
+                        size=int(package_paragraph["Size"]), **_get_checksums(package_paragraph)
+                    ),
```
mdellweg, line 627: This will add validation of the size of the file when downloading it, i think. Let's hope metadata does not lie.

mdellweg (commented):
The code change looks good to me. :+1:
Haven't looked much into the test changes, however.

pulp_deb/app/tasks/synchronizing.py
m-bucher: In a perfect world it would be a good feature to validate the size, because it makes the package-checksum more secure.
However, in our future :crystal_ball: I see an issue that complains about why repository xyz cannot have the correct package size in the package-index file and should be synced nevertheless :roll_eyes:

quba42:
The test fails with: `ModuleNotFoundError: No module named '/usr/local/lib/python3'`
What?!
And this happens in unittests which you did not change... Maybe we need the latest plugin template?!

pulp_deb/app/tasks/synchronizing.py
mdellweg: I am mostly concerned that some repositories will count the final linebreak in metadata files differently. Not sure if this comes from a distant memory, or is just overcautious.

quba42: I can't think of a better approach than merging and later adding exception handling if it becomes a problem...

pulp_deb/tests/functional/api/test_publish_complex_dists.py
m-bucher: Given that some of our fixture Package indizes are empty to begin with (e.g. https://fixtures.pulpproject.org/debian-complex-dists/dists/ragnarok-backports/asgard/binary-armeb/Packages).
I do not think we should do this.

Unless we want to store for the package-indizes, how much packages we expect.
But this would make the pulp_deb test-code even more dependant on the fixtures, which IMHO can be a problem, because it ties versions/commits of pulpcore-fixtures close together to versions/commits in pulp_deb.

pulp_deb/tests/functional/api/test_publish_complex_dists.py
m-bucher: Makes sense :sweat_smile:

pulp_deb/tests/functional/api/test_publish_complex_dists.py
m-bucher: Agreed, thanks

pulp_deb/tests/functional/api/test_publish_complex_dists.py
quba42: I wasn't aware of the empty index... I guess we just leave it as it is.

quba42 (approved):
I performed one final manual test with `policy="on_demand"` which was now successfull, I am merging.

PR #308, opened 2021-07-28
Add touch statements to prevent false positives during orphan cleanup

quba42:
Closes #9162
https://pulp.plan.io/issues/9162

### DIFF
--- a/CHANGES/9162.misc
+++ b/CHANGES/9162.misc
@@ -1,0 +1,1 @@
+Added touch statements to prevent false positives during orphan cleanup.
--- a/pulp_deb/app/serializers/content_serializers.py
+++ b/pulp_deb/app/serializers/content_serializers.py
@@ -72,6 +72,7 @@ class GenericContentSerializer(SingleArtifactContentUploadSerializer, ContentChe
             sha256=data["sha256"], relative_path=data["relative_path"]
         )
         if content.exists():
+            content.first().touch()  # Orphan cleanup protection so the user has a chance to use it!
             raise ValidationError(
                 _(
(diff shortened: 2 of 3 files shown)


quba42:
I am very unsure if the two identified touch statements are needed, and also unsure if I need extra touch statements...

mdellweg:
This patch should be accompanied with a version bump to pulpcore>=3.14.
Alternatively you can only call "touch" if `has_attr`.

quba42:
Some more references:

- https://github.com/pulp/community/discussions/59
- https://github.com/pulp/community/discussions/58

mdellweg (approved):
I didn't look for cases...

quba42:
@mdellweg, @daviddavis suggested [this](https://github.com/pulp/pulp_deb/blob/33c878e640bd232b8684bc84e0b0df49b405243b/pulp_deb/app/serializers/content_serializers.py#L74-L82) might be a candidate, but I don't see it. Looks to me like that artifact is only retrieved to then throw the error (so does not need cleanup protection?). Could you have one more look at that example if I am getting this right or wrong?

daviddavis:
Adding a touch call to `deferred_validate` is not essential but it would prevent the content from being cleaned up before the user has a chance to use it (ie add it to a repo version).

mdellweg:
It's hard to tell. Look to me like the user wanted to create such a thing if he received that answer, so maybe touch is a good idea.

quba42:
I added three more touch calls. I also added comments on what the intention behind each call is.
Unless there is more feedback, I will merge this version, and backport it to the 2.14 branch. (Gives me a chance to try out cherry-pick automation.)

PR #309, opened 2021-07-28
Update release process documentation

quba42:
[noissue]


quba42:
Note to self: I completely removed all references to the deprecation policy, this was not my intention...

PR #311, opened 2021-07-28
Raise the floor on pulpcore to 3.15

quba42:
[noissue]

This arguably should have been part of:
6b56c176da909cdb4b715b071b2bb3673ee76fa1

See also issues #9036, #9164


quba42:
I have learned this is not needed, because older pulpcore versions will work with the newer python versions in due course.

PR #312, opened 2021-07-28
Add python-apt as a library dependency

quba42:
[noissue]

Also update the library dependencies to the latest that are available.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,3 @@
 pulpcore>=3.15
-python-debian>=0.1.36
+python-debian>=0.1.40
+python-apt>=0.7.8


mdellweg (approved):

evgeni:
The version on pypi is from 2012 (!).

https://launchpad.net/python-apt/ (the homepage linked on pypi) lists a version 2.2.1 from 2021, and that is also what https://tracker.debian.org/pkg/python-apt lists.

Conan-Kudo:
We ship python-apt 2.2.0 in Fedora right now: https://src.fedoraproject.org/rpms/python-apt

2.2.1 seems to be some minor Debian packaging changes with no code changes, but I'll update to it sometime soon.

quba42:
I opened a discussion on how to proceed with this here: https://github.com/pulp/community/discussions/73

PR #317, opened 2021-07-30
Update links

fao89:
[noissue]


quba42 (approved):
How did I miss these? I remember having a ticket for updating them and everything. Thanks for the catch!

PR #318, opened 2021-07-30
Fix  db migration doc

jshen28:
Installation guide seems obsolete on db migration
instructions.

Signed-off-by: ushen <yshxxsjt715@gmail.com>

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -36,8 +36,8 @@ Make and Run Migrations
 
 .. code-block:: bash
 
-   pulp-manager makemigrations pulp_deb
-   pulp-manager migrate pulp_deb
+   pulpcore-manager makemigrations deb
+   pulpcore-manager migrate deb
 
 
 Run Services


docs/installation.rst
```diff
@@ -36,8 +36,8 @@ Make and Run Migrations
 
 .. code-block:: bash
 
-   pulp-manager makemigrations pulp_deb
-   pulp-manager migrate pulp_deb
+   pulpcore-manager makemigrations deb
```
quba42, line 39: This is a command that should only really be used by developers who know what they are doing in a development environment. These "installation" docs are fatally flawed and confusing to average users. I feel like the best thing to do would be to replace the entire guide with a reference to the pulpcore installation docs, a long with a short explanation why... I have opened a new ticket for this: https://pulp.plan.io/issues/9186

You certainly have my thanks for bringing this issue back to my attention!

quba42 (commented):
Like I said, I think this entire section should really be dropped from the pulp_deb docs.
We could merge your changes anyway, since they are an obvious improvement on a command that is simply incorrect... There is another instance of `pulp-manger` in line 48, but that whole section is also pretty flawed...

docs/installation.rst
jshen28: hmm.. it might be good to have a section in pulpcore to tell users how to install plugins... but put it in plugin pages looks fine to me as well, since putting everything (there are lot of plugins..) together might be too overwhelming for a single page..

quba42: If you are using the pulp Ansible installer to install pulp, then adding additional plugins, is just a matter of adding them to a list variable and rerunning the installer. In generall the installation for most plugins should be very similar, so long as you substitute the correct plugins name.

jshen28: that's true..

quba42:
I am going to merge this as a clear improvement on the status quo. Thanks for the contribution!

jshen28:
thank you very much for merging this.. and looking forward to making more  :=)

quba42:
@jshen28 You did inspire me to do a follow on rework of the installation docs: https://github.com/pulp/pulp_deb/pull/322

PR #330, opened 2021-08-03
Disable the failing plugin-from-pypi test for nightly CI

quba42:
[noissue]

This test is failing because the signing service creation hack is not
applied by this job. The test should be re-enabled once signing services
are created as part of the regular test setup.

See also: https://pulp.plan.io/issues/8640


quba42:
This alone will not be enough to fix the nightly CI workflow, see also: https://github.com/pulp/plugin_template/issues/464

PR #331, opened 2021-08-04
Update CI files

pulpbot:
[noissue]


mdellweg (approved):

quba42:
@mdellweg Do you know what happens if I don't merge this by the next time the "CI Update" workflow runs?

mdellweg:
It should keep updating this PR.
It's worth a try, eh?

mdellweg:
This one doesn't seem to change anything. So maybe it is just about the right one for an experiment.

mdellweg:
With the new change, it may be necessary to adjust the "required checks" after merging this.

quba42:
> With the new change, it may be necessary to adjust the "required checks" after merging this.

What do you mean by "required checks"?

mdellweg:
I mean the fact that they are marked as required for a PR to be merged, but nevermind, it looks like github does that by filename and not workflow name.
image

PR #333, opened 2021-08-19
Fix import of delete_orphans in tests

mdellweg:


quba42 (approved):
I presume this change should not be back-ported to old branches?

PR #348, opened 2021-09-02
Ignore package fields with corrupt values

Manisha15:
Fixes #9333
https://pulp.plan.io/issues/9333

### DIFF
--- a/pulp_deb/app/serializers/content_serializers.py
+++ b/pulp_deb/app/serializers/content_serializers.py
@@ -29,6 +29,12 @@ from pulp_deb.app.models import (
     ReleaseFile,
 )
 
+from pulp_deb.app.models.content import BOOL_CHOICES
+
+import logging
+
+log = logging.getLogger(__name__)
+
 
 class YesNoField(Field):
     """
@@ -272,6 +278,45 @@ class BasePackage822Serializer(SingleArtifactContentSerializer):
             elif k not in skip:
                 # also save the fields not in TRANSLATION_DICT
                 args["custom_fields"][k] = v
+
+        # Delete package fields with values of incorrect type
+        if "installed_size" in args:
+            try:
+                int(args["installed_size"])
+            except (TypeError, ValueError):
+                log.warn(
+                    _(
+                        "Dropping 'Installed-Size' field from '{}', "
+                        "since the value '{}' is of incorrect type."
+                    ).format(args["package"], args["installed_size"])
+                )
+                del args["installed_size"]
+        message = (
+            "Dropping '{}' field from package '{}', "
+            "since the value '{}' is not in the allowed values list '{}'"
+        )
+        bool_values = [value[1] for value in BOOL_CHOICES]
+        if "essential" in args and args["essential"] not in bool_values:
+            log.warn(
+                _(message).format("Essential", args["package"], args["essential"], bool_values)
+            )
+            del args["essential"]
+        if "build_essential" in args and args["build_essential"] not in bool_values:
+            log.warn(
+                _(message).format(
+                    "Build-Essential", args["package"], args["build_essential"], bool_values
+                )
+            )
+            del args["build_essential"]
+        if "multi_arch" in args:
+            allowed_values = [value[1] for value in BasePackage.MULTIARCH_CHOICES]
+            if args["multi_arch"] not in allowed_values:
+                log.warn(
+                    _(message).format(
+                        "Multi-Arch", args["package"], args["multi_arch"], allowed_values
+                    )
+                )
+                del args["multi_arch"]
 
         return cls(data=args, **kwargs)
 
--- a/CHANGES/9333.misc
+++ b/CHANGES/9333.misc
@@ -1,0 +1,1 @@
+Added fix to delete package fields with values of an incorrect type.


Manisha15:
Fixes #9333
https://pulp.plan.io/issues/9333

pulp_deb/app/serializers/content_serializers.py
```diff
@@ -265,9 +265,24 @@ def from822(cls, data, **kwargs):
...
+        MULTIARCH_CHOICES = [
+        ("no", "no"),
+        ("same", "same"),
+        ("foreign", "foreign"),
+        ("allowed", "allowed"),
+    ]
         for k, v in data.items():
             if k in cls.TRANSLATION_DICT_INV:
                 key = cls.TRANSLATION_DICT_INV[k]
+                if key == 'installed_size' and not isinstance(args[key], int):
```
quba42, line 277: I wonder if it would be more efficient to build the entire `args` dict without considering the special fields first, and then to remove any broken fields after. So have something like the following just before the return:
```
if 'installed_size' in args and not isinstance(args['installed_size'], int):
    del args['installed_size']
```
and so on.

quba42 (changes requested):
Some ideas.

pulp_deb/app/serializers/content_serializers.py
```diff
@@ -273,6 +273,13 @@ def from822(cls, data, **kwargs):
                 # also save the fields not in TRANSLATION_DICT
                 args["custom_fields"][k] = v
 
+        # Delete package fields with corrupt values
```
quba42, line 276:
```suggestion
        # Delete package fields with values of an incorrect type
```

pulp_deb/app/serializers/content_serializers.py
```diff
@@ -273,6 +273,13 @@ def from822(cls, data, **kwargs):
...
+        del args['installed_size'] if 'installed_size' in args and not isinstance(args['installed_size'], int)
```
quba42, line 277: I need to test to be sure, but I think the current logic won't work because at this point all the fields contain strings. If I am right then what we need to test is whether we have a string that can be successfully casted to the type we want. So the strings `"55"` or even `"   55 "` should count as an int, but `"NaN"` or `" 5 5 "` should not. I am guessing anything that can be successfully converted with `int(string_value)` is fine (though I am not sure that is what is being used to cast things to the right type for the DB field in question).

pulp_deb/app/serializers/content_serializers.py
```diff
@@ -273,6 +273,13 @@ def from822(cls, data, **kwargs):
...
+        del args['essential'] if 'essential' in args and not isinstance(args['essential'], bool)
+        del args['build_essential'] if 'build_essential' in args and not isinstance(args['build_essential'], bool)
+        if 'multi_arch' in args and not isinstance(args['multi_arch'], str):
```
quba42, line 280: If I am right about all fields containing strings at this point then we don't need this test:
```suggestion
        if 'multi_arch' in args:
```

quba42 (changes requested):
I also just remembered we wanted to add a log warning whenever we drop a field!

pulp_deb/app/serializers/content_serializers.py
```diff
@@ -273,6 +277,38 @@ def from822(cls, data, **kwargs):
                 # also save the fields not in TRANSLATION_DICT
                 args["custom_fields"][k] = v
 
+        # Delete package fields with values of incorrect type
+        if "installed_size" in args and not args["installed_size"].strip().isdigit():
```
quba42, line 281: As far as I can tell this works well. Note that this validation is different from what the Django `IntegerField` type uses. Django simply tries to cast the incoming value to integer using `int()` ([see here](https://github.com/django/django/blob/ca9872905559026af82000e46cde6f7dedc897b6/django/db/models/fields/__init__.py#L1823)).

We could consider aligning our test with Django, by using something like:
```
if "installed_size" in args:
        try:
            int(args["installed_size"])
        except (TypeError, ValueError):
            log.warn(...)
            del args["installed_size"]
```
Not sure what is better.

pulp_deb/app/serializers/content_serializers.py
```diff
@@ -273,6 +277,38 @@ def from822(cls, data, **kwargs):
...
+            log.warn(
+                ("Removing 'Installed-Size' field from {} as it contains incorrect type.").format(
+                    args["package"]
+                )
+            )
+            del args["installed_size"]
+        if "essential" in args and not isinstance(args["essential"], bool):
```
quba42, line 288: This currently does not work, since we need to do something similar as for the "multi_arch" field:
```suggestion
        if "essential" in args and args["essential"] not in [ value[1] for value in BOOL_CHOICES ]:
```
Note that `BOOL_CHOICES` will need to be imported since it is not part of the `BasePackage` class (like `MULTIARCH_CHOICES` is).

pulp_deb/app/serializers/content_serializers.py
```diff
@@ -273,6 +277,38 @@ def from822(cls, data, **kwargs):
...
+            log.warn(
+                ("Removing 'Essential' field from {} as it contains incorrect type.").format(
+                    args["package"]
+                )
+            )
+            del args["essential"]
+        if "build_essential" in args and not isinstance(args["build_essential"], bool):
```
quba42, line 295: Same problem as with the `essential` field.

pulp_deb/app/serializers/content_serializers.py
```diff
@@ -273,6 +277,38 @@ def from822(cls, data, **kwargs):
...
+            log.warn(
+                ("Removing 'Build Essential' field from {} as it contains incorrect type.").format(
+                    args["package"]
+                )
+            )
+            del args["build_essential"]
+        if "multi_arch" in args and args["multi_arch"] not in [
+            value[0] for value in BasePackage.MULTIARCH_CHOICES
```
quba42, line 303:
```suggestion
            value[1] for value in BasePackage.MULTIARCH_CHOICES
```
Comparing with the `BOOL_CHOICES` this is probably more correct even though it does not make a difference for `MULTIARCH_CHOICES`...

pulp_deb/app/serializers/content_serializers.py
```diff
@@ -273,6 +277,38 @@ def from822(cls, data, **kwargs):
...
+        ]:
+            log.warn(
+                ("Removing 'Multi Arch' field from {} as it contains incorrect type.").format(
+                    args["package"]
```
quba42, line 307: I am wondering if we can be more specific for our warnings, something like:
```
allowed_values = [ value[1] for value in BasePackage.MULTIARCH_CHOICES ]
message="Dropping 'Multi-Arch' field from package {}, since the value '{}' is not in the allowed values list '{}'"
log.warn(_(message).format(args["package"], args["multi_arch"], allowed_values))
```
Printing the value that caused the problem really helps with debugging!

quba42 (changes requested):
I think we are getting there. :stuck_out_tongue_winking_eye:

pulp_deb/app/serializers/content_serializers.py
Manisha15: Does the args["essential"] already contains the BOOL_CHOICES value or is it still having value from package? If it is from former then it already is boolean but if it from latter then we need to import BOOL_CHOICES.

quba42: `args["essential"]` contains what was read from the `Packages` file using the `python-debian.deb822` library.
In my test it was `args["essential"] = 'yes'` which is a string and not a bool, so the field was dropped despite containing the correct value "yes".

pulp_deb/app/serializers/content_serializers.py
```diff
@@ -273,6 +277,43 @@ def from822(cls, data, **kwargs):
...
 
+        # Delete package fields with values of incorrect type
+        if "installed_size" in args:
+            try:
+                int(args["installed_size"])
+            except (TypeError, ValueError):
+                log.warn(
+                    _(
+                        "Dropping 'Installed-Size' field from '{}', \
+                         since the value '{}' is of incorrect type."
```
quba42, line 288:
```suggestion
                        (
                            "Dropping 'Installed-Size' field from '{}', "
                            "since the value '{}' cannot be casted to type 'int'."
                        )
```
This is needed to avoid having indentation make it into the final string, your version resulted in the following warning:
```
pulp_deb.app.serializers.content_serializers:WARNING: Dropping 'Installed-Size' field from code-insiders,                          since the value NaN is of incorrect type.
```

pulp_deb/app/serializers/content_serializers.py
```diff
@@ -273,6 +277,43 @@ def from822(cls, data, **kwargs):
...
+                    ).format(args["package"], args["installed_size"])
+                )
+                del args["installed_size"]
+        message = "Dropping '{}' field from package '{}', \
+                       since the value '{}' is not in the allowed values list '{}'"
```
quba42, line 293:
```suggestion
        message = (
            "Dropping '{}' field from package '{}', "
            "since the value '{}' is not in the allowed values list '{}'"
        )
```

quba42 (changes requested):
Only very minor suggestions now, once these are in I am happy to merge!

quba42 (approved):

PR #350, opened 2021-09-07
Fix pulpcore deprecations

quba42:
Closes #9349
https://pulp.plan.io/issues/9349


pulp_deb/app/viewsets/publication.py
```diff
@@ -44,8 +44,8 @@ def create(self, request):
         repository_version = serializer.validated_data.get("repository_version")
 
         result = dispatch(
-            tasks.publish_verbatim,
-            [repository_version.repository],
+            func=tasks.publish_verbatim,
+            exclusive_resources=[repository_version.repository],
```
quba42, line 48: I don't actually have any knowledge of when to prefer `exclusive_resources`, and when to prefer `shared_resources`. The former just sounded safer, so I defaulted to it. Feedback is appreciated!

quba42 (commented):
I looked through all deprecations in the pulpcore changelog, and I am reasonably confident that this is all there is.

pulp_deb/app/viewsets/publication.py
quba42: Thanks for the feedback/explanation @ipanova!

quba42 (commented):
If the tests turn green again, I will merge this.

PR #352, opened 2021-09-13
Opt out of the single commit CI check

quba42:
[noissue]

Opting out of single commit, resulting in only trivial changes to the CI
files themselves!


quba42:
I am merging this in spite of the failed tests, because those are clearly unrelated to this change!

PR #353, opened 2021-09-19
Update CI files

pulpbot:
[noissue]


mdellweg (approved):

template_config.yml
```diff
@@ -57,6 +57,8 @@ release_email: pulp-infra@redhat.com
 release_user: pulpbot
 single_commit_check: false
 sync_ci: true
+tasking_allow_async_unsafe: true
+test_azure: false
```
quba42, line 61: @mdellweg Should we have a go at setting this to `true`?

mdellweg: Sure, why not. Anyway, this needs to be updated by the workflow once to resolve the conflict.

quba42: I will leave it for now, and then enable on a separate branch.

PR #354, opened 2021-09-22
Activate check that tasking is async safe wrt ORM

mdellweg:
[noissue]


mdellweg:
@quba42 Looks like you fixed it everywhere. :tada:

quba42 (approved):
Yeah!

PR #357, opened 2021-09-30
Activate azure tests

mdellweg:
[noissue]

### DIFF
--- a/.github/workflows/scripts/post_before_script.sh
+++ b/.github/workflows/scripts/post_before_script.sh
@@ -1,10 +1,10 @@
 set -euv
 
-if [ "$TEST" = "pulp" ] || [ "$TEST" = "performance" ] || [ "$TEST" = "s3" ]
+if [ "$TEST" = "pulp" ] || [ "$TEST" = "performance" ] || [ "$TEST" = "azure" ] || [ "$TEST" = "s3" ]
 then
   # Add signing service script and setup script:
-  cat pulp_deb/tests/functional/sign_deb_release.sh | cmd_stdin_prefix bash -c "cat > /root/sign_deb_release.sh"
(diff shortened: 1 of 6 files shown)


.github/workflows/scripts/post_before_script.sh
```diff
@@ -1,10 +1,10 @@
 set -euv
 
-if [ "$TEST" = "pulp" ] || [ "$TEST" = "performance" ] || [ "$TEST" = "s3" ]
+if [ "$TEST" = "pulp" ] || [ "$TEST" = "performance" ] || [ "$TEST" = "azure" ] || [ "$TEST" = "s3" ]
 then
   # Add signing service script and setup script:
-  cat pulp_deb/tests/functional/sign_deb_release.sh | cmd_stdin_prefix bash -c "cat > /root/sign_deb_release.sh"
-  cat pulp_deb/tests/functional/setup_signing_service.py | cmd_stdin_prefix bash -c "cat > /tmp/setup_signing_service.py"
+  cmd_stdin_prefix bash -c "cat > /root/sign_deb_release.sh" < pulp_deb/tests/functional/sign_deb_release.sh
+  cmd_stdin_prefix bash -c "cat > /tmp/setup_signing_service.py" < pulp_deb/tests/functional/setup_signing_service.py
```
quba42, line 7: For future reference, why is `<` preferable to `cat |`?

quba42 (approved):

.github/workflows/scripts/post_before_script.sh
mdellweg: It's one process less to start up from the shell.

quba42: Shellcheck says: "useless use of cat"

mdellweg: The one thing i don't like about it is that the beginning of the pipe is moved to the end of the line.

PR #376, opened 2021-12-03
Update pulpcore branch + update CI

fao89:
[noissue]

PR #377, opened 2021-12-05
Update CI files

pulpbot:
[noissue]


quba42:
See https://github.com/pulp/pulp_deb/pull/376 instead!

PR #379, opened 2021-12-16
Update CI files

pulpbot:
[noissue]


mdellweg (approved):

.github/workflows/scripts/publish_docs.sh
```diff
@@ -19,7 +19,7 @@ chmod 600 ~/.ssh/pulp-infra
 echo "docs.pulpproject.org,8.43.85.236 ecdsa-sha2-nistp256 AAAAE2VjZHNhLXNoYTItbmlzdHAyNTYAAAAIbmlzdHAyNTYAAABBBGXG+8vjSQvnAkq33i0XWgpSrbco3rRqNZr0SfVeiqFI7RN/VznwXMioDDhc+hQtgVhd6TYBOrV07IMcKj+FAzg=" >> /home/runner/.ssh/known_hosts
 chmod 644 /home/runner/.ssh/known_hosts
 
-pip3 install -r doc_requirements.txt
+pip3 install packaging
```
quba42, line 22: How does "packaging" replace `-r doc_requirements.txt`? And should we remove `doc_requirements.txt` then?

quba42: Info: The plugin template says: "No deprecated files found."

mdellweg: As far as i see, this script is not building the docs, but only publishing them. So less requirements do not sound completely surprising.
Also i think `doc_requirements` are still used to build the docs elsewhere.

quba42: Ah OK, that makes sense then.

quba42 (approved):

PR #381, opened 2021-12-27
Add advanced copy mechanism in Pulp3 DEB without dependency_solving

Manisha15:
Adds Copy mechanism to be used by Katello to copy content in CV filters

### DIFF
--- a/CHANGES/396.feature
+++ b/CHANGES/396.feature
@@ -1,0 +1,1 @@
+Added experimental advanced copy API with support for structured copying.
--- a/pulp_deb/app/serializers/__init__.py
+++ b/pulp_deb/app/serializers/__init__.py
@@ -23,4 +23,4 @@ from .publication_serializers import (
 
 from .remote_serializers import AptRemoteSerializer
 
-from .repository_serializers import AptRepositorySerializer
+from .repository_serializers import AptRepositorySerializer, CopySerializer
--- a/pulp_deb/app/serializers/repository_serializers.py
+++ b/pulp_deb/app/serializers/repository_serializers.py
@@ -1,6 +1,11 @@
-from pulpcore.plugin.serializers import RepositorySerializer
+from gettext import gettext as _
+from pulpcore.plugin.serializers import RepositorySerializer, validate_unknown_fields
 
 from pulp_deb.app.models import AptRepository
+
+from jsonschema import Draft7Validator
+from rest_framework import serializers
+from pulp_deb.app.schema import COPY_CONFIG_SCHEMA
 
 
 class AptRepositorySerializer(RepositorySerializer):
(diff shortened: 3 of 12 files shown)


quba42:
This change needs to add docs here (and possibly elsewhere): https://github.com/pulp/pulp_deb/blob/main/docs/feature_overview.rst

quba42:
I guess this closes the following issue: https://pulp.plan.io/issues/7661

pulp_deb/app/tasks/copy.py
```diff
@@ -0,0 +1,113 @@
...
+        for entry in config:
+            (
+                source_repo_version,
+                dest_repo_version,
+                dest_repo,
+                content_filter,
+                dest_version_provided,
+            ) = process_entry(entry)
+
+            content_to_copy = source_repo_version.content.filter(content_filter)
```
quba42, line 102: At this point, `content_filter` is just based on the list of `pulp_href` supplied by the user. This list can include href for packages, but also href for metadata types. I have done some testing (by including silly content in the copy on purpose), and so far have not found any errors resulting from this. It does however result in nonsensical content ending up in the target repo, which could cause god knows what problems down the line. I suggest we should be a bit more opinionated here, and provide some sanity checking and guardrails for the user.

I suggest we try doing this in the `_process_config` function, by simply dropping any content of the wrong type from the users list with a warning if `structured` is set to true.

If this comment is too unclear, let's have a meeting about this.

quba42 (changes requested):
Most of the requested changes are small nitpicks in docstring wording or naming.
I would make one substantive addition to the `_process_config` function.

Once that is done, this should be very much nearing completion.

pulp_deb/app/serializers/__init__.py
```diff
@@ -23,4 +23,4 @@
 
 from .remote_serializers import AptRemoteSerializer
 
-from .repository_serializers import AptRepositorySerializer
+from .repository_serializers import AptRepositorySerializer, CopySerializer
```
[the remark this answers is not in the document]
quba42: We decided to leave it as is.

CHANGES/396.feature
```diff
@@ -0,0 +1 @@
+AAdd experimental advanced copy workflow for structured copying.
```
quba42, line 1:
```suggestion
Added experimental advanced copy API with support for structured copying.
```
Noticed a typo, remembered the convention was for passed tense, and made a minor adjustment. :smile:

pulp_deb/app/serializers/repository_serializers.py
```diff
@@ -11,3 +16,50 @@ class AptRepositorySerializer(RepositorySerializer):
...
+    """
+
+    config = serializers.JSONField(
+        help_text=_("A JSON document describing sources, destinations, and content to be copied")
+    )
+
+    structured = serializers.BooleanField(
+        help_text=_(
+            "Also copy any distributions, components, and releases as needed for any packages "
+            "being copied. This will allow for structured publications of the target repository."
```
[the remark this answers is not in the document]
Manisha15: Fixed.

pulp_deb/app/tasks/copy.py
```diff
@@ -0,0 +1,119 @@
...
+
+from pulp_deb.app.models import AptRepository, Package, ReleaseArchitecture, PackageReleaseComponent
+
+import logging
+from gettext import gettext as _
+
+log = logging.getLogger(__name__)
+
+
+def find_structured_publish_attributes(content, src_repo_version):
```
quba42, line 14:
```suggestion
def find_structured_publish_content(content, src_repo_version):
```
Final naming nitpick.

pulp_deb/app/tasks/copy.py
```diff
@@ -0,0 +1,119 @@
...
+        pulp_type=PackageReleaseComponent.get_pulp_type()
+    ).only("pk")
+    architecture_ids = src_repo_version.content.filter(
+        pulp_type=ReleaseArchitecture.get_pulp_type()
+    ).only("pk")
+    package_release_components = PackageReleaseComponent.objects.filter(
+        pk__in=package_release_component_ids
+    )
+
+    children = set()
```
quba42, line 35:
```suggestion
    structured_publish_content = set()
```
If we can update all instances of `children` and `find_structured_publish_attributes` we are done!

quba42 (changes requested):
Nearly done!

pulp_deb/app/tasks/copy.py
quba42: Seems to work as intended.

quba42:
I added some minimal "feature overview" documentation.
A full workflow documentation will not be added until there are pulp-cli-deb commands for this feature, and we find the time to write that documentation.

To that end: https://github.com/pulp/pulp-cli-deb/issues/22

PR #390, opened 2022-01-04
Use Github issues instead of redmine

fao89:
[noissue]


quba42 (approved):
LGTM, and I don't think there is any point delaying trying it out. (The pulp_deb plan.io project is already read-only!)

PR #440, opened 2022-01-31
Fix wrong architecture in index handling

quba42:
[noissue]

This fixes the fix introduced by commit 041638348533e8bfe6be08a8d4bcf6d8dab6eb13
Fortunately this has been caught before it was released!


quba42:
The fact that this was not caught for the original PR is a symptom of insufficient test coverage!

Issue #449, opened 2022-03-07
Make various release file fields user configurable


quba42:
**Is your feature request related to a problem? Please describe.**

This feature is intended to provide a long term solution for https://github.com/pulp/pulp_deb/issues/443 while also empowering users to use pulp_deb repos in combination with "APT pinning" and "unattended upgrade" workflows.

**Describe the solution you'd like**

- The upstream Origin, Label, and Version fields should be stored on the release content during sync (if present).
- Users should be able to define their own Origin, Label, and Version values on repositories and/or publications.
- User supplied values trump upstream repo values during publish

If possible, consider the possibility of respecting arbitrary upstream and user supplied release file fields?

**Additional context**

#443 Should provide a backportable bugfix solution for the same problem, while this ticket provides a new design approach as a long term solution.

quba42:
The title of this ticket remains as valid as ever, but the proposed implementation requires significant rethink!

quba42:
`Description` is another Release file field that should be considered as part of this issue.

mikalstill:
@quba42, I'd be happy to take a swing at this one (I am already looking for a starter project to contribute to this plugin), but it is not clear to me from your comment why the design needs a rethink. Is there further information available somewhere else for that?

quba42:
@mikalstill I am afraid, I don't always keep the pulp_deb GitHub issues fully up to date (it's on my todo list), because we also use an internal issue tracker in addition to this one. "The design issues" I allude to in this thread are not reflected in an issue right now (I will open one today and link it from this thread). The short version as relates to this issue, is that we no longer want to store any additional fields on the `Release` model. It would probably be safe/fine, to just store them on the `ReleaseFile` model instead though, so that would not be a big difference for how to implement this issue.

quba42:
Note: I have now added an issue for the above mentioned "design issues": https://github.com/pulp/pulp_deb/issues/599

eman-koch:
I upvote this. In our scenario we would like to have all mirrored repositories in pulp made available to the system and then differ what to install with pin-priorities through apt preferences. This makes it easier to manage the systems for us but results in the possibility that we sometimes have 2 or 3 repositories with different versions active, let's take the example with mariadb:

![image](https://user-images.githubusercontent.com/96131915/229139159-c4c2fb29-b2f3-4d76-a37f-a769ee533278.png)

[...]

PR #450, opened 2022-03-07
Do not set the release file Label or Version during publish

### LINKED ISSUE #443: Release file version changes require manual confirmation in APT
**Version**
Affects all versions of pulp_deb, when used with APT version 1.5 and greater.

**Describe the bug**
pulp_deb publishes the Pulp repo version as the Release file version.
Starting with APT 1.5 and greater, APT asks for user confirmation in order to keep using a repo with a changed Release file version, unless `apt-get update --allow-releaseinfo-change` is used.

This can break automation not using this flag, whenever the Release file version changes.

**Expected behavior**
Do not use the Release file version field in a way that clashes with basic APT design.
We should check if there are other release file fields that require manual confirmation when they change as well!

**Additional context**
See also: https://community.theforeman.org/t/pulp-deb-incorrect-version-in-release-and-inrelease/27416/3

quba42:
Closes #443

This is a quick and dirty workaround to prevent apt-secure from
constantly requiering users to run the following:

apt-get update --allow-releaseinfo-change

This fix changes the pulp_deb publication behaviour. To mitigate, the
old behaviour can be renabled by setting SET_RELEASE_FILE_LABEL and
SET_RELEASE_FILE_VERSION to True.

For a long term fix, additional new features and behaviour changes are
now planned: https://github.com/pulp/pulp_deb/issues/449


CHANGES/443.bugfix
```diff
@@ -0,0 +1,2 @@
+Changed the release file publication behaviour of the APT publisher to prevent a design clash with apt-secure.
+You may set PUBLISH_RELEASE_FILE_LABEL and PUBLISH_RELEASE_FILE_VERSION to True to revert to the old behaviour.
```
quba42, line 2: Should I be placing this in a single line? Not sure how towncrier likes it...

quba42:
Successfully tested using the config file values manually.

CHANGES/443.bugfix
mdellweg: No, you should follow the same coding style as you want to see it land in the Changelog. towncrier will insert it literally (maybe indented...).

quba42: Thanks for the tip!

Issue #422, opened 2022-01-05
Handle upstream package indices containing packages with the wrong architecture


quba42:
Minimal reproducer:
```
pip install pulp-cli-deb

# Constants:
ENTITIES_NAME='test'
REMOTE_OPTIONS='--url=https://apt.releases.hashicorp.com --distribution=bionic --component=main --architecture=amd64 --policy=on_demand'

# Steps using CLI:
pulp deb remote create --name=${ENTITIES_NAME} ${REMOTE_OPTIONS}
pulp deb repository create --name=${ENTITIES_NAME} --remote=${ENTITIES_NAME}
pulp deb repository sync --name=${ENTITIES_NAME}
APT_PUBLICATION_HREF=$(pulp deb publication create --repository=${ENTITIES_NAME} --simple=True --structured=True | jq -r '.pulp_href')
# Creating the publication fails!
pulp deb distribution create --name=${ENTITIES_NAME} --base-path=${ENTITIES_NAME} --publication=${APT_PUBLICATION_HREF}
```

Error message:
```
    "error"=>
     {"traceback"=>
       "  File \"/usr/lib/python3.6/site-packages/pulpcore/tasking/pulpcore_worker.py\", line 317, in _perform_task\n" +
       "    result = func(*args, **kwargs)\n" +
       "  File \"/usr/lib/python3.6/site-packages/pulp_deb/app/tasks/publishing.py\", line 168, in publish\n" +
       "    prc.package\n" +
       "  File \"/usr/lib/python3.6/site-packages/pulp_deb/app/tasks/publishing.py\", line 207, in add_package\n" +
       "    self.package_index_files[package.architecture][0]\n",
      "description"=>"'arm64'"},
```

quba42:
> Do you know what HCP uses for managing that repo? Might be worth escalating there?

I do not know where I might report the upstream issue, though it would be good to do so. The number of errors in that repo makes me think it might be something they built themselves... Perhaps just email them?

quba42:
Note: I alerted HashiCorp to our findings via the from found at: https://support.hashicorp.com/hc/en-us/requests/new

quba42:
Apparently the form I used is for paying enterprise customers only, so I also opened a thread in the HashiCorp community forum: https://discuss.hashicorp.com/t/buggy-weirdly-structured-hashicorp-apt-repo/34814

PR #451, opened 2022-03-09
Rework No-Support-for-Architecture-all repo format sync handling

### LINKED ISSUE #456: Rework No-Support-for-Architecture-all format sync handling
Many upstream repos declare the `No-Support-for-Architecture-all: Packages` option in their release file. For example, most official Debian bullseye repos.

This format results in a number of edge cases requiring special handling.
For example, we should not warn about `Architecture: all` packages in architecture specific indices for such repos.

Even more problematically, deb-mirror may mirror the metadata indicating `No-Support-for-Architecture-all: Packages` format, while failing to mirror the `binary-all` package indices (which actually results in "old style repo format" while declaring "mixed style repo format"). This two requires special handling, or the `Architecture: all` packages will either not be synced or cause outright sync failures for such mirrors! (which depends on the value of `ignore_missing_package_indices`).

quba42:
Closes #456

When a repo uses "No-Support-for-Architecture-all: Packages" style
format, we should not warn about all architecture packages in
architecture specific package indices! They are expected to be there!
We still skip those packages since there is no need to download them
twice!

When a mirror mirrors metadata indicating
"No-Support-for-Architecture-all: Packages" format, while failing to
mirror the binary-all package indices we introduce special handling to
ensure we nevertheless sync "Architecture: all" packages.

### DIFF
--- a/pulp_deb/app/tasks/synchronizing.py
+++ b/pulp_deb/app/tasks/synchronizing.py
@@ -92,8 +92,9 @@ class NoPackageIndexFile(Exception):
         """
         Exception to signal, that no file representing a package index is present.
         """
+        self.relative_dir = relative_dir
         message = (
-            "No suitable Package index file found in '{}'. If you are syncing from a partial "
+            "No suitable package index files found in '{}'. If you are syncing from a partial "
             "mirror, you can ignore this error for individual remotes "
             "(ignore_missing_package_indices='True') or system wide "
             "(FORCE_IGNORE_MISSING_PACKAGE_INDICES setting)."
@@ -114,6 +115,25 @@ class MissingReleaseFileField(Exception):
         """
         message = "The release file for distribution '{}' is missing the required field '{}'."
         super().__init__(_(message).format(distribution, field), *args, **kwargs)
+
+
+class UnknownNoSupportForArchitectureAllValue(Exception):
+    """
+    Exception Signifying that the Release file contains the 'No-Support-for-Architecture-all' field,
+    but with a value other than 'Packages'. We interpret this as an error since this would likely
+    signify some unknown repo format, that pulp_deb is more likely to get wrong than right!
+    """
+
+    def __init__(self, release_file_path, unknown_value, *args, **kwargs):
+        message = (
+            "The Release file at '{}' contains the 'No-Support-for-Architecture-all' field, with "
+            "unknown value '{}'! pulp_deb currently only understands the value 'Packages' for "
+            "this field, please open an issue at https://github.com/pulp/pulp_deb/issues "
+            "specifying the remote you are attempting to sync, so that we can improve pulp_deb!"
+        )
+        super().__init__(_(message).format(unknown_value), *args, **kwargs)
+
+    pass
 
 
 def synchronize(remote_pk, repository_pk, mirror):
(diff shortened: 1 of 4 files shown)


quba42:
This is a bugfix for a bug introduced by the as of yet unreleased #422 !

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -577,19 +585,61 @@ async def _handle_distribution(self, distribution):
...
+                        no_support_for_architecture_all=no_support_for_architecture_all,
+                    )
+                )
+            except NoPackageIndexFile:
+                message = (
+                    "The Release file at '{}' advertised 'No-Support-for-Architecture-all: "
+                    "Packages', however the binary-all index at '{}' appears to be missing! "
+                    "Defaulting back to mixed package index handling."
+                )
+                bla = "RETRIEVE relative_dir HERE"
```
quba42, line 623: Wups, forgot about this! This still needs fixing!

quba42: Fixed.

m-bucher (approved):

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -577,19 +586,60 @@ async def _handle_distribution(self, distribution):
...
         )
         release_component = await self._create_unit(release_component_dc)
+
+        # Try handling the architecture='all' index first (if present). That way, we can recover the
+        # special case, where a partial mirror does not mirror this index inspite of indicating in
+        # the mirrored metadata that it should.
+        if "all" in architectures and "Packages" in no_support_for_architecture_all:
+            architectures.remove("all")
+            try:
+                await asyncio.gather(
```
mdellweg, line 609: No need for gather. Just await the single coroutine.

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -577,19 +586,60 @@ async def _handle_distribution(self, distribution):
...
+                    self._handle_package_index(
+                        release_file=release_file,
+                        release_component=release_component,
+                        architecture="all",
+                        file_references=file_references,
+                        no_support_for_architecture_all=no_support_for_architecture_all,
```
mdellweg, line 615: Wondering if we couldn't pass a boolean here.

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -706,11 +757,12 @@ async def _handle_package_index(
             if (
                 settings.FORCE_IGNORE_MISSING_PACKAGE_INDICES
                 or self.remote.ignore_missing_package_indices
-            ):
-                log.info(_("No packages index for architecture {}. Skipping.").format(architecture))
-                return
+            ) and architecture != "all":
```
mdellweg, line 760: I would expect this extra check to also depend on "no_support_for_architecture_all"...

quba42: Well adding the additional condition did fix a failing test, which suggests you are right...

pulp_deb/app/tasks/synchronizing.py
quba42: Fixed.

pulp_deb/app/tasks/synchronizing.py
m-bucher: ...or yet to come :see_no_evil: 

Given that the value itself is not a boolean or boolean-like value I am with @quba42 here :smile:

quba42: I basically went with fail hard and fail early. If somebody tries to sync a Release file containing the `No-Support-for-Architecture-all` with any value other than `Packages` I throw an error instructing the user to open an issue. Such a case probably would signify some crazy new repo format with a thousand edge cases we can't handle anyway, so better to just have them open the issue...

I now have a short bit at the start for "extracting and interpreting" the `No-Support-for-Architecture-all` value, which then sets the internal `hybrid_format` boolean, which is then used throughout the rest of the sync.

mdellweg: That sounds like an excellent plan, not trying to cover unknown usecases by accident.

PR #454, opened 2022-03-16
Import export pulp_deb content

Manisha15:


pulp_deb/app/modelresource.py
```diff
@@ -0,0 +1,160 @@
...
+    ReleaseComponent,
+    ReleaseFile,
+)
+
+
+class AptContentResource(BaseContentResource):
+    """
+    Resource for import/export of apt content.
+
+    This Content Resource also handles exporting of any content for subrepos.
```
mdellweg, line 27: Is subrepos a _thing_ in pulp_deb?

quba42: Nope. Probably a copy paste artifact in the docstring from pulp_rpm.

Manisha15: Yeah, docstring was copied from pulp_rpm. It should be fixed now.

pulp_deb/app/modelresource.py
```diff
@@ -0,0 +1,158 @@
...
+    PackageIndex,
+    PackageReleaseComponent,
+    Release,
+    ReleaseArchitecture,
+    ReleaseComponent,
+    ReleaseFile,
+)
+
+
+class AptContentResource(BaseContentResource):
```
mdellweg, line 23: Ok let me ask again. Do we need this base class in any way? I think, the whole `content_mapping` is to handle the subrepos content.

Manisha15: Removed this base class and directly use Pulpcore base class as pulp_deb repositories don't have subrepos.

hstct (commented):
I focused the review on the actual code and not the CI changes as I understood most of them are just changes to make the pipelines happy.

That being said I tested exporting and importing using the provided tests as well as doing it manually. Both worked fine and as expected.

The only note I have is that all tests will be converted to the `pytest` framework eventually but given the fact that those provided here are already written and are working this can be done in a separate PR.

Otherwise LGTM.

mdellweg:
Actually, i think you should not update from the plugin template by hand in a feature pr. These changes should come in via https://github.com/pulp/pulp_deb/pull/610

Manisha15:
> Actually, i think you should not update from the plugin template by hand in a feature pr. These changes should come in via #610

I only need the `install.sh` script to change using template_config.yml https://github.com/pulp/pulp_deb/blob/ef245d3d417a5924e9f19b9b8ef657276e1ed75b/.github/workflows/scripts/install.sh#L78 in order for tests to run successfully. 

Should I keep that script and revert the changes in other scripts?

mdellweg:
I see now. I merged the other PR. If you rebase on top of main and then reapply the template with your config adjustments, the difference should be minimal.

mdellweg:
> The only note I have is that all tests will be converted to the `pytest` framework eventually but given the fact that those provided here are already written and are working this can be done in a separate PR.

There is an example of a pytest test for import export:
https://github.com/pulp/pulp_ansible/blob/main/pulp_ansible/tests/functional/api/test_export_import.py

mdellweg (approved):
LGTM

quba42:
I may still add some minimal documentation that this feature now exists in a separate MR.

PR #455, opened 2022-03-16
Improve log message

quba42:
[noissue]

~~I am fairly certain, this condition never happens.~~ The only way for the underlying functions not to return anything should involve them throwing unhandled errors.

Further evidence: We routinely end up within the conditional in line 688, which checks for a very similar condition.

I am happy for the tests to convince me I am wrong of course! :stuck_out_tongue_winking_eye:


quba42:
I stand corrected: Apparently this happens if the Package index being synced is not in the `file_references`, which happens when we instruct a remote to sync udebs, but the repo contains no installer indexes.

I will think about adding a comment, or elese refactoring to make this clearer to the code reader.

PR #471, opened 2022-04-10

pulpbot:
[noissue]


template_config.yml
```diff
@@ -28,6 +28,7 @@ flake8: true
 github_org: pulp
 issue_tracker: github
 keep_ci_update_for_latest_branches: 5
```
quba42, line 30: Should we reduce this? Not sure if we currently still care about so many pulp_deb versions.

fao89: This isn't being used anymore, we replaced it for a setting that specifies the lower branch you want to keep update:
https://github.com/pulp/pulpcore/blob/main/template_config.yml#L37

quba42: Ok, in that case I will amend this PR to use the new setting and remove this one.

PR #485, opened 2022-04-21
Add upper bounds to requirements for release CI

quba42:
[noissue]

It looks like the release CI now requires upper bounds on all dependencies: https://github.com/pulp/pulp_deb/runs/6102557469?check_suite_focus=true

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
 pulpcore>=3.17.0,<3.20
-python-debian>=0.1.36
+python-debian>=0.1.36,<0.2.0


mdellweg (approved):

requirements.txt
```diff
@@ -1,2 +1,2 @@
 pulpcore>=3.17.0,<3.20
-python-debian>=0.1.36
+python-debian>=0.1.36,<0.2.0
```
mdellweg, line 2: This is equivalent to `python-debian~=0.1.36`.

quba42: And is `~=` somehow better? I find it less readable. Does it mean "the number behind the final dot is allowed to change"?

PR #526, opened 2022-05-24
Fix local tests broken by DeclarativeArtifacts

hstct:
DeclarativeArtifacts parameters need to explicitly set.

See: https://github.com/pulp/pulp_python/pull/536/files

### DIFF
--- a/pulp_deb/app/tasks/synchronizing.py
+++ b/pulp_deb/app/tasks/synchronizing.py
@@ -434,12 +434,12 @@ class DebUpdatePackageIndexAttributes(Stage):  # TODO: Needs a new name
                         relative_dir = os.path.dirname(d_content.content.relative_path)
                         filename = _uncompress_artifact(d_content.d_artifacts, relative_dir)
                         da = DeclarativeArtifact(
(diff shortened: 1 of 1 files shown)


ipanova (approved):

ipanova:
you might want to do same for DeclarativeFailsafeArtifact

hstct:
> you might want to do same for DeclarativeFailsafeArtifact

@ipanova thanks I've missed that.

quba42 (approved):
Code change looks right, and the tests are now green.

quba42:
@ipanova Does it make sense to backport this change to any existing Y-release branches, so as to eventually make those pulpcore 3.20 compatible?

ipanova:
> > @ipanova Does it make sense to backport this change to any existing Y-release branches, so as to eventually make those pulpcore 3.20 compatible?
> 
> 3.20 will be a breaking changes release

What @fao89 has said. With this change you should  actually bump the floor for pulpcore requirement https://github.com/pulp/pulp_deb/blob/main/requirements.txt#L1

Issue #531, opened 2022-05-31
Publication: simple mode does not generate a list of packages


tibeer:
**Version**
- pulpcore: latest (sha256:40956e8c5def23ac5ddcc2c307e23a9cdd8eaa2ce3c2c8b6a4f2dcb5221e3d07)
- pulp_deb: latest(sha256:40956e8c5def23ac5ddcc2c307e23a9cdd8eaa2ce3c2c8b6a4f2dcb5221e3d07)

Installed via podman-compose with the latest containers.

**Describe the bug**
Creating a publication with "simple" does not generate a list of packages for the web frontend.

**To Reproduce**

```sh
export name=test
pulp deb repository create --name "${name}"
pulp deb remote create --name "${name}" --url http://nginx.org/packages/debian --distribution buster
pulp deb distribution create --name "${name}" --base-path "${name}" --repository "${name}"
pulp deb publication create --repository "${name}" --simple True
pulp deb repository sync --name "${name}" --mirror --remote "${name}"
```

Head over to your Pulp content website (http://1.2.3.4/pulp/content/test) and search for the packages.

**Expected behavior**
The packages should be listed like in the "structured" mode.

http://1.2.3.4/pulp/content/test/pool/ is completely missing.

**Additional context**
Actual file structure that is generated:

- http://1.2.3.4/pulp/content/test/
- http://1.2.3.4/pulp/content/test/dists/
- http://1.2.3.4/pulp/content/test/dists/default/
- http://1.2.3.4/pulp/content/test/dists/default/Release
- http://1.2.3.4/pulp/content/test/dists/default/all/
- http://1.2.3.4/pulp/content/test/dists/default/all/binary-all/
- http://1.2.3.4/pulp/content/test/dists/default/all/binary-all/Packages
- http://1.2.3.4/pulp/content/test/dists/default/all/binary-all/Packages.gz

quba42:
Looking at your reproducer steps: Is it possible, that you created the distribution and publication before syncing any packages to the repository? If so, then the repo version you published would be expected to contain no packages, and as a result no pool folder.

Perhaps you were relying on the autopublish feature as present in pulp_rpm to update your publication after your sync? I am afraid there is currently no such feature for pulp_deb.

tibeer:
Different order of commands and it works! Sometimes it's the simple solutions.

If someone else is stumbling across this, this is how you update things after a sync:

```sh
export name=test
[...]

PR #533, opened 2022-06-01
added a quickstart documentation

tibeer:
Started messing with Pulp and had a pretty simple issue (#531).
Thought it would be a good idea to create a quickstart so others do not fall for the same issue.

Signed-off-by: Tim Beermann <beermann@osism.tech>

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -1,0 +1,50 @@
+Quickstart
+================================================================================
+
+.. include:: external_references.rst
+
+
+If you want to have a quick introduction on how to use this plugin and mirror
+an example remote, this is the right place.
+
+
+Example for a mirror
+--------------------------------------------------------------------------------
+
+After you've added the CLI plugin, you can start by creating a repository:
+
+.. code-block:: bash
+
+   export name=test
+   pulp deb repository create --name "${name}"
+
+As an example remote we are going to add the nginx repository. Once the remote is added, you should
+sync it directly. You can still sync it later if you want, therefore look a bit further down.
+If you want to mirror only on demand (and have only the meta-data information available), use ``--no-mirror`` instead of ``--mirror``.
+
+.. code-block:: bash
+
+   pulp deb remote create --name "${name}" --url http://nginx.org/packages/debian --distribution buster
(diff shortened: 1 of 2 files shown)


docs/quickstart.rst
```diff
@@ -0,0 +1,50 @@
...
+Sync (again) AFTER distribution and publication
+--------------------------------------------------------------------------------
+
+If you need to sync after you have already distributed and published a repository,
+use the following commands to do so. The autopublish feature like in ``pulp_rpm`` is not implemented (yet).
+
+.. code-block:: bash
+
+   pulp deb repository sync --name "${name}" --mirror --remote "${name}"
+   pulp deb distribution update --name "${name}" --repository "${name}" --base-path "${name}"
```
mdellweg, line 48: This is a no-op and should not be necessary.

tibeer: What exactly should not be necessary?
1. pulp deb repository sync
2. pulp deb distribution update
3. both above

mdellweg: Sorry, the distribution update is not needed, as the distribution is already set up for auto-distributing.

tibeer: fixed.

quba42: I may be mixing something up, but it was my understanding that there currently is no auto-distributing for pulp_deb...

tibeer:
not stale

quba42 (approved):
First of all my apologies for taking so long to get around to this. These examples look valuable for users to have.

I propose the following for moving forward with this: Rather than endlessly going back and forth on minor details (documentation can always be tweaked further), I will merge this as is and may then add a small commit of my own making minor adjustments on top.

If you agree, could you rebase this onto latest main branch one more time to give the tests a chance to turn green? I will then merge it.

tibeer:
> First of all my apologies for taking so long to get around to this. These examples look valuable for users to have.
> 
> I propose the following for moving forward with this: Rather than endlessly going back and forth on minor details (documentation can always be tweaked further), I will merge this as is and may then add a small commit of my own making minor adjustments on top.
> 
> If you agree, could you rebase this onto latest main branch one more time to give the tests a chance to turn green? I will then merge it.

@quba42 sure, done!

PR #534, opened 2022-06-01
Convert CharField to TextField

### LINKED ISSUE #532: Convert most CharField instances to TextField
There now seems to be a consensus for both pulpcore and pulp_rpm, that TextField is always better than CharField for Pulp.
I gather this is because of how PostgreSQL implements these two fields on the DB side.

For consistency pulp_deb should also follow suit.

See: https://github.com/pulp/pulp_rpm/pull/2515 as a reference.

hstct:
closes #532


pulp_deb/app/models/content.py
```diff
@@ -22,7 +22,7 @@ class GenericContent(Content):
     TYPE = "generic"
 
     relative_path = models.TextField(null=False)
-    sha256 = models.CharField(max_length=255, null=False)
+    sha256 = models.TextField(null=False)
```
mdellweg, line 25: I know, it's hard to keep up, but for checksum fields, we decided that the length is known exactly, so we do not expect this field to change.

pulp_deb/app/models/content.py
```diff
@@ -38,13 +38,13 @@ class ReleaseFile(Content):
...
-    suite = models.CharField(max_length=255)
-    distribution = models.CharField(max_length=255)
+    codename = models.TextField()
+    suite = models.TextField()
+    distribution = models.TextField()
     components = models.TextField(blank=True)
     architectures = models.TextField(blank=True)
     relative_path = models.TextField()
-    sha256 = models.CharField(max_length=255)
+    sha256 = models.TextField()
```
mdellweg, line 47: Same here

pulp_deb/app/models/content.py
```diff
@@ -80,10 +80,10 @@ class PackageIndex(Content):
     TYPE = "package_index"
 
     release = models.ForeignKey(ReleaseFile, on_delete=models.CASCADE)
-    component = models.CharField(max_length=255)
-    architecture = models.CharField(max_length=255)
+    component = models.TextField()
+    architecture = models.TextField()
     relative_path = models.TextField()
-    sha256 = models.CharField(max_length=255)
+    sha256 = models.TextField()
```
mdellweg, line 86: here

pulp_deb/app/models/content.py
```diff
@@ -113,10 +113,10 @@ class InstallerFileIndex(Content):
     FILE_ALGORITHM = {"SHA256SUMS": "sha256", "MD5SUMS": "md5"}  # Are there more?
 
     release = models.ForeignKey(ReleaseFile, on_delete=models.CASCADE)
-    component = models.CharField(max_length=255)
-    architecture = models.CharField(max_length=255)
+    component = models.TextField()
+    architecture = models.TextField()
     relative_path = models.TextField()
-    sha256 = models.CharField(max_length=255)
+    sha256 = models.TextField()
```
mdellweg, line 119: and here.

pulp_deb/app/models/content.py
hstct: Thanks for clarifying I missed that. I will revert the change for the checksum fields.

mdellweg (approved):
LGTM

pulp_deb/app/migrations/0018_auto_20220601_1254.py
```diff
@@ -0,0 +1,73 @@
+# Generated by Django 3.2.13 on 2022-06-01 12:54
```
quba42, line 1: Can we rename the auto-generated migration file to something more meaningful?

See the corresponding change in pulp_rpm as a reference: https://github.com/pulp/pulp_rpm/pull/2515/files

quba42 (changes requested):
Everything else LGTM.

quba42:
The failed tests appear to be related to https://github.com/pulp/pulp_deb/pull/541.

quba42 (approved):
Rebase to let the tests go green, then this is good to merge.

PR #540, opened 2022-06-08
Provide SigningService setup for publish tests

### LINKED ISSUE #402: Signature tests should provide proper SigningService setup
Author: @quba42 (quba42)

Redmine Issue: 8640, https://pulp.plan.io/issues/8640

---

The setup of the SigningService for the test pipeline is currently a dirty hack, that causes frequent failures and complications in new pipelines.

It should be possible for the test itself to handle the setup using the pulpcore-manager utility functions provided by pulp-smash.

hstct:
closes #402


mdellweg:
> This works perfectly fine on my local vagrant dev environment, even with multiple runs. But the CI fails because it cannot access the `django_db_blocker` fixture. Even though the `pytest-django` package is definitely being [installed](https://github.com/pulp/pulp_deb/runs/6863282901?check_suite_focus=true#step:12:1222). Resulting into this error:

I think there is some confusion around the kind of tests we perform here. As far as i understand, `pytest-django` is meant to run unittests against a (local) test database, usually performing them in a transaction that will never be committed. The tests we run here, are better described as functional or integration tests. They perform calls against a "real" server and so test the whole stack. Since this very server is a separate container in the ci, you cannot easily access the database under test from the testing process. There is some effort on the way to change this however.

Maybe you want to split your PR in two pieces, because the transition to using pytest tests would still be beneficial today. And the signing-service fixture may become in the future.

hstct:
> Maybe you want to split your PR in two pieces, because the transition to using pytest tests would still be beneficial today. And the signing-service fixture may become in the future.

That seems reasonable to me. To avoid adding a new `conftest.py` file I'll wait for another PR that I had already prepared (#544) to get merged before I split this one.

hstct:
I have updated this PR using [pulp_container](https://github.com/pulp/pulp_container/blob/main/pulp_container/tests/functional/conftest.py#L152-L221) as inspiration. They use the cli to create the signing service as a workaround rather than doing some django db shenanigans.

.github/workflows/scripts/post_before_script.sh
```diff
@@ -1,17 +0,0 @@
...
-  curl -L https://github.com/pulp/pulp-fixtures/raw/master/common/GPG-KEY-pulp-qe | cmd_stdin_prefix bash -c "cat > /tmp/GPG-KEY-pulp-qe"
-
-  # Add private pulp-qe test key and set ownertrust:
-  curl -L https://github.com/pulp/pulp-fixtures/raw/master/common/GPG-PRIVATE-KEY-pulp-qe | cmd_stdin_prefix gpg --import
-  echo "6EDF301256480B9B801EBA3D05A5E6DA269D9D98:6:" | cmd_stdin_prefix gpg --import-ownertrust
-
-  # Create the signing service:
-  cmd_prefix chmod a+x /tmp/setup_signing_service.py /root/sign_deb_release.sh
-  cmd_prefix /tmp/setup_signing_service.py /root/sign_deb_release.sh /tmp/GPG-KEY-pulp-qe
-fi
```
hstct, line 17: Pretty sure this file can be deleted as we do not need to set up the signing service beforehand anymore.

pulp_deb/tests/functional/conftest.py
```diff
@@ -46,3 +55,131 @@ def apt_verbatim_publication_api(apt_client):
...
+        sync_response = apt_repository_api.sync(repo.pulp_href, repository_sync_data)
+        return monitor_task(sync_response.task)
+
+    return _deb_sync_repository
+
+
+@pytest.fixture(scope="session")
+def signing_script_filename(signing_gpg_homedir_path):
+    """A fixture for a script that is suited for signing packages."""
+    raw_script = (
```
quba42, line 102: Can we read this from `pulp_deb/tests/functional/sign_deb_release.sh` instead?
This would preserve the script as a separate file, that we can point to from the docs as an example signing script (which we are doing right now).

hstct: So I had another go at it and my best solution for this was to emulate the `raw_script` variable by reading the contents of the `pulp_deb/tests/functional/sign_deb_release.sh` script line by line and transform it to a tuple. Something like this:

```python
    dirname = os.path.dirname(__file__)
    filename = os.path.join(dirname, "sign_deb_release.sh")

    with open (filename) as fi:
        lines = [line.rstrip() for line in fi]

    raw_script = tuple(l for l in lines)
```

However this results in a different error:

```
E           pulp_smash.exceptions.CalledProcessError: Command ('pulpcore-manager', 'add-signing-service', '1a529327-8b68-4ea1-b45b-3634677a1e69', '/tmp/pytest-of-vagrant/pytest-13/6a872b43-63c9-4272-8b12-61eb912029940/bash-script.sh', '05A5E6DA269D9D98', '--class', 'deb:AptReleaseSigningService', '--gnupghome', local('/tmp/pytest-of-vagrant/pytest-13/6a872b43-63c9-4272-8b12-61eb912029940')) returned non-zero exit status 1.
E
E           stdout:
E
E           stderr: Traceback (most recent call last):
E             File "/usr/local/lib/pulp/bin/pulpcore-manager", line 8, in <module>
E               sys.exit(manage())
E             File "/home/vagrant/devel/pulpcore/pulpcore/app/manage.py", line 11, in manage
E               execute_from_command_line(sys.argv)
E             File "/usr/local/lib/pulp/lib64/python3.10/site-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
E               utility.execute()
E             File "/usr/local/lib/pulp/lib64/python3.10/site-packages/django/core/management/__init__.py", line 413, in execute
E               self.fetch_command(subcommand).run_from_argv(self.argv)
E             File "/usr/local/lib/pulp/lib64/python3.10/site-packages/django/core/management/base.py", line 354, in run_from_argv
E               self.execute(*args, **cmd_options)
E             File "/usr/local/lib/pulp/lib64/python3.10/site-packages/django/core/management/base.py", line 398, in execute
E               output = self.handle(*args, **options)
E             File "/home/vagrant/devel/pulpcore/pulpcore/app/management/commands/add-signing-service.py", line 83, in handle
E               SigningService.objects.create(
E             File "/usr/local/lib/pulp/lib64/python3.10/site-packages/django/db/models/manager.py", line 85, in manager_method
E               return getattr(self.get_queryset(), name)(*args, **kwargs)
E             File "/usr/local/lib/pulp/lib64/python3.10/site-packages/django/db/models/query.py", line 453, in create
E               obj.save(force_insert=True, using=self.db)
E             File "/home/vagrant/devel/pulpcore/pulpcore/app/models/content.py", line 826, in save
E               self.validate()
E             File "/home/vagrant/devel/pulp_deb/pulp_deb/app/models/signing_service.py", line 41, in validate
E               return_value = self.sign(test_release_path)
E             File "/home/vagrant/devel/pulpcore/pulpcore/app/models/content.py", line 766, in sign
E               raise RuntimeError(str(completed_process.stderr))
E           RuntimeError: b'gpg: skipped "Pulp QE": No secret key\ngpg: signing failed: No secret key\n'

/usr/local/lib/pulp/lib64/python3.10/site-packages/pulp_smash/cli.py:128: CalledProcessError
```

I'm not really sure what else I can try at this point. No Idea why defining the tuple variable explicitly works and why reading it from a file doesn't.

mdellweg: Is it this line?
`f'export GNUPGHOME="{signing_gpg_homedir_path}"',`

hstct: > Is it this line? `f'export GNUPGHOME="{signing_gpg_homedir_path}"',`

Indeed! I totally overlooked that line. I have to add it manually in the fixture but it works now.

Reading the lines from the shell script is adding some boilerplate to the code. However this way the script its own file and could be technically be reused. It could be removed at a later point once the metadata signing gets updated in the docs.

quba42 (approved):

PR #550, opened 2022-06-20
Make ReleaseFile and PackageIndex artifacts immutable

### LINKED ISSUE #549: When the circumstances are just right, pulp_deb will produce inconsistent verbatim publications
**Version**
Affects every version tested (at least the cutting edge all the way back to 3.14)

**Describe the bug**
Does not happen with download policy immediate, (reproduced with streamed, but on_demand is probably the same).

When re-syncing a repo with a unchanged Package index file, but a changed associated `Package.gz` file. (Can happen when syncing from another Pulp instance that uses a new APT publication with the same repo version.), then the existing PackageIndex is given a new artifact, rather than creating a new PackageIndex content. As a result existing repo versions have a release file referencing a `Package.gz` file with a different checksum than the one now included. When verbatim publishing such a repo versions clients will rightly complain about failed checksums.

**Expected behavior**
Metadata content should be completely immutable. If anything changes new metadata should be created, not existing content altered!

**Additional context**
The problem is the uniqueness constraints on the affected content types!

quba42:
Closes #549
https://github.com/pulp/pulp_deb/issues/549


pulp_deb/app/tasks/synchronizing.py
```diff
@@ -443,6 +444,18 @@ async def run(self):
...
                         await _save_artifact_blocking(da)
+                    file_dict = {}
+                    for da in d_content.d_artifacts:
+                        filename = os.path.basename(da.relative_path)
+                        sha256 = da.artifact.sha256
+                        file_dict[filename] = sha256
+                    hash_string = ""
+                    file_list = ["Packages", "Packages.gz", "Packages.xz", "Release"]
+                    for filename in file_list:
+                        hash_string = hash_string + file_dict.get(filename, "")
```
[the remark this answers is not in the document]
quba42: This also makes sense in case we ever expand the set of possible artifacts for one of these types. => Will add a placeholder.

quba42:
A unit test for `_get_artifact_set_sha256()` would be perfect for this change!

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -1027,6 +1034,23 @@ def _save_artifact_blocking(d_artifact):
         d_artifact.artifact.touch()
 
 
+def _get_artifact_set_sha256(d_content, supported_artifacts):
```
quba42, line 1037: I think a unit test for this function, would be exactly what is needed to test this change!

quba42: Added two unit tests. :wink:

hstct (approved):
LGTM

PR #555, opened 2022-06-23

pulpbot:
[noissue]


.ci/ansible/smash-config.json
```diff
@@ -6,7 +6,7 @@
     ],
     "selinux enabled": false,
     "version": "3",
-    "aiohttp_fixtures_origin": "172.0.0.1"
+    "aiohttp_fixtures_origin": "127.0.0.1"
```
quba42, line 9: I think I added a nonsense value to the template_config here...

However, apparently this does not stop this job from trying to correct, and it also did not cause any tests to fail?!

PR #570, opened 2022-07-04
Add optimize mode for sync tasks

### LINKED ISSUE #564: Add optimize mode to repository sync
Similar to `pulp_rpm` add the option for the `synchronize` task to optimize sync (which will be the default behavior).

The idea is to check if the `artifact_set_sha256` of a `ReleaseFile` or `PackageIndex` has changed before syncing it. If the checksum has not changed that means the release or the given package has not changed and can be skipped in the syncing process. This should yield better performance for syncing overall.

hstct:
Required PR: https://github.com/pulp/pulpcore/pull/2999

closes #564

Co-Authored-By: Quirin Pamp <pamp@atix.de>

This will change the default behaviour of the sync to use optimize
unless it is explicitly set to False.

### DIFF
--- a/pulp_deb/app/tasks/synchronizing.py
+++ b/pulp_deb/app/tasks/synchronizing.py
@@ -22,7 +22,6 @@ from pulpcore.plugin.models import (
     Artifact,
     ProgressReport,
     Remote,
-    Repository,
 )
 
 from pulpcore.plugin.stages import (
@@ -51,6 +50,7 @@ from pulp_deb.app.models import (
     PackageReleaseComponent,
     InstallerPackage,
     AptRemote,
+    AptRepository,
 )
 
 from pulp_deb.app.serializers import (
@@ -137,7 +137,7 @@ class UnknownNoSupportForArchitectureAllValue(Exception):
     pass
 
 
-def synchronize(remote_pk, repository_pk, mirror):
+def synchronize(remote_pk, repository_pk, mirror, optimize):
     """
     Sync content from the remote repository.
 
@@ -147,18 +147,20 @@ def synchronize(remote_pk, repository_pk, mirror):
         remote_pk (str): The remote PK.
         repository_pk (str): The repository PK.
         mirror (bool): True for mirror mode, False for additive.
+        optimize (bool): Optimize mode.
 
     Raises:
         ValueError: If the remote does not specify a URL to sync
 
     """
     remote = AptRemote.objects.get(pk=remote_pk)
-    repository = Repository.objects.get(pk=repository_pk)
+    repository = AptRepository.objects.get(pk=repository_pk)
+    previous_repo_version = repository.latest_version()
 
     if not remote.url:
         raise ValueError(_("A remote must have a url specified to synchronize."))
 
-    first_stage = DebFirstStage(remote)
+    first_stage = DebFirstStage(remote, optimize, mirror, previous_repo_version)
     DebDeclarativeVersion(first_stage, repository, mirror=mirror).create()
 
 
@@ -209,6 +211,7 @@ class DebDeclarativeVersion(DeclarativeVersion):
             list: List of :class:`~pulpcore.plugin.stages.Stage` instances
 
         """
+        self.first_stage.new_version = new_version
         pipeline = [
             self.first_stage,
             QueryExistingArtifacts(),
@@ -498,17 +501,32 @@ class DebFirstStage(Stage):
     The first stage of a pulp_deb sync pipeline.
     """
 
-    def __init__(self, remote, *args, **kwargs):
+    def __init__(self, remote, optimize, mirror, previous_repo_version, *args, **kwargs):
         """
         The first stage of a pulp_deb sync pipeline.
 
         Args:
-            remote (FileRemote): The remote data to be used when syncing
-
+            remote (AptRemote): The remote data to be used when syncing
+            optimize (Boolean): If optimize mode is enabled or not
+            previous_repo_version repository (RepositoryVersion): The previous RepositoryVersion.
         """
         super().__init__(*args, **kwargs)
         self.remote = remote
+        self.optimize = optimize
+        self.previous_repo_version = previous_repo_version
+        self.previous_sync_info = defaultdict(dict, previous_repo_version.info)
+        self.sync_info = defaultdict()
+        self.sync_info["remote_options"] = self._gen_remote_options()
+        self.sync_info["sync_options"] = {
+            "optimize": optimize,
+            "mirror": mirror,
+        }
         self.parsed_url = urlparse(remote.url)
+        self.sync_options_unchanged = (
+            self.previous_sync_info["remote_options"] == self.sync_info["remote_options"]
+            and self.previous_sync_info["sync_options"]["mirror"]
+            == self.sync_info["sync_options"]["mirror"]
+        )
 
     async def run(self):
         """
@@ -520,6 +538,8 @@ class DebFirstStage(Stage):
         await asyncio.gather(
             *[self._handle_distribution(dist) for dist in self.remote.distributions.split()]
         )
+
+        self.new_version.info = self.sync_info
 
     async def _create_unit(self, d_content):
         await self.put(d_content)
@@ -535,6 +555,19 @@ class DebFirstStage(Stage):
             remote=self.remote,
             deferred_download=False,
         )
+
+    def _gen_remote_options(self):
+        return {
+            "distributions": self.remote.distributions,
+            "components": self.remote.components,
+            "architectures": self.remote.architectures,
+            "policy": self.remote.policy,
+            "sync_sources": self.remote.sync_sources,
+            "sync_udebs": self.remote.sync_udebs,
+            "sync_installer": self.remote.sync_installer,
+            "gpgkey": self.remote.gpgkey,
+            "ignore_missing_package_indices": self.remote.ignore_missing_package_indices,
+        }
 
     async def _handle_distribution(self, distribution):
         log.info(_('Downloading Release file for distribution: "{}"').format(distribution))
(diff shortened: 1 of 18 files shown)


quba42:
I would adjust the structure of `last_sync_details` somewhat as follows:
```
last_sync_details:
  remote_options:
    distributions: "<value>"
    components: "<value>"
    architectures: "<value>"
    download_policy: "<value>"
  <distribution>:
    artifact_set_sha256: "<value>"
    <component>:
      <architecture>:
        artifact_set_sha256: "<value>"
```
Remote options belong on the top level, since they are the same for all distributions.
We also need to make sure download_policy is one of the remote options we check.
Rather than having a list of (relative_path, package_index_artifact_set_sha256) tuples we could just make the comonents and architectures nested dict keys just like the distribution. This is mostly a preference thing, and should not ultimately make much of a difference. (The relative_path, can be generated from the component and the architecture, so this contains essentially the same information).

mdellweg:
Let me just add another thought here. Can we not add information to the repository-version itself how it was made? This would be immutable and the repository can always look at it's latest version to retrieve the thing you are implementing here.

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -498,16 +499,19 @@ class DebFirstStage(Stage):
...
-            remote (FileRemote): The remote data to be used when syncing
-
+            remote (AptRemote): The remote data to be used when syncing
+            optimize (Boolean): If optimize mode is enabled or not
+            repository (AptRepository): The repository data used for last sync detail checks
         """
         super().__init__(*args, **kwargs)
         self.remote = remote
+        self.optimize = optimize
+        self.repository = repository
```
quba42, line 514: Might be enough to just save the last_sync_details here?
```
self.last_sync_details = repository.last_sync_details
```
I don't think we are using anything else from the repository and this would make the dict accesses shorter/more readable.

quba42: Never mind we _are_ using the repository to set things for next time...

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -553,7 +573,32 @@ async def _handle_distribution(self, distribution):
...
         if release_file is None:
             return
-        # Create release object
+        if self.optimize and distribution in self.repository.last_sync_details:
+            last_sync_dist_details = self.repository.last_sync_details[distribution]
+            if not self._has_remote_changed():
+                if (
+                    last_sync_dist_details["artifact_set_sha256"]
+                    == release_file.artifact_set_sha256
+                ):
```
quba42, line 582: The entire condition could go in a single level as follows:
```suggestion
        if (
            self.optimize
            and distribution in self.repository.last_sync_details
            and not self._has_remote_changed()
            and self.repository.last_sync_details[distribution]["artifact_set_sha256"]
            == release_file.artifact_set_sha256
        ):
```

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -553,7 +573,32 @@ async def _handle_distribution(self, distribution):
...
+                    return
+        self.repository.last_sync_details["remote_options"] = self._gen_remote_options()
+        if distribution in self.repository.last_sync_details:
+            self.repository.last_sync_details[distribution][
+                "artifact_set_sha256"
+            ] = release_file.artifact_set_sha256
+        else:
+            self.repository.last_sync_details[distribution] = {
+                "artifact_set_sha256": release_file.artifact_set_sha256,
+            }
```
quba42, line 600: This looks like another argument for saving these dicts on the repository version. In that case we are always building a new dict, and never overwriting an existing one, so we would not need to go through these contortions and have something that is hopefully a lot less error prone...

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -799,6 +844,45 @@ async def _handle_package_index(
             else:
                 raise NoPackageIndexFile(relative_dir=package_index_dir)
 
+        distribution = release_file.distribution
+        component = release_component.component
```
quba42, line 848: Reminder to myself: Double check if we should use `.component` or `.plain_component` here!

quba42 (changes requested):
Most of my suggestions just reflect small style changes.
Overall this looks pretty good to me.
There are still some details where I added a reminder to double check things.

The big question is whether we want to go ahead with the `repository.last_sync_details` design, or if we want to rework to use something like `repo_version.creation_details` instead. Unfortunately it does not make sense to punt down the road by accepting the implementation we have and then "maybe adjusting it later", since we should get these model changes right from the get go.

pulp_deb/app/models/repository.py
```diff
@@ -17,6 +18,14 @@
 )
 
 
+class AptRepositoryVersion(RepositoryVersion):
+    """
+    A RepositoryVersion for DebContent.
+    """
+
+    sync_details = models.JSONField(default=dict)
```
mdellweg, line 26: Can we maybe not subclass this and add some extra_data field on the base class?

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -147,18 +147,19 @@ def synchronize(remote_pk, repository_pk, mirror):
...
         mirror (bool): True for mirror mode, False for additive.
+        optimize (bool): Optimize mode.
 
     Raises:
         ValueError: If the remote does not specify a URL to sync
 
     """
     remote = AptRemote.objects.get(pk=remote_pk)
-    repository = Repository.objects.get(pk=repository_pk)
+    repository = AptRepository.objects.get(pk=repository_pk)
```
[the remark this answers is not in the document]
mdellweg: I guess, you are right for calling the correct `finalize_repository_version` method on it.

So does this fix an existing bug here?

pulp_deb/app/models/repository.py
quba42: It does look a lot like there is no acceptable way to do this just on the plugin side. We will trial adding the field on the pulpcore side. If that works, we will open a pulpcore issue and PR, if not, we are back to the pulp_rpm design of saving the information on the repository in a mutable fashion. :wink:

hstct:
The current version relies on https://github.com/pulp/pulpcore/pull/2999 to be merged before the tests can pass.

pulp_deb/app/models/repository.py
mdellweg: Yes, a pulpcore addition is what i had in mind.

pulp_deb/app/tasks/synchronizing.py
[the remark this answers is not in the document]
quba42: I am not aware it previously caused a problem, but it does strike me as better to change it.

quba42:
The latest version of this PR, exclusively stores remote options on the new RepositoryVersion field from pulpcore.
The needed `artifact_set_sha256` comparison values are instead queried from the content in the previous repository version (this information has always been present there).

We have successfully tested skipping entire Release files as well as individual package indices. As such this change is now mostly ready for final review.

quba42 (approved):
I now approve these changes, but since I have been more involved in the latest version as an author, I will look for another secondary review amongst my colleagues. :wink:

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -733,6 +771,14 @@ async def _handle_flat_repo(self, file_references, release_file, release):
         # Await all tasks
         await asyncio.gather(*pending_tasks)
 
+    def _nested_defaultdict(self, existing=None, **kwargs):
```
quba42, line 774: We probably don't need this anymore!

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -1017,6 +1077,37 @@ async def _handle_translation_files(self, release_file, release_component, file_
             )
 
 
+@sync_to_async
+def _readd_previous_package_indices(previous_version, new_version, distribution):
+    new_version.add_content(
+        previous_version.get_content(
+            PackageIndex.objects.filter(relative_path__contains=distribution)
+        )
+    )
```
quba42, line 1086: Also re-add InstallerFileIndexes!

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -536,6 +540,22 @@ def _to_d_artifact(self, relative_path, data=None):
...
+            if ro_last_sync[key] != ro_compare_dict[key]:
+                return True
+        return False
+
+    def _gen_remote_options(self):
+        return {
+            "distributions": self.remote.distributions,
+            "components": self.remote.components,
+            "architectures": self.remote.architectures,
+            "policy": self.remote.policy,
```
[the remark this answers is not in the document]
quba42: For the remote options we should add the following fields:
```
sync_sources
sync_udebs
sync_installer
gpgkey
ignore_missing_package_indices
```

Finally, not a remote option, but if the value for `mirror` sync policy has changed, we also need to do a full sync!

pulp_deb/app/tasks/synchronizing.py
quba42: Wasn't being called anymore, removed.

pulp_deb/app/tasks/synchronizing.py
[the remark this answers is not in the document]
quba42: Done.

pulp_deb/app/tasks/synchronizing.py
quba42: Done.

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -498,17 +501,32 @@ class DebFirstStage(Stage):
...
         """
         super().__init__(*args, **kwargs)
         self.remote = remote
+        self.optimize = optimize
+        self.previous_repo_version = previous_repo_version
+        self.previous_sync_info = defaultdict(dict, previous_repo_version.info)
+        self.sync_info = defaultdict()
+        self.sync_info["remote_options"] = self._gen_remote_options()
+        self.sync_info["sync_options"] = {
+            "optimize": optimize,
```
quba42, line 521: Not actually needed but possibly interesting information for debugging purposes.

quba42:
After final internal review/debate, we have decided to merge the current state of this new feature (once the tests turn green).

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -147,18 +147,20 @@ def synchronize(remote_pk, repository_pk, mirror):
...
         mirror (bool): True for mirror mode, False for additive.
+        optimize (bool): Optimize mode.
 
     Raises:
         ValueError: If the remote does not specify a URL to sync
 
     """
     remote = AptRemote.objects.get(pk=remote_pk)
-    repository = Repository.objects.get(pk=repository_pk)
+    repository = AptRepository.objects.get(pk=repository_pk)
```
quba42, line 157: This ended up revealing a previously masked problem with our sanity checks.
We fixed it again using: https://github.com/pulp/pulp_deb/pull/633

PR #572, opened 2022-07-07
Convert download content test to pytest

hstct:
[noissue]

### DIFF
--- a/pulp_deb/tests/functional/api/test_download_content.py
+++ b/pulp_deb/tests/functional/api/test_download_content.py
@@ -1,7 +1,6 @@
-# coding=utf-8
 """Tests that verify download of content served by Pulp."""
+import pytest
 import hashlib
-import unittest
 from random import choice
 from urllib.parse import urljoin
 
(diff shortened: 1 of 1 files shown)


quba42 (approved):
As far as I can tell, this does what the commit message says, namely convert the test to pytest without changing what is being tested.

Looking over this test, I can think of several ways the test could be improved (that are not related to this conversion, so out of scope for this PR):

- The old test was "pretending" to loop over `DOWNLOAD_POLICIES` without actually using the different values anywhere. `policy` is a value set for the remote. We could expand this test to test download for each of the policies. I don't think this would be needed for both the verbatim and the apt publisher, so it would probably require some restructuring.
- Improving test efficiency: The `gen_repository_with_synced_remote` fixture, should be the same for both verbatim and normal version of the test. Are we currently generating this fixture twice (for each version), or are we generating it once and then reusing for both tests?
- It is pretty hard to tell what these tests are actually testing, because the crucial bits are hidden away in `pulp_deb.tests.functional.utils`. Just something to keep in the back of our mind as we continue to work on tests!

My question is: Should we record any of these points somewhere? Perhaps a comment in the test source file? (I don't want to open a bunch of issues we won't look at for a long time).

hstct:
> The old test was "pretending" to loop over `DOWNLOAD_POLICIES` without actually using the different values anywhere. `policy` is a value set for the remote. We could expand this test to test download for each of the policies. I don't think this would be needed for both the verbatim and the apt publisher, so it would probably require some restructuring.

The testing of the policies will be part of the conversion of this test https://github.com/pulp/pulp_deb/blob/main/pulp_deb/tests/functional/api/test_download_policies.py

> Improving test efficiency: The gen_repository_with_synced_remote fixture, should be the same for both verbatim and normal version of the test. Are we currently generating this fixture twice, or are we generating it once and then reusing for both tests?

iirc it will only generate the fixture once per test run

> It is pretty hard to tell what these tests are actually testing, because the crucial bits are hidden away in `pulp_deb.tests.functional.utils`. Just something to keep in the back of our mind as we continue to work on tests!

For now I will continue with the converting tests 1:1. After that I had planned to go over it again and clean up fixtures and other stuff hopefully. This way one has better overview on what is actually needed for the `pytest` tests and what can be dropped, replaced, simplified. I will create an issue for that once there is a clear overview what all these changes entail.

PR #585, opened 2022-07-14
Make more dependencies explicit

quba42:
[noissue]

Note that I have no intention of making any django or drf dependencies
explicit. I want those to be handled by pulpcore. If pulpcore decides to
throw something out, I *want* things to break, so I am forced to fix
them for the next release.

Other candidates for handling via pulpcore: asgiref, asyncio, aiohttp

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,6 @@
+# All things django and asyncio are deliberately left to pulpcore
+# Example transitive requirements: asgiref, asyncio, aiohttp
 pulpcore>=3.21.0.dev,<3.25
 python-debian>=0.1.44,<0.2.0
+python-gnupg>=0.4.9,<0.6
+jsonschema>=4.6,<5.0


requirements.txt
```diff
@@ -1,2 +1,7 @@
 pulpcore>=3.20.0.dev
 python-debian>=0.1.44,<0.2.0
+python-gnupg~=0.4.9
```
bmbouter, line 3: I recommend using a range compatibility to avoid incompatible dependency situations. Basically, declare as broad a compatibility as you can. If you're unsure what the lower bound should be, I recommend leaving it unset, until you hear of a problem and then either add compat for that version or add the lower bound then.

I'm going to try to document this in the plugin writer docs soon, but I wanted to drop my advice here also.

hstct (approved):
I don't see a problem in making these dependencies more explicit. Since those are both required by `pulp_deb` it adds as sanity check if they somehow are not included anymore in parent dependencies.

PR #586, opened 2022-07-15
Add structured repo docs

daviddavis:
The script in the docs will be tested by the CI:

https://github.com/pulp/pulp_deb/blob/0429dddf355f7917b047b2696fc15148c641cc8e/.github/workflows/scripts/script.sh#L20
https://github.com/pulp/pulp_deb/blob/0429dddf355f7917b047b2696fc15148c641cc8e/.github/workflows/scripts/script.sh#L41-L43

### DIFF
--- a/CHANGES/586.docs
+++ b/CHANGES/586.docs
@@ -1,0 +1,1 @@
+Added workflow docs on manually creating structured repos.
--- a/docs/_scripts/setup.sh
+++ b/docs/_scripts/setup.sh
@@ -1,0 +1,30 @@
+#!/usr/bin/env bash
+set -e
+
+export BASE_ADDR=${BASE_ADDR:-http://localhost:24817}
+export CONTENT_ADDR=${CONTENT_ADDR:-http://localhost:24816/pulp/content}
+
+# Poll a Pulp task until it is finished.
+wait_until_task_finished() {
+    echo "Polling the task until it has reached a final state."
+    local task_url=$1
+    while true
+    do
+        local response=$(http $task_url)
+        local state=$(echo ${response} | jq -r .state)
+        case ${state} in
+            failed|canceled)
+                echo "Task in final state: ${state}"
+                exit 1
+                ;;
+            completed)
+                echo "$task_url complete."
+                break
+                ;;
+            *)
+                echo "Still waiting..."
+                sleep 1
+                ;;
+        esac
+    done
+}
--- a/docs/_scripts/structured_repo.sh
+++ b/docs/_scripts/structured_repo.sh
@@ -1,0 +1,35 @@
+#!/usr/bin/env bash
+
+set -ev
+
+trap "rm frigg_1.0_ppc64.deb" EXIT
+
+# download a package
+wget https://fixtures.pulpproject.org/debian/pool/asgard/f/frigg/frigg_1.0_ppc64.deb
+
+# upload the package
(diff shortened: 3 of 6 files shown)


hstct (commented):
If we can add an expected output this looks good to me.

docs/_scripts/setup.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+set -e
+
+export BASE_ADDR=${BASE_ADDR:-http://localhost:24817}
```
hstct, line 4: The current documentation doesn't define the hostname:
```
export BASE_ADDR=http://<hostname>:24817
```

However this is probably fine as is I just added this as a note rather than a suggestion.

docs/_scripts/structured_repo.sh
```diff
@@ -0,0 +1,35 @@
...
+# add our content to the repository
+TASK_HREF=$(http ${BASE_ADDR}${REPO_HREF}modify/ add_content_units:="[\"$RELEASE_HREF\", \"$COMP_HREF\", \"$PACKAGE_HREF\", \"$PKG_COMP_HREF\", \"$ARCH_HREF\"]" | jq -r .task)
+wait_until_task_finished $BASE_ADDR$TASK_HREF
+
+# publish our repo
+TASK_HREF=$(http ${BASE_ADDR}/pulp/api/v3/publications/deb/apt/ repository=$REPO_HREF structured=true | jq -r .task)
+wait_until_task_finished $BASE_ADDR$TASK_HREF
+
+# check that our repo has our release
+http --check-status ${CONTENT_ADDR}/myrepo/dists/mydist/mycomp/binary-ppc64/
```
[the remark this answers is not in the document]
daviddavis: Added to structured_repo.rst.

docs/_scripts/setup.sh
daviddavis: This code is being run by the CI (see post_docs_test.sh) whereas the commands from the other pulp_deb workflow docs aren't. I don't think `export BASE_ADDR=http://<hostname>:24817` is valid bash.

CHANGES/586.docs
```diff
@@ -0,0 +1 @@
+Adding structured repo workflow docs.
```
quba42, line 1:
```suggestion
Added workflow docs on manually creating structured repos.
```
I am hoping this will make clearer that we are talking about structured repos not created via sync.

docs/_scripts/setup.sh
quba42: No, in the past we have not made the CI run any docs workflows.
In general, everything that is there already, is due for a rework.

quba42 (changes requested):
Since a full rework of the existing docs won't happen any time soon, and this PR brings an obvious benefit (docs for a workflow that was previously undocumented), I am not too worried about making the style perfectly consistent with the existing docs.

Since @hstct says he tried out the workflow and it worked, I say we merge. Perhaps after my nitpick with the change note is resolved :smile:.

Thanks a ton @daviddavis for providing these docs, and @hstct for the initial review!

CHANGES/586.docs
mdellweg: It also makes it clearer that it's not the docs being structured. ;)

hstct (approved):

quba42 (approved):

PR #600, opened 2022-08-04
Convert download policies test to pytest

hstct:
[noissue]

### DIFF
--- a/pulp_deb/tests/functional/api/test_crud_remotes.py
+++ b/pulp_deb/tests/functional/api/test_crud_remotes.py
@@ -5,189 +5,189 @@ import pytest
 from pulpcore.client.pulp_deb.exceptions import ApiException
 
 from pulp_smash import utils
-from pulp_smash.pulp3.bindings import monitor_task
 
 from pulp_deb.tests.functional.constants import DOWNLOAD_POLICIES, DEB_SIGNING_KEY
 from pulp_deb.tests.functional.utils import gen_deb_remote
 
 
-def test_create_remote_repository(verbose_remote_data, remote_repository):
-    """Test creation of the remote repository.
-    NOTE: The repository is actually created in this files fixture and this
-          test only confirms its creation.
-    """
-    for key, val in verbose_remote_data.items():
+@pytest.mark.parallel
+def test_create_remote_repository(deb_remote_custom_data_factory):
+    """Test creation of the remote."""
+    data = gen_verbose_remote_data()
+    remote = deb_remote_custom_data_factory(data)
+
+    for key, val in data.items():
         if key != "username" and key != "password":
-            assert remote_repository.to_dict()[key] == val
+            assert remote.to_dict()[key] == val
 
 
-def test_create_remote_repository_with_same_name(
-    apt_remote_api, gen_object_with_cleanup, remote_repository, gen_verbose_remote_data
-):
-    """Verify whether it is possible to create a remote repository with the same
-    name.
-    """
-    body = gen_verbose_remote_data()
-    body["name"] = remote_repository.name
+@pytest.mark.parallel
+def test_create_remote_repository_with_same_name(deb_remote_custom_data_factory):
+    """Verify whether it is possible to create a remote with the same name."""
+    data = gen_verbose_remote_data()
+    deb_remote_custom_data_factory(data)
+
     with pytest.raises(ApiException) as exc:
-        gen_object_with_cleanup(apt_remote_api, body)
+        deb_remote_custom_data_factory(data)
+
     assert exc.value.status == 400
 
 
 @pytest.mark.parallel
-def test_create_remote_repository_without_url(
-    apt_remote_api, gen_object_with_cleanup, gen_verbose_remote_data
-):
-    """Verify whether it is possible to create a remote repository without an URL.
-    This test targets the following issues:
-    * `Pulp #3395 <https://pulp.plan.io/issues/3395>`_
-    * `Pulp Smash #984 <https://github.com/pulp/pulp-smash/issues/984>`_
-    """
-    body = gen_verbose_remote_data()
-    del body["url"]
+def test_create_remote_repository_without_url(deb_remote_custom_data_factory):
+    """Verify whether it is possible to create a remote without an URL."""
+    data = gen_verbose_remote_data()
+    del data["url"]
+
     with pytest.raises(ApiException) as exc:
-        gen_object_with_cleanup(apt_remote_api, body)
+        deb_remote_custom_data_factory(data)
+
     assert exc.value.status == 400
 
 
-def test_read_remote_by_href(apt_remote_api, remote_repository):
+@pytest.mark.parallel
+def test_read_remote_by_href(
+    deb_remote_custom_data_factory,
+    deb_get_remote_by_href,
+):
     """Verify whether it is possible to read a remote repository by its href."""
-    href = apt_remote_api.read(remote_repository.pulp_href)
-    for key, val in remote_repository.to_dict().items():
-        assert href.to_dict()[key] == val
+    data = gen_verbose_remote_data()
+    remote = deb_remote_custom_data_factory(data)
+    read_remote = deb_get_remote_by_href(remote.pulp_href)
+
+    for key, val in remote.to_dict().items():
+        assert read_remote.to_dict()[key] == val
 
 
-def test_read_remote_repository_by_name(apt_remote_api, remote_repository):
+@pytest.mark.parallel
+def test_read_remote_by_name(
+    deb_remote_custom_data_factory,
+    deb_get_remotes_by_name,
+):
     """Verify whether it is possible to read a remote repository by its name."""
-    page = apt_remote_api.list(name=remote_repository.name)
-    assert len(page.results) == 1
-    for key, val in remote_repository.to_dict().items():
-        assert page.results[0].to_dict()[key] == val
+    data = gen_verbose_remote_data()
+    remote = deb_remote_custom_data_factory(data)
+    read_remote = deb_get_remotes_by_name(remote.name)
+
+    assert len(read_remote.results) == 1
+
+    for key, val in remote.to_dict().items():
+        assert read_remote.results[0].to_dict()[key] == val
 
 
-def test_update_remote_repository_with_patch(
-    apt_remote_api, gen_verbose_remote_data, remote_repository
+@pytest.mark.parallel
+def test_patch_remote(
+    deb_remote_custom_data_factory,
+    deb_get_remote_by_href,
+    deb_patch_remote,
 ):
-    """Test if a remote repository is updated by PATCH."""
-    previous_repo = apt_remote_api.read(remote_repository.pulp_href)
-    body = gen_verbose_remote_data()
-    response = apt_remote_api.partial_update(remote_repository.pulp_href, body)
-    monitor_task(response.task)
-    repo = apt_remote_api.read(remote_repository.pulp_href)
-    for key, val in body.items():
+    """Verify whether it is possible to update a remote with PATCH."""
+    data = gen_verbose_remote_data()
+    remote = deb_remote_custom_data_factory(data)
+    patch_data = gen_verbose_remote_data()
+    deb_patch_remote(remote, patch_data)
+    patch_remote = deb_get_remote_by_href(remote.pulp_href)
+
+    for key, val in patch_data.items():
         if key != "username" and key != "password":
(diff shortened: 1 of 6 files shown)


hstct:
Since this is the only `pytest` conversion PR that is still open, I also moved and renamed some fixtures that are also used by other tests. We will need to do this at some point and it gets annoying with the merge conflicts when you rebase a PR. So now would be a good time to include some clean up.

pulp_deb/tests/functional/conftest.py
```diff
@@ -99,7 +155,6 @@ def _deb_sync_repository(remote, repo):
 @pytest.fixture(scope="session")
 def signing_script_filename(signing_gpg_homedir_path):
```
mdellweg, line 156: This name sounds a little too generic too.
Or am i mistaken and it is meant to overshadow another fixture?

pulp_deb/tests/functional/conftest.py
```diff
@@ -69,17 +87,55 @@ def _deb_gen_repository():
...
+    """Fixture that returns the deb remote of a given pulp_href."""
+
+    def _deb_get_remote_by_href(href):
+        return apt_remote_api.read(href)
+
+    return _deb_get_remote_by_href
+
+
+@pytest.fixture
+def deb_is_latest_repository_version():
```
mdellweg, line 119: This may be overdoing the fixture thing. What you provide here is almost a pure function. You can provide this as a utility or even have the check inline everywhere...
`assert repo.latest_version_href.endswith(f"/{version}/")`

pulp_deb/tests/functional/conftest.py
hstct: No, I agree it sounds a bit too generic.

hstct:
I noticed that I missed to include two previous conversions. They needed a bit more refactoring so I ended up splitting the commits just in case it should not be included to this PR (which already has become a bit more than it sought out to be). Also added a bunch of `@pytest.mark.parallel` hopefully speeding some things up.

hstct:
I was also considering renaming all the `apt_remote_api`, `apt_repository_api` fixtures to start  with `deb_` instead for a more consistent naming scheme. I don't really remember why I started naming them `apt`.

mdellweg:
> I was also considering renaming all the `apt_remote_api`, `apt_repository_api` fixtures to start with `deb_` instead for a more consistent naming scheme. I don't really remember why I started naming them `apt`.

Yeah, that's hard. I consider it to be a Repository containing debs, and the remote to talk to an apt repository format (There are distros using yum to fetch deb-packages, i think.) So the publication should again be an AptPublication. Not sure i really thought it through in the beginning.

pulp_deb/tests/functional/api/test_crud_remotes.py
```diff
@@ -5,112 +5,158 @@
...
+    for key, val in data.items():
         if key != "username" and key != "password":
-            assert remote_repository.to_dict()[key] == val
+            assert remote.to_dict()[key] == val
 
 
+@pytest.mark.parallel
 def test_create_remote_repository_with_same_name(
-    apt_remote_api, gen_object_with_cleanup, remote_repository, gen_verbose_remote_data
+    deb_gen_remote_custom_data,
```
bmbouter, line 26: You can name these whatever you want, but for clarity we've tried to standardize on "factory pattern fixtures" ending in `_factory` you can see an example of this naming convention [here](https://github.com/pulp/pulpcore/blob/main/pulpcore/tests/functional/__init__.py#L221-L225).

pulp_deb/tests/functional/api/test_crud_remotes.py
```diff
@@ -5,112 +5,158 @@
...
-def test_create_remote_repository(verbose_remote_data, remote_repository):
-    """Test creation of the remote repository.
-    NOTE: The repository is actually created in this files fixture and this
-          test only confirms its creation.
-    """
-    for key, val in verbose_remote_data.items():
+@pytest.mark.parallel
+def test_create_remote_repository(deb_gen_remote_custom_data, deb_gen_verbose_remote_data):
+    """Test creation of the remote."""
+    data = deb_gen_verbose_remote_data()
```
bmbouter, line 16: We've debated some on to what extend everything should be a fixture, and I'll share the conclusion here. If it's a function that isn't providing server side "arrangement" or server side "cleanup" it likely should just be a utility method that uses a standard import instead of pytest "dependency injection" style. This is very much a stylistic thing and do what makes the most sense to you.

pulp_deb/tests/functional/api/test_crud_remotes.py
```diff
@@ -5,112 +5,158 @@
...
+    remote = deb_gen_remote_custom_data(data)
```
bmbouter, line 17: Arguably your fixture could call the utility function directly and you could have slightly different versions of the fixture to give you the various "verbosity" that you're looking for. Also the way you have it here is just fine too!

pulp_deb/tests/functional/api/test_crud_remotes.py
```diff
@@ -122,8 +168,8 @@ def test_remote_repository_download_policies(
        Verify an `ApiException` is given for the invalid policy as well
        as the policy remaining unchanged.
 
-    For more information on the remote policies, see the Pulp3
-    API on an installed server:
+    For more information on the remote policies, see the Pulp 3 API
+    on an installe server:
```
bmbouter, line 172:
```suggestion
    on an installed server:
```

pulp_deb/tests/functional/api/test_crud_remotes.py
```diff
@@ -132,41 +178,33 @@ def test_remote_repository_download_policies(
...
-    assert repo.policy == "immediate"
+    assert remote.policy == "immediate"
 
+    policies = DOWNLOAD_POLICIES
     changed_policy = choice([item for item in policies if item != "immediate"])
-    response = apt_remote_api.partial_update(repo.pulp_href, {"policy": changed_policy})
-    monitor_task(response.task)
-    repo = apt_remote_api.read(repo.pulp_href)
-    assert repo.policy == changed_policy
+    deb_patch_remote(remote, {"policy": changed_policy})
```
bmbouter, line 189: This is interesting this is a fixture that seems to use the bindings in a specific way. That's cool we haven't done something like this yet. :+1:

pulp_deb/tests/functional/api/test_download_content.py
```diff
@@ -51,103 +30,76 @@ def test_download_content(
...
-       content unit from Pulp, and verify that the content unit has the
-       same checksum when fetched directly from Pulp=Fixtures.
-
     Both versions of a repository (normal and verbatim) will be tested in
-    in this case.
+    this case.
 
     This test targets the following issues:
 
     * `Pulp #2895 <https://pulp.plan.io/issues/2895>`_
```
bmbouter, line 38: We stopped doing this "issue referencing" altogether in our docstrings. They just seemed to clutter thing and get out of date as tests got various updates over time.

pulp_deb/tests/functional/api/test_download_content.py
```diff
@@ -51,103 +30,76 @@ def test_download_content(
...
-        monitor_task(sync_response.task)
-        return apt_repository_api.read(repo.pulp_href)
-
-    return _gen_repository_with_synced_remote
+    """Fixture that does the following:
 
+    1. Create, populate, publish, and distribute a (verbatim) repository.
+    2. Select a random content unit in the distribution. Download that
+       content unit from Pulp, and verify that the content unit has the
+       same checksum when fetched directly from Pulp-Fixtures.
```
bmbouter, line 60: We started eliminating these step-by-step docstrings and instead putting one-line comments in the code below. What we found was that having the one line closer to the code doing it caused it to stay up to date more over time. You can see examples of this style [here](https://github.com/pulp/pulp_file/pull/769/files#diff-c7aed8492114ffb111e7a0b46bb0659312d12f8bf49be370172b6d3bdc8a30ecR36).

pulp_deb/tests/functional/api/test_download_policies.py
```diff
@@ -1,269 +1,181 @@
...
+        assert repo.latest_version_href.endswith("/1/")
 
-        repo = repo_api.create(gen_repo())
-        self.addCleanup(repo_api.delete, repo.pulp_href)
+        content = apt_package_api.list()
 
-        body = gen_deb_remote(policy=policy)
-        remote = remote_api.create(body)
-        self.addCleanup(remote_api.delete, remote.pulp_href)
+        assert content.count == DEB_FIXTURE_PACKAGE_COUNT
```
bmbouter, line 133: Usually all the asserts occur outside of the fixtures themselves allowing the fixtures to provide only "arrangement" and "cleanup". In this case you can use `yield` to allow for your test to receive control for assertion purposes and then resume your fixture code after the test assertions are done.

pulp_deb/tests/functional/api/test_publish.py
```diff
@@ -13,19 +18,7 @@
...
-        {"structured": True},
-        {"simple": True, "structured": True},
-        {"simple": True, "structured": True, "signing_service": apt_signing_service.pulp_href},
-        {},
-    ]
-    return params
-
-
+@pytest.mark.parallel
 @pytest.mark.parametrize(
```
bmbouter, line 22: :heart_eyes: nice parametrize usage.

pulp_deb/tests/functional/conftest.py
```diff
@@ -57,9 +59,38 @@ def apt_distribution_api(apt_client):
     return DistributionsAptApi(apt_client)
 
 
+@pytest.fixture
```
bmbouter, line 62: We try to make fixtures "session" level whenever possible. For the entirely reusable objects like bindings they are good candidates for that.

pulp_deb/tests/functional/conftest.py
```diff
@@ -69,19 +100,98 @@ def _deb_gen_repository():
...
+
+    def _deb_delete_remote(remote):
+        response = apt_remote_api.delete(remote.pulp_href)
+        return monitor_task(response.task)
+
+    return _deb_delete_remote
+
+
+@pytest.fixture
+def deb_patch_remote(apt_remote_api):
```
bmbouter, line 174: @dkliban @mdellweg this is a neat fixture style we should consider doing too (where it combines with `monitor_task`).

pulp_deb/tests/functional/conftest.py
```diff
@@ -69,19 +100,98 @@ def _deb_gen_repository():
...
+    """Fixture that returns the deb remotes of a given name."""
+
+    def _deb_get_remotes_by_name(name):
+        return apt_remote_api.list(name=name)
+
+    return _deb_get_remotes_by_name
+
+
+@pytest.fixture
+def deb_delete_remote(apt_remote_api):
```
bmbouter, line 163: Typically our code never has a "delete" fixture and instead we make heavy use of https://github.com/pulp/pulp-smash/blob/main/pulp_smash/pulp3/pytest_plugin/__init__.py#L494 which provides creates an object and auto-deletes it (in reverse order) after the test code completes. It avoids you having to provide this fixture, and call it in all your tests (which further mixes cleanup and assertion code, the test code ideally should just be assertions).

mdellweg: Well, we _do_ deletions in case we want to test the deletion itself.

pulp_deb/tests/functional/conftest.py
mdellweg: Here, it's impossible. As the apt_client adds the test specific correlation-id to the mix.

pulp_deb/tests/functional/api/test_crud_remotes.py
hstct: Ah nice, yeah that's a reasonable distinction between pytest fixture and utility method. I like that.

pulp_deb/tests/functional/api/test_download_content.py
hstct: +1

I just copy pasted them from the old tests and I wasn't going to change them. The very long docstrings were a bit cumbersome for browsing through the files. Happy to remove them.

pulp_deb/tests/functional/api/test_download_policies.py
hstct: After having another look at this particular test I'm not even sure why I split it up in different fixtures in the first place. But still I'll keep the note regarding assertions inside fixture in mind.

quba42 (approved):
Thanks to all three of you for this great work and discussion.

After a final presentation I will now merge this work.

Issue #599, opened 2022-08-02
ReleaseComponent content with different codename or suite but equal distribution collides


quba42:
# Workarounds for `DuplicateDistributionException`

If you are getting an error message as follows:
```
Cannot create the new repository version, since it contains multiple Release content with the same distribution '<distribution>'.
This known issue is tracked here: https://github.com/pulp/pulp_deb/issues/599
You can check the issue for known workarounds. Please also consider posting what you did before getting this error, to help us to fix the underlying problem more quickly.
```

Then you have hit our error (`DuplicateDistributionException`) for this issue.

- If you have hit it while syncing, you might be able to work around the problem by using the sync option `mirror=True` on your next sync attempt.
- Alternatively, you can sync the affected remote into a new empty repository, since this error cannot occur on a first time sync into an empty Pulp repository.

Please also consider posting below, to describe exactly what you did, before getting the above error.

# Problem Description

We noticed this while working on the advanced copy mode, which currently works as follows: If a user copies packages from one repo version to another, the advanced copy will find any PackageReleaseComponents that reference those packages, as well as any associated Components, and Releases, and move them along with the packages to preserve structure. However, this can result in multiple Releases with the same distribution (differing only in their codename or suite) within the same repository version. This will collide during publication since the APT publisher will try to create the path to the duplicate distribution twice.

There are several conceivable ways to address this issue:

### The clean design

* Structure content should contain exclusively structure information. Therefore we remove the codename and suite fields from the `Release` model (and probably rename that model `Distribution` while we are at it). Collisions are now no longer possible and structure content is equally valid (to describe structure), regardless what repo or repo version it is in.

**Pros:** This is clearly the conceptually cleanest possible design.

**Cons:** This would require a massive DB migration to merge existing colliding structured content (and then to regenerate all the associated ReleaseComponents and PackageReleaseComponents for all existing repo versions!

### Embracing our existing (bad) design

* Structure content only makes sense within some given RepositoryVersion. It can never be moved anywhere else. The advanced copy needs to be redesigned to create new structured content in the target version, rather than taking along structured content from the source version.

**Pros:** No massive DB migration.

**Cons:** We will remain limited by our early design choices for ever, whatever the consequences.

### For completeness sake (the complete mess)

The APT publisher needs to learn to accept that there can be multiple distributions in a single repo version and just publish them anyway.

[...]

PR #604, opened 2022-08-04
Do not allow duplicate distributions in a single repo version

### LINKED ISSUE #603: Do not allow multiple identical distributions in a single repo version
**Describe the bug**
This is essentially https://github.com/pulp/pulp_deb/issues/599, but under a new issue, for providing an initial fix (while retaining the old ticket for a potential long term fix).

**To Reproduce**
Either manually, or using the advanced copy API add multiple `Release` content units to a repository, that differ only in the `suite` or `codename`, but not in the `distribution`.

Publishing such a repo version will fail, because the duplicate distributions will collide.

**Expected behavior**
As an initial fix, we should use `repo_key_fields = ('distribution', )` to prevent the creation of broken repo versions.
For the manual creation workflow this is all that is needed. (pulp_deb will tell users "not to do that").

For the advanced copy workflow, this will still be frustrating, since advanced copy between certain repos simply will not work. However, it is still better to fail during the copy, then to pretend everything is peachy until the user tries to publish (fail early)! Any further steps to fix the advanced copy workflow should then be done here: https://github.com/pulp/pulp_deb/issues/599

quba42:
closes #603

ref #599

### DIFF
--- a/pulp_deb/app/models/repository.py
+++ b/pulp_deb/app/models/repository.py
@@ -64,5 +64,14 @@ class AptRepository(Repository):
                 finalize.
 
         """
+        from pulp_deb.app.tasks.exceptions import DuplicateDistributionException
+
         remove_duplicates(new_version)
         validate_repo_version(new_version)
+        releases = new_version.get_content(Release.objects.all())
+        distributions = []
+        for release in releases:
+            distribution = release.distribution
+            if distribution in distributions:
+                raise DuplicateDistributionException(distribution)
+            distributions.append(distribution)
--- a/CHANGES/603.bugfix
+++ b/CHANGES/603.bugfix
@@ -1,0 +1,1 @@
+Added a better error message when users try to create a repository version containing duplicate APT distributions.
--- a/pulp_deb/app/tasks/exceptions.py
+++ b/pulp_deb/app/tasks/exceptions.py
@@ -1,0 +1,18 @@
+from gettext import gettext as _
+
+
+class DuplicateDistributionException(Exception):
+    """
+    Exception to signal, that we are creating a repository versions containing multiple Release
+    content with the same distribution.
+    """
+
+    def __init__(self, distribution, *args, **kwargs):
+        message = (
+            "Cannot create the new repository version, since it contains multiple Release content "
+            "with the same distribution '{}'.\n"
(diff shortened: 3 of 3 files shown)


quba42:
Hmm... I thought this was going to raise an error, but it really just silently removes the duplicates. I am not sure if that is a good solution, since it can lead to some really counter intuitive results...

pulp_deb/app/models/repository.py
```diff
@@ -66,3 +68,10 @@ def finalize_new_version(self, new_version):
         """
         remove_duplicates(new_version)
         validate_repo_version(new_version)
+        releases = new_version.get_content(Release.objects.all())
+        distributions = []
+        for release in releases:
+            distribution = release.distribution
+            if distribution in distributions:
+                raise DuplicateDistributionException(distribution)
+            distributions.append(distribution)
```
mdellweg, line 77: Don't you want to check that no two distributions with the same name are added in that repository version, but still be able to replace an existing one?

quba42: Yes.

My understanding was that `releases = new_version.get_content(Release.objects.all())` will get me all the releases that are in the new repository version just before it is saved (but not any releases that might be in old repository versions, but not in the new one being created). If that understanding is wrong then this code won't work, I have not tested it yet...

quba42: @mdellweg Ok, I think I understand your objection now: "What if somebody wants to re-sync some upstream repository where the suite has changed (as happens when a new Debian release goes stable for example)? Won't this change (in its current version) prevent that sync from happening (at least with policy 'additive')?" I think that could indeed be a problem. However if Pulp would just remove the older Release in that situation, I don't think that would be a good result either, since I think that would leave dangeling `ReleaseComponents` and `PackageReleaseCompoenents` in the repo version, that are no longer attached to any `Release` that is also in the repo version.

What I am doing here just makes the design flaw painfully visible, it does not actually fix anything, leaving it to users how they want to work around the situation. Which is why I think we ultimately need https://github.com/pulp/pulp_deb/issues/599 for a full fix.

quba42:
After some internal discussion, we have decided to add this exception (with a error message pointing people at issue #599) for now.
Hopefully, we can replace this with a better fix soon.

PR #642, opened 2022-09-16
Fix sync not including packages only found in component `all`

### LINKED ISSUE #612: Synching repositories doesn't include `binary-all` packages
**Version**
pulpcore: 3.20.0
pulp_deb: 2.19.1

**Describe the bug**
After synching a debian repository the `Packages` file in `binary-all` is empty. This results to those packages not being there at all.

**To Reproduce**
```
pulp deb remote create --name bullseye-updates --url https://deb.debian.org/debian/ --policy immediate --distribution bullseye-updates
pulp deb repository create --name bullseye-updates
pulp deb repository sync --name bullseye-updates --remote bullseye-updates
APT_PUBLICATION_HREF=$(pulp deb publication create --repository=bullseye-updates --simple=True --structured=True | jq -r '.pulp_href')
pulp deb distribution create --name=bullseye-updates --base-path=bullseye-updates --publication=${APT_PUBLICATION_HREF}
```

Check if the file in `/pulp/content/bullseye-updates/dists/bullseye-updates/main/binary-all/Packages` is not empty.

**Expected behavior**
The `Packages` file in `binary-all` should match the one from the upstream repository and all the packages are also available.

**Additional context**
It's highly likely that this is caused by the `hybrid_format` checks in the `synchronizing` tasks. [here](https://github.com/pulp/pulp_deb/blob/main/pulp_deb/app/tasks/synchronizing.py#L585-L597) and [here](https://github.com/pulp/pulp_deb/blob/main/pulp_deb/app/tasks/synchronizing.py#L644-L645).

hstct:
closes #612

### DIFF
--- a/CHANGES/612.bugfix
+++ b/CHANGES/612.bugfix
@@ -1,0 +1,1 @@
+Fixed a bug where architecture "all" packages were missing when syncing Debian 11 style repositories.
--- a/pulp_deb/app/tasks/synchronizing.py
+++ b/pulp_deb/app/tasks/synchronizing.py
@@ -697,7 +697,6 @@ class DebFirstStage(Stage):
         # first. That way, we can recover the special case, where a partial mirror does not mirror
         # this index inspite of indicating "hybrid format" in the mirrored metadata.
         if hybrid_format and "all" in architectures:
-            architectures.remove("all")
             try:
                 await self._handle_package_index(
                     release_file=release_file,
@@ -719,6 +718,14 @@ class DebFirstStage(Stage):
                 release_file.architectures = " ".join(
                     [x for x in release_file.architectures.split() if x != "all"]
                 )
+
+        # Putting this here because it fixes an issue where debian packages with the parameter
+        # architectures='all' are missing after sync/publish. It is not really clear why and
+        # needs investigation once there are tests for this issue. Best guess it hasis something do
+        # with the asynchronous handling of the tasks and removing something from a dict without
+        # a copy.
+        if hybrid_format and "all" in architectures:
+            architectures.remove("all")
 
         pending_tasks = []
         # Handle package indices


pulp_deb/app/tasks/synchronizing.py
```diff
@@ -720,6 +719,9 @@ async def _handle_component(
                     [x for x in release_file.architectures.split() if x != "all"]
                 )
 
+        if hybrid_format and "all" in architectures:
+            architectures.remove("all")
```
hstct, line 723: Moving this because it probably interferes with the async handling of the packages. But seems to work as a workaround

hstct:
Ideally we include a test for this but the current fixture repository with the test packages are not using the newer format.

pulp_deb/app/tasks/synchronizing.py
hstct: To clarify what I mean, if the architecture "all" is removed before we try to force the `_handle_package_index` with `architectures=all` it will not successfully run the method. I tried moving it in the same if-statement after the try/catch block but this results or move it to a finally block. Both will break the whole syncing process.

Moving it outside of the if-statement will work. I'm however not really sure why and my suspicion is that it has something to with the async nature of the task?

quba42:
Closing, since we decided not to go ahead with this until we find a clear reproducer.

CHANGES/612.bugfix
```diff
@@ -0,0 +1 @@
+Fixed synching issue that did not include packages which are only found in the "all" component.
```
quba42, line 1: I think for upstream users the changelog entry might be easier to understand if reformulated to something like:
```suggestion
Fixed a bug where architecture "all" packages were missing when syncing Debian 11 style repos.
```

quba42 (approved):
As discussed we have verified this fixes a very real critical problem, even if we are not 100% sure why. We should merge this and perform bugfix releases ASAP.

Feel free to take or leave (or amend) my changelog suggestion.

PR #644, opened 2022-09-25

pulpbot:
[noissue]


template_config.yml
```diff
@@ -14,6 +14,7 @@ check_openapi_schema: true
 check_stray_pulpcore_imports: true
 ci_env: {}
 ci_trigger: '{pull_request: {branches: [''*'']}}'
+ci_update_branches: []
```
mdellweg, line 17: we should come up with a list here.

quba42: I guess this replaces `keep_ci_update_since_branch: '2.16'`, the equivalent list would be `['2.16', '2.17', '2.18', '2.19']`.

Personally, I am not actually interested in `2.17`, so if we are Ok with maintaining some branches but not all intervening branches, we could drop that. Alternatively, we can wait until latest Katello uses `2.18` and then drop both `2.16` and `2.17` from the list. I think I prefer this latter option.

template_config.yml
```diff
@@ -28,7 +29,7 @@ flake8: true
 flake8_ignore: []
 github_org: pulp
 issue_tracker: github
-keep_ci_update_since_branch: '2.16'
+kanban: true
```
quba42, line 32: I am not sure what this does, but since we do not use any Kanban boards in `pulp_deb` development, I am guessing we don't need it?

quba42: I was made aware of this: https://github.com/orgs/pulp/projects/8/views/1

Looks like there is one board for all of pulp. I am neutral on whether pulp_deb issues should appear on this board. I feel like the community should decide.

mdellweg: I would opt for `false`.

bmbouter: It's a shared tool so `pulp_deb` certainly could be, but if it's not a "yes I want pulp_deb to be on that" then I think it should be set to `false`.

template_config.yml
mdellweg: Dropping branches here does not mean we can never release on them any more. It just may be a deferred amount of work to do so. We started internally to call the ones listed here "hot release branches".

template_config.yml
quba42: I have set it to `false` for now.

template_config.yml
quba42: I have set it to `['2.16', '2.17', '2.18', '2.19']` for now, which should be equivalent to what we had before.

I will probably drop `2.16` and `2.17` from the list after the next Y-release.

mdellweg (approved):

PR #657, opened 2022-10-14
Fix typo

daviddavis:
[noissue]


quba42:
I am going to go ahead and merge this without asking for a rebase so the tests can turn green. :wink: 

I really don't see how this could possibly go wrong.

PR #667, opened 2022-10-23
Unpin azure-storage-blob

daviddavis:
Testing out the new version of azure-storage-blob which has the fix for
empty files.

https://pypi.org/project/azure-storage-blob/12.14.1/
https://github.com/Azure/azure-sdk-for-python/releases/tag/azure-storage-blob_12.14.1


daviddavis:
If this CI passes, I'll open a PR against plugin_template to remove this code as well.

quba42:
Now that this has turned green, should we merge this PR, or simply close and wait for the "CI Update" job to do its thing?

daviddavis:
Let’s close and wait for the CI update.

PR #675, opened 2022-11-01
Fix packages missing from Packages index

### LINKED ISSUE #674: Missing packages from Packages file
**Version**
pulpcore 3.22.0.dev (2054f5022c24d0a4bfc4515d4967d64fff2b109d)
pulp_deb 2.20.0.dev (f9592776c9ace4d727892b8aa3e74a1eef573208)

**Describe the bug**
It appears that packages go missing from Packages files if they appear in more than one release.

**To Reproduce**
```
pulp deb remote create --name sonic --url "http://packages.microsoft.com/repos/sonic/" --distribution "jessie trusty"
pulp deb repository create --name sonic --remote sonic
pulp deb distribution create --name sonic --base-path sonic --repository sonic
pulp deb repository sync --name sonic
pulp deb publication create --repository sonic --structured
http :24816/pulp/content/sonic/dists/jessie/main/binary-amd64/Packages | grep s6000
http :24816/pulp/content/sonic/dists/trusty/main/binary-amd64/Packages | grep s6000
```

**Expected behavior**
The package should show up in both releases, like this:

```
$ http http://packages.microsoft.com/repos/sonic/dists/jessie/main/binary-amd64/Packages | grep s6000
Package: platform-modules-s6000
Filename: pool/main/p/platform-modules-s6000/platform-modules-s6000_1.0+0~20160606043549.24~1.gbpbc91df_amd64.deb
$ http http://packages.microsoft.com/repos/sonic/dists/jessie/main/binary-amd64/Packages | grep s6000
Package: platform-modules-s6000
Filename: pool/main/p/platform-modules-s6000/platform-modules-s6000_1.0+0~20160606043549.24~1.gbpbc91df_amd64.deb
```

However, only one of the two calls with httpie will actually output anything.

The problem is that when [adding the package to the component](https://github.com/pulp/pulp_deb/blob/f9592776c9ace4d727892b8aa3e74a1eef573208/pulp_deb/app/tasks/publishing.py#L173-L178), two things happen:
1. [the published artifact gets created](https://github.com/pulp/pulp_deb/blob/f9592776c9ace4d727892b8aa3e74a1eef573208/pulp_deb/app/tasks/publishing.py#L206-L211)
2. [the package gets added to the package index file](https://github.com/pulp/pulp_deb/blob/f9592776c9ace4d727892b8aa3e74a1eef573208/pulp_deb/app/tasks/publishing.py#L212-L216)

If 1 fails, then two never happens. So a package only gets added to one Packages index file.

daviddavis:
fixes #674

### DIFF
--- a/CHANGES/674.bugfix
+++ b/CHANGES/674.bugfix
@@ -1,0 +1,2 @@
+Fixed a bug where packages were only showing up in one Packages index file if they belonged to two
+or more releases.
--- a/pulp_deb/app/tasks/publishing.py
+++ b/pulp_deb/app/tasks/publishing.py
@@ -1,5 +1,6 @@
 import os
 import shutil
+from contextlib import suppress
 
 from datetime import datetime, timezone
 from debian import deb822
@@ -170,12 +171,9 @@ def publish(repository_version_pk, simple=False, structured=False, signing_servi
                         pk__in=repo_version.content.order_by("-pulp_created"),
                         release_component__in=components,
                     ):
-                        try:
-                            release_helper.components[prc.release_component.component].add_package(
-                                prc.package
-                            )
-                        except IntegrityError:
-                            continue
+                        release_helper.components[prc.release_component.component].add_package(
+                            prc.package
+                        )
                     release_helper.finish()
 
     log.info(_("Publication: {publication} created").format(publication=publication.pk))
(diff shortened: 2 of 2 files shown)


quba42 (approved):
I finally looked at this seriously with @hstct 

From code review this looks correct to me, and it appears to be fixing a critical bug.

We want to try and run some more internal tests before we merge, but with the goal of merging today or tomorrow, and then doing immediate bugfix releases to get this out!

quba42:
@hstct We also need to decide which release branches we want to backport this to. Assuming they are all affected, I would say `2.20`, `2.19`, `2.18`. Can you have a look if those look like they are affected?

quba42:
I suspect this is the same issue: https://projects.theforeman.org/issues/35362

quba42:
It looks like this is not the only bug we have that causes missing packages (in certain situations).

We will merge, but may hold off with bugfix releases until we have analyzed whether to add additional bugfixes...

PR #676, opened 2022-11-02
Fix download_policies test for the CI

hstct:
[noissue]


hstct:
The issue is being fixed by not running the test in parallel. The confusing thing is that it [used to work](https://github.com/pulp/pulp_deb/actions/runs/3318052322) until [7 days ago](https://github.com/pulp/pulp_deb/actions/runs/3326003406) without anything changing in between (at least to my knowledge)

PR #678, opened 2022-11-16
Remove mirror default of True

daviddavis:
The `mirror` field defaults to False in the RepositorySyncURLSerializer:

https://github.com/pulp/pulpcore/blob/ab7f4b7c7e1939ae605a8d7cf996c6498c83ddd7/pulpcore/app/serializers/repository.py#L336

So it should always be set and it should be False by default. This line however gives the impression that mirror is True by default so update it so it no longer has a default. This shouldn't impact users.

[noissue]


quba42 (approved):
Especially since we don't set default values for other values we get from serializers in the same way (e.g.: `optimize` one line down), this change makes a lot of sense as a code readability improvement!

hstct (approved):

PR #695, opened 2022-12-05
Prevent sync from breaking if a custom field is blank

### LINKED ISSUE #694: Cannot sync repo with blank custom field
**Version**
main (fc96426ee6fa0e6445bf678ffd773b5e2dfbb533)

**Describe the bug**
If there's a blank custom field in the Packages index, sync fails.

**To Reproduce**
```
pulp deb remote create --name aplinux --url "https://packages.microsoft.com/repos/aplinux-prod/" --distribution xenial --architecture amd64
pulp deb repository create --name aplinux --remote aplinux
pulp deb repository sync --name aplinux
```

Output:

```
Started background task /pulp/api/v3/tasks/1b8ffdc3-fbb1-43d2-bb75-902c2e4a3317/
.Error: Task /pulp/api/v3/tasks/1b8ffdc3-fbb1-43d2-bb75-902c2e4a3317/ failed: '{'custom_fields': {'Build-Depends': [ErrorDetail(string='This field may not be blank.', code='blank')]}}'
```

**Expected behavior**
I guess I would expect sync to work but I don't know if this behavior is intentional?

daviddavis:
fixes #694

### DIFF
--- a/CHANGES/694.bugfix
+++ b/CHANGES/694.bugfix
@@ -1,0 +1,1 @@
+Fix bug where sync failed if "Build-Depends" field in Package index was blank.
--- a/pulp_deb/app/serializers/content_serializers.py
+++ b/pulp_deb/app/serializers/content_serializers.py
@@ -255,7 +255,7 @@ class BasePackage822Serializer(SingleArtifactContentSerializer):
     pre_depends = CharField(required=False)
     provides = CharField(required=False)
     replaces = CharField(required=False)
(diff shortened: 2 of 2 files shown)


daviddavis:
I now realize that this solution is perhaps not a good one since it also permits packages with blank custom_fields to be uploaded as well which is probably not ideal.

daviddavis:
I'm going to close out this PR. I don't think this change is desired.

PR #698, opened 2022-12-07
Temporarly disabling broken test to fix the CI

hstct:
[noissue]

### DIFF
--- a/pulp_deb/tests/functional/api/test_download_policies.py
+++ b/pulp_deb/tests/functional/api/test_download_policies.py
@@ -50,6 +50,8 @@ def test_download_policy(
     assert content.count == DEB_FIXTURE_PACKAGE_COUNT
 
 
+# TODO: Should think of a better approach to test this case.
+@pytest.mark.skip(reason="Currently breaking the CI")
 @pytest.mark.parametrize("policy", ["on_demand", "streamed"])
 def test_lazy_sync_immediate_download_test(
     artifacts_api_client,


mdellweg (approved):
I may not have made the best choice moving the cleanup fixtures to model scope...
Do you think these tests can be rewritten without relying on the global count result of the artifacts endpoint?

quba42:
> Do you think these tests can be rewritten without relying on the global count result of the artifacts endpoint?

Once we move fixtures into the pulp_deb repo, one might test for the presence of specific artifacts by checksum. Not sure if that is a good idea though. (Not  sure how constant our fixture artifact checksums will be going forward).

quba42 (approved):

mdellweg:
> > Do you think these tests can be rewritten without relying on the global count result of the artifacts endpoint?
> 
> Not on the top of my head otherwise I would have implemented it. Just disabling it for now with a note as I don't have the time to fix it right now and having a working CI seems more important. :)

Can you try if setting the cleanup fixtures to "class" scope here helps?
https://github.com/pulp/pulpcore/blob/main/pulpcore/tests/functional/__init__.py#L505

hstct:
> > > Do you think these tests can be rewritten without relying on the global count result of the artifacts endpoint?
> > 
> > 
> > Not on the top of my head otherwise I would have implemented it. Just disabling it for now with a note as I don't have the time to fix it right now and having a working CI seems more important. :)
> 
> Can you try if setting the cleanup fixtures to "class" scope here helps? https://github.com/pulp/pulpcore/blob/main/pulpcore/tests/functional/__init__.py#L505

The test only fails on github. Locally it runs fine. 

To test it I would need to make a PR in pulpcore and use it here, right?

hstct:
Yea, I'm not able to reproduce this issue locally. Running all the tests at once only results fails an unrelated test:

```
==================================================================================== test session starts ====================================================================================
platform linux -- Python 3.8.13, pytest-7.2.0, pluggy-1.0.0
rootdir: /var/lib/pulp
plugins: xdist-3.1.0, custom-exit-code-0.3.0, pulp-smash-1!0.13.0, pulpcore-3.22.0.dev0
collected 68 items

../../var/lib/pulp/api/test_crud_content_unit.py ..........................                                                                                                           [ 38%]
../../var/lib/pulp/api/test_crud_remotes.py .........                                                                                                                                 [ 51%]
../../var/lib/pulp/api/test_download_content.py ..                                                                                                                                    [ 54%]
../../var/lib/pulp/api/test_download_policies.py ....                                                                                                                                 [ 60%]
../../var/lib/pulp/api/test_publish.py .....                                                                                                                                          [ 67%]
../../var/lib/pulp/api/test_publish_complex_dists.py FF                                                                                                                               [ 70%]
../../var/lib/pulp/api/test_publish_empty_repository.py .                                                                                                                             [ 72%]
../../var/lib/pulp/api/test_publish_flat_repo_format.py ......                                                                                                                        [ 80%]
../../var/lib/pulp/api/test_sync.py .......                                                                                                                                           [ 91%]
../../var/lib/pulp/api/test_sync_missing_architecture.py .                                                                                                                            [ 92%]
../../var/lib/pulp/api/test_sync_optimize.py .....                                                                                                                                    [100%]
```

Running only `oci-env test -p pulp_deb functional -k test_download_policies` works without failure.

mdellweg:
OK, i think we should go forward merging this to unblock development.

mdellweg:
> Not really sure why the deprecations stage is failing?

That's the latest plugin_template change not yet being applied:

 https://github.com/pulp/pulp_deb/actions/runs/3674104931/jobs/6213363175#step:3:7

hstct:
> That's the latest plugin_template change not yet being applied:

Okay so I will merge this and the CI changes from main and it should work again.

PR #700, opened 2022-12-11

pulpbot:
[noissue]


requirements.txt
```diff
@@ -2,5 +2,5 @@
 # Example transitive requirements: asgiref, asyncio, aiohttp
 pulpcore>=3.21.0.dev,<3.25
 python-debian>=0.1.44,<0.2.0
-python-gnupg>=0.4.9,<0.6
+python-gnupg>=0.5,<0.6
```
quba42, line 5: I guess it makes sense for us to raise our lower bound whenever pulpcore does, (because we are clearly not getting anything older than what pulpcore needs anyway!) And I guess that is what that test is telling us, so we should go ahead and merge this.

quba42 (approved):

PR #701, opened 2022-12-11
Update CI files from 2.19 branch

pulpbot:
[noissue]


hstct:
The [signing service issue](https://github.com/pulp/pulp_deb/actions/runs/3667062695/jobs/6199301496#step:11:1767) is fixed by using the pytest conversion.

I cherry-picked all relevant pytest commits for this and those include everything but the optimize sync test

Alternatively I can limit it to only the signing service changes but I don't see a downside to just backport some of the pytest conversions. What do you think?

(This needs to be done with the 2.18 branch as well)

quba42:
> Alternatively I can limit it to only the signing service changes but I don't see a downside to just backport some of the pytest conversions. What do you think?

No point in limiting, IMHO the more similar the CI and tests are for all our currently supported release branches, the less problems we will have releasing (at least we are less likely to have different problems on each branch...).

quba42 (approved):

PR #703, opened 2022-12-11
Update CI files from 2.18 branch

pulpbot:
[noissue]


hstct:
Disabling the signing service test. It breaks with the current implementation of the CI and the 2.18 branch uses a pulpcore version that's missing key pytest fixtures needed for the pytest signing service implementation.

PR #704, opened 2022-12-15
Add local test data repositories

### LINKED ISSUE #395: Refactor the debian test fixtures building
Author: @quba42 (quba42)

Redmine Issue: 7442, https://pulp.plan.io/issues/7442

---

The idea is to use best Makefile practices.

Currently, we are building all packages multiple times for each repository.

Instead, packages should have there own make targets which are dependent targets of the repositories that need them.

Ideally all built repositories could share a common pool folder (either directly or using symlinks), so we do not build an unnecessarily large fixtures container.
This might be prohibitively hard to achieve with the current repo building tooling.

hstct:
closes #395


hstct:
This is still WIP.

The approach is still open for discussion. Currently using precompiled `.deb` packages from the `pulp-fixtures` repository to make it as close to the previous fixtures as possible (to avoid refactoring of the tests). It uses symlinks to avoid duplicates.

There is also the possibility to use symlinks for the `Release`, `Release.gpg`, `InRelease`, `Package` and `Package.gz` files. It just will be very messy to keep track of all the files that are different.

Another approach would be only using the `.deb` packages and let pytest generate the relevant repository files itself. This will require more effort as debian repositories need so many files to be generated.

The overal goals imo should be:

- Local test data ready to use for CI containers and dev environments
- Adding new packages and repositories for tests should be way easier than adding them in `pulp-fixtures` and wait for it to be build.

hstct:
I updated the PR and it's now a proof of concept that works locally with pytest versions of the tests (e.g. `test_sync_optimize`).

I need to still need to add it to some remaining tests as well as cleanup.

Not really sure how to make the linter accept all the new file types. Especially ones without a type like `Packages`

I tried to include it to the `MANIFEST.in` file like this: `include pulp_deb/tests/functional/fixtures/**/*` but that didn't work out.

hstct:
Apparently python `setuptools` and `distutils` can't handle symlinks very well. If I symlink a folder the manifest check will complain, single specific files (e.g. `.deb` packages) work however if the filetype is included in the `MANIFEST.in`.

To circumvent the problem I want to create the symlinks for the `.deb` packages using at runtime using pytest. However it seems that [pulp doesn't have permission to create symlinks in the container](https://github.com/pulp/pulp_deb/actions/runs/3872194493/jobs/6600771907#step:11:2466). This happens locally using `oci-env` as well in the GitHub CI.

I removed symlinks for the distribution files (e.g. `Release` `Packages` etc.) and checked in all the files. This means there will be some duplicates for now, for the sake of less complexity. This can be improved in the future however, as those files are not huge it shouldn't be a big problem.

mdellweg:
How much added data is this for the published packages? Can you give an estimate?
Is there a way to only include the data for special builds (packages to be used in testing environments)?
Maybe it would even be beneficial to not duplicate the packages, but have actually different artifacts for the same package names in different fixture repos.

mdellweg:
For the sake of future us, can you include all necessary information needed to regenerate these fixtures?

hstct:
> Maybe it would even be beneficial to not duplicate the packages, but have actually different artifacts for the same package names in different fixture repos.

This is I think a good compromise. Doing this would make the implementation slightly simpler and we could use a direct copy of the repositories on fixtures.pulpproject.org.

I noticed that adding a new repository using the symlinked packages might not be the best idea because if one wants to add a new repository and lets it generate using cli tools you would also need to manually edit all the checksums of the packages and so on.

The intention of using the same packages was just to have less clutter and binary files. We discussed to use this PoC for now since it is already an improvement (using local data and one does not need to go to another repository to add new files) and we can build to improve it over time with more automation.

hstct:
> How much added data is this for the published packages? Can you give an estimate?

Not sure I understand the question. There is a about 4kb per package and the repository files are 596kb in total and around 140kb per repository.

hstct:
> For the sake of future us, can you include all necessary information needed to regenerate these fixtures?

Yes of course I will provide a small documentation on how to generate and add new data to the repository. I'm currently working on that.

quba42 (approved):
@hstct Feel free to merge as you see fit.

I am looking forward to starting to work with this change on every new branch.
A big step forward for the ease of adding new tests when compared to working with the pulp-fixtures repo!

hstct:
Added some (very) minimal instructions on how to add your own test data.

hstct:
cc @daviddavis @sdherr

PR #706, opened 2022-12-16
Add repo_key_fields constraint to PackageReleaseComponent

### LINKED ISSUE #705: PackageReleaseComponents are not being properly handled in the case of duplicates
**Version**
2.20.1

**Describe the bug**
It looks like when a duplicate package gets added to a release in a repository, Pulp properly removes any old duplicate packages (with the same NVA). However, it does not properly remove the PackageReleaseComponents. So if/when the package is removed from the repo, the old package will get published since its PackageReleaseComponent still exists.

**To Reproduce**
1. Upload two packages into Pulp with the same NVA (I'll call them A1 and A2)
2. Create a repo with a release
3. Add A1 to the repo release
4. Add A2 to the repo release
5. Remove A2 from the repo
6. Publish the repo and observe that A1 is in the publication

The problem occurs after step 4. There's only one package in the repo version (correct) but yet there are two PackageReleaseComponents:

```
>>> DebPackage.objects.filter(pk__in=RepositoryVersion.objects.order_by("pulp_created").last().content)
<BulkTouchQuerySet [<Package: azureauth_0.5.4-1_amd64>]>
>>> PackageReleaseComponent.objects.filter(pk__in=RepositoryVersion.objects.order_by("pulp_created").last().content).values_list("pk", "package__package", "package__version", "package__architecture")
<BulkTouchQuerySet [(UUID('aa1ebf24-5c7b-4ef2-baef-9033c5b3b0c4'), 'azureauth', '0.5.4-1', 'amd64'), (UUID('11eeb353-353a-4dd1-aacc-e7c823571207'), 'azureauth', '0.5.4-1', 'amd64')]>
```

**Expected behavior**
If pulp_deb automatically removes the old duplicate package, then I'd argue it should remove the PackageReleaseComponent too. In fact, I think adding a repo_key_fields constraint to PackageReleaseComponent should solve the problem:

```
repo_key_fields = ("release_component", "package__package", "package__version", "package__architecture")
```

daviddavis:
fixes #705


lubosmj:
I am not sure about the performance implications. Can this be evaluated by @pulp/debian?

daviddavis:
@lubosmj the repo_key_fields are only used in the [`remove_duplicates `](https://github.com/pulp/pulpcore/blob/fcb355c1f8fec6780524c010cf5bba7e1aa1d51b/pulpcore/plugin/repo_version_utils.py#L16) function which is called once during the repo version finalize step. What sort of performance problems were you imagining?

quba42:
So I fully understand what you are trying to achieve here, but before I hit merge, I would like to discuss the following possible problem (I am not sure if it can happen):

* Assume we have PackageReleaseComponent A pointing at Package A.
* We now add Package B with the same name, version, architecture as Package A, and a new PackageReleaseComponent B (pointing at Package B)

=> `remove_duplicates` should now boot Package A and PackageReleaseComponent A from the repo version being created.

My question is: Can we think of any set of circumstances where we end up with Package A, but PackageReleaseComponent B or vice versa, being booted from the new repo version? This would be a "consistent" final state as far as `remove_duplicates` is concerned (I think), though it would of course be very inconsistent from the point of view of the final repo version being created.

My gut feeling is that this should not be a problem, since both the duplicate Package, and the duplicate PackageReleaseComponent should kick out the older of each duplicate pair, but I would like to see others weigh in!

quba42:
Also, I seem to remember we have a pretty detailed test for duplicate package removal, I imagine this could be extended quite easily to include PackageReleaseComponents as well. If we wait for https://github.com/pulp/pulp_deb/pull/744 then adjusting the test might be as simple as using the new "structured upload" option to set up the test, testing both in the process.

Now that I think about it, we should maybe just adjust this as part of #744.

quba42:
> Yes, this is totally possible. Suppose you have Package A and PRC A in your repo but then you add PRC B to your repo (which would replace/remove PRC A). Then you'd end up with an inconsistent state. I'd say that this is more of a user error.

Yes, when manually building repos one content unit at a time, there is currently no way to prevent people from building inconsistent repos that way. I was wondering if it might be possible for the mechanisms provided by the plugin (sync and the structured upload from #744) to result in the inconsistent state, but I don't think it is. With those mechanisms some package A with its PRC A are always added to the repo together, so if they do kick out any duplicates they will always win out together, so things will always remain consistent.

And with that analysis I am happy to approve this change. It certainly makes sense for a PRC to follow the same definition of "duplicate" as a package itself.

quba42 (approved):

quba42:
Decided to go ahead and merge this, since it is such a obvious improvement over the status quo. In an ideal world I would like to see a test, but that will be easier to add in combination with https://github.com/pulp/pulp_deb/pull/744 which is stuck on a mysterious test failure.

PR #732, opened 2023-03-07
Remove kanban board workflows

lubosmj:
[noissue]


quba42 (commented):
Won't this come back in with plugin template? Should we at least open an issue with the plugin_template to add a `template_config.yml` parameter for disabling Kanban board permanently?

lubosmj:
The option to disable the kanban board workflows was shipped with this commit: https://github.com/pulp/pulp_deb/commit/4817d3a5eefa511788ccdf04ea5372108d633bea (Sep 28, 2022).

The workflow file itself was already present in the repository at that time: https://github.com/pulp/pulp_deb/commit/0412ee7b54488a100b0fc5ce722398215c236cf4 (Sep 13, 2022). We do not support automatic removal. Thus, it is necessary to remove the residue manually, as I did.

quba42 (approved):
@lubosmj Thanks for the clarification, then I think we can merge this.

PR #733, opened 2023-03-09
Improve "No valid Release file found" error message

Manisha15:


pulp_deb/app/tasks/synchronizing.py
```diff
@@ -364,16 +385,19 @@ async def run(self):
...
+                                gpg_verify_count += 1
                             else:
                                 log.warning(_("Verification of InRelease failed. Dropping it."))
                                 d_content.d_artifacts.remove(da_names.pop("InRelease"))
                         else:
                             release_file_artifact = da_names["InRelease"].artifact
                             release_file.relative_path = da_names["InRelease"].relative_path
 
-                    if not d_content.d_artifacts:
+                    if gpg_verify_count == 0:
```
quba42, line 396: I think we need a slightly modified condition here. Something like:
```
if not d_content.d_artifacts:
    if dropped_count > 0:
        raise NoValidGPG(url=self.remote.url)
    else:
        raise NoReleaseFile(url=self.remote.url)
```
where `dropped_count` counts the number of release files we dropped instead of counting the number of release files that successfully verified with `gpg_verify_count`. That way, if the user provides an incorrect URL/distribution, AND also provides a GPG key, they will get the error for incorrect URL/distribution, and not the error for incorrect GPG key. Does that make sense?

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -75,12 +75,31 @@ class NoReleaseFile(Exception):
...
         super().__init__(
-            "No valid Release file found for '{}'.".format(distribution), *args, **kwargs
+            "Could not find a Release file at '{}', try checking the 'url' and "
+            "'distributions' option on your remote".format(url),
+            *args,
+            **kwargs,
+        )
+
+
+class NoValidGPG(Exception):
```
quba42, line 90: Nitpick:
```suggestion
class NoValidSignatureForKey(Exception):
```
I feel like this is a better name, since we don't know if the problem is with the GPG key or with the signature.

quba42 (changes requested):
I think we need a small change to the logic, the rest LGTM.

pulp_deb/app/tasks/synchronizing.py
Manisha15: Changed the name.

CHANGES/399.feature
```diff
@@ -0,0 +1 @@
+Improve the pulp_deb "No valid Release file found" error message for gpg validation fail.
```
hstct, line 1: According to the issue this a bugfix rather than a feature. I'd personally would mark it as misc. But bugfix is fine as well.

hstct (changes requested):

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -367,13 +388,17 @@ async def run(self):
...
                             release_file_artifact = da_names["InRelease"].artifact
                             release_file.relative_path = da_names["InRelease"].relative_path
 
                     if not d_content.d_artifacts:
                         # No (proper) artifacts left -> distribution not found
-                        raise NoReleaseFile(distribution=release_file.distribution)
+                        if dropped_count > 0:
+                            raise NoValidSignatureForKey(url=self.remote.url)
+                        else:
+                            raise NoReleaseFile(url=self.remote.url)
```
quba42, line 401: I feel like it would be clearer for users if we used the URL to the affected release file folder here (instead of the URL to the repo root). So we would need to set something like `release_file_folder = os.path.dirname(d_content.d_artifacts[0].relative_path)` and then combine `self.remote.url` with that `release_file_folder`. We might need to watch out in case `d_content.d_artifacts` is the empty list though, so this is a bit messy...

CHANGES/399.feature
quba42: Hmm, I am unsure what it is. Allthough I have heard people say things like: "Unclear documentation/error messages are a bug" so maybe it is a bugfix?

pulp_deb/app/tasks/synchronizing.py
quba42: I think `d_content.content.relative_path` might also contain the path we want (and avoids the empty list issue). I am not 100% sure though.

CHANGES/399.feature
hstct: Bugfix is fine with me :slightly_smiling_face:

hstct (approved):

CHANGES/399.feature
Manisha15: Changed it to bugfix.

hstct:
LGTM

PR #737, opened 2023-03-26
Update CI files from 2.19 branch

pulpbot:
[noissue]


mdellweg:
Looks like we either need to backport a fixture in pulpcore, or bump the lower bound on it here.

quba42:
> Looks like we either need to backport a fixture in pulpcore, or bump the lower bound on it here.

Do you know what version of pulpcore has the fixture in question?

quba42:
@mdellweg Just checking if I understood what is happening here: Before the CI was testing against the tip of the pulpcore `3.20` branch using the template config option `pulpcore_branch: '3.20'`. Now it is testing against a released version, and no released version `3.20.*` actually has everything present on the tip of the `3.20` branch. So the new way of testing has revealed we were actually broken against all released versions of `3.20.*`, which is why we are now raising the floor to `3.21.6`?

quba42:
I am not particularly attached to the combination of pulp_deb 2.19.* with pulpcore 3.20.* so I am fine with merging this. Especially if that combination was broken anyway!

quba42 (approved):

mdellweg:
> @mdellweg Just checking if I understood what is happening here: Before the CI was testing against the tip of the pulpcore `3.20` branch using the template config option `pulpcore_branch: '3.20'`. Now it is testing against a released version, and no released version `3.20.*` actually has everything present on the tip of the `3.20` branch. So the new way of testing has revealed we were actually broken against all released versions of `3.20.*`, which is why we are now raising the floor to `3.21.6`?

Worse: Before we only ever tested against pulpcore:main, because pulp components were exempt from the lower bounds mechanism. The main reason for raising to 3.21 instead of 3.20 is that the 3.20 branch can be regarded discontinued. And it is unlikely it will ever again receive a next release. We would not advise anyone to have it still installed. The next best thing to a still supported release branch is 3.21.

PR #738, opened 2023-03-26
Update CI files from 2.18 branch

pulpbot:
[noissue]


mdellweg:
I think this needs a newer sphinx version to build. So remove the version constraint in doc_requirements.txt.

mdellweg (approved):

PR #741, opened 2023-04-03
Make Release structure independent

quba42:
This moves the colliding `Releases` problem (https://github.com/pulp/pulp_deb/issues/599) into the `ReleaseComponent` model.

This does not make the original problem better or worse, but if frees up the `Release` model to be used for https://github.com/pulp/pulp_deb/pull/656 without increasing the potential for collisions.

Once done, the `Release` model will no longer be structure relevant, and obtain a completely new meaning.

The set of models that define repo metadata structure will be reduced to just `ReleaseComponent`, `PackageReleaseComponent`, and `ReleaseArchitecture` (though it may be possible to eliminate the role of the latter as well. As such the overall representation of metadata structure becomes significantly simpler, literally requiring less entities to carry the same information.


quba42:
I think I might be unexpectedly close to fully solving the problem, by making the publish robust with respect to "duplicate" structure content.

quba42:
> I think I might be unexpectedly close to fully solving the problem, by making the publish robust with respect to "duplicate" structure content.

Never mind, I am back to the original plan of this being a pure refactor for now. This does not solve the colliding Release issue, it only transforms it into a colliding ReleaseComponent issue.

pulp_deb/app/migrations/0021_remove_release_from_structure_types.py
```diff
@@ -0,0 +1,158 @@
...
+            preserve_default=False,
+        ),
+        migrations.AddField(
+            model_name='releasecomponent',
+            name='suite',
+            field=models.TextField(default=''),
+            preserve_default=False,
+        ),
+        migrations.RunPython(move_structure_info, reverse_code=use_related_release, elidable=True),
+        # TODO: No idea what elidable=True does!
```
mdellweg, line 111: It tells Django that this datamigration is a noop on an empty database (i.e. only migrating existing data, not creating new one.). It is mainly important for squashing migrations, which we will most probably never be able to do, but still a good habit.

pulp_deb/app/viewsets/content.py
```diff
@@ -252,7 +254,7 @@ class ReleaseArchitectureFilter(ContentFilter):
 
     class Meta:
         model = models.ReleaseArchitecture
-        fields = ["architecture", "release"]
+        fields = ["architecture", "distribution", "codename", "suite"]
```
daviddavis, line 257: I think you should probably call out these changes to the filters with a changelog entry.

daviddavis:
Overall, this looks good to me. I have a question though. Today for example we'll have to repos (A and B) and repo A will have release A1 and repo B will have release A2. These releases will have the same distribution/codename/suite but different ids since they have different values for origin/label/etc. We're able to look up their components by the release id and this works (no problems). I'm imagining after this change though, we'll need to look up their components by distribution, suite, codename **AND** repository version, correct? If we just switch from using release id to distribution/suite/codename, I think we'll get the components/architectures for both releases since they share distribution/suite/codename.

quba42:
> I'm imagining after this change though, we'll need to look up their components by distribution, suite, codename **AND** repository version, correct? If we just switch from using release id to distribution/suite/codename, I think we'll get the components/architectures for both releases since they share distribution/suite/codename.

The basic point is correct, the association between a Release and its ReleaseComponents etc. is only really meaningful if both are present in the same repository (version).

This is arguably true even before this change. If you just look for all the ReleaseComponents associated with a particular Release today, you might find ReleaseComponents that are not actually in any latest repository version (or in some repository but not in others), and therefore not particularly relevant to some given repository you might be interested in.

On the details I think I have a slight correction to how one might querry for the ReleaseComponents associated with a given Release (in a given repository version): With this change, the Release model picks up a `repo_key_fields = ("distribution",)`. This means any repository version can have at most one Release per distribution (and by extension each Release corresponds to exactly one Release file once you publish). As a result, if you want to find all the ReleaseComponents that belong with the one Release associated with a particular distribution of a particular repository version, you can simply query for all ReleaseComponents within the same repository version that have the same distribution. There is no need to consider the suite or codename. If there are mutliple ReleaseComponents in your repository with the same distribution, and the same component (but differences in codename/suite), then they will essentially be merged into one during publish.

pulp_deb/app/viewsets/content.py
quba42: See the `CHANGES/748.removal` I added. Does that changelog entry communicate everything you would expect here?

m-bucher (approved):
LGTM

quba42 (commented):
We still plan to do some test runs of the DB migration on large grown systems.

However, we have decided to merge this PR as is now, since the code changes are blocking so many other PRs.

If our migration tests still turn up anything, we will add any fixes in a new PR.

PR #757, opened 2023-05-04
Fix structured copy

quba42:
[noissue]

This was broken by commit f5bb7c8b6e1d50ec2a6d40029152e6d9ca3704ea

I am adding this fix now, but we also need a regression test.

### DIFF
--- a/pulp_deb/app/tasks/copy.py
+++ b/pulp_deb/app/tasks/copy.py
@@ -3,7 +3,13 @@ from django.db.models import Q
 
 from pulpcore.plugin.models import Content, RepositoryVersion
 
-from pulp_deb.app.models import AptRepository, Package, ReleaseArchitecture, PackageReleaseComponent
+from pulp_deb.app.models import (
+    AptRepository,
+    Package,
+    PackageReleaseComponent,
+    Release,
+    ReleaseArchitecture,
+)
 
 import logging
 from gettext import gettext as _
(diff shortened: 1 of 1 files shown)


pulp_deb/app/tasks/copy.py
```diff
@@ -50,9 +56,12 @@ def find_structured_publish_content(content, src_repo_version):
     # Package release components, release components, release to be copied based on packages
     for pckg in package_release_components.iterator():
         if pckg.package in packages:
-            structured_publish_content.update(
-                [pckg.pk, pckg.release_component.pk, pckg.release_component.release.pk]
-            )
+            structured_publish_content.update([pckg.pk, pckg.release_component.pk])
+            release = Release.objects.filter(
+                pk__in=src_repo_version.content, distribution=pckg.release_component.distribution
```
hstct, line 61: Can this query return more than one object? If not can we transform it to a single `.get()`?

quba42: It should not be possible to have more than one, since we are filtering from a single repo version, and Release has `repo_key_fields = ("distribution",)`. I suppose it could be possible for historical repo_versions to have more than one if they were created before we added the `repo_key_fields` (very rare edge case).

While this would not happen for synced content, it is legal for it to be zero results though in which case `get()` would then throw an error we would need to catch. I am undecided, but leaning towards leaving it as is for robustness against rare edge cases...

hstct: Using `get()` would be more performant in this case but I'm fine with keeping it in for edge cases.

hstct (approved):

PR #760, opened 2023-05-09
Merge colliding structure content

### LINKED ISSUE #599: ReleaseComponent content with different codename or suite but equal distribution collides

quba42:
closes #599

This will require a DB migration that alters the structure content in existing repository versions.

The plan is to release this change as pulp_deb 3.0.0!


quba42:
We now have a working state for the DB migration. However, I have yet to introduce batching. Once this is done the migration requires performance testing with installations with large amounts of APT content.

pulp_deb/app/migrations/0025_merge_colliding_structure_content.py
```diff
@@ -0,0 +1,233 @@
...
+
+def merge_colliding_structure_content(apps, schema_editor):
+    ReleaseArchitecture = apps.get_model('deb', 'ReleaseArchitecture')
+    ReleaseComponent = apps.get_model('deb', 'ReleaseComponent')
+    PackageReleaseComponent = apps.get_model('deb', 'PackageReleaseComponent')
+    RepositoryContent = apps.get_model('core', 'RepositoryContent')
+    Content = apps.get_model('core', 'Content')
+
+    migration_log = open("/var/log/pulp_deb_migration_0025.txt", "a")
+    migration_log.write("{}: Starting data migration\n".format(datetime.now()))
```
quba42, line 19: Any opinions about this migration writing its own log file?

quba42: So the CI found a potential Problem with this:
```
E       PermissionError: [Errno 13] Permission denied: '/var/log/pulp_deb_migration_0025.txt'
```

hstct: I assume the migration is run by the pulp user which doesn't have access there.

`/tmp/` should work though.

hstct: Instead of manually writing the log couldn't we use the django logger here? And have it log `django.db.migrations`?

quba42: We may need to use the django logger to make this consistent with everything else, but there is real value in having the information what this migration did in one place that sticks around for a long time => for example its own log file. :wink:

hstct: Huh, but setting up the django logger to write a specific log e.g. `django.db.migrations` to a file is also possible? We just need to make adjustments in the `settings.py`. Something like this:

```
# settings.py

LOGGING = {
    'handlers': {
        'file': {
            'level': 'INFO',
            'class': 'logging.FileHandler',
            'filename': '/path/to/your/logfile.log',
        },
    },
    'loggers': {
        'django.db.migrations': {
            'handlers': ['file'],
            'level': 'INFO',
            'propagate': True,
        },
    },
}
```

It should extend the `pulpcore` settings shouldn't it? And `pulpcore` does not use this logger or has a file handler defined. (ref: https://docs.djangoproject.com/en/4.2/howto/logging/#configure-a-handler)

Unless we are talking about the `/tmp/` path not being persistent but there I don't really know where we have access to at runtime probably `/src/` as well but that's also not a great location. I ran into similar issues with creating symlinks for the local test data. https://github.com/pulp/pulp-oci-images/issues/409

So maybe @mikedep333 can tell us more where we could store a persistent log file?

ggainey: discussion from open-floor: opening one's own logfile is Fraught With Peril. Log "the standard way", and use filtering at the journalctl level to find "your" log entries.

hstct: If this means writing to the syslog (e.g. `/var/run/log`) I'm pretty sure that pulp does not have the permission to do so either doesn't it?

quba42: At least in oci-env, it looks like the current version (with `log = logging.getLogger(__name__)`) writes to stdout when running `pulpcore-manager migrate`, which is not exactly ideal.

hstct: > At least in oci-env, it looks like the current version with `log = logging.getLogger(__name__)`, writes to stdout when running `pulpcore-manager migrate`, which is not exactly ideal.

That's because there is only one log handler defined in pulpcore: https://github.com/pulp/pulpcore/blob/main/pulpcore/app/settings.py#L215

Which is the console log. You need to define new logging rules in the `settings.py` as I mentioned above. E.g. a handler for file or syslog. But we will run into permission issues with both I believe

quba42:
Ugh. Why are the tests now failing. These were already green before. It looks suspiciously like some change somewhere else broke this test.

quba42:
This looks like a good candidate for what broke the tests: https://github.com/pulp/pulpcore/pull/3904

hstct:
I think it is because the hosted data fixtures got updated? I ran into the same problem with the converted publish tests last week this is what i did to fix it for now: https://github.com/pulp/pulp_deb/pull/776/files#diff-39a085d2db2d6234c049924d46d2c3bac7ef2029a532cb9aa0e09d3a4a91f282R443

But it still worked for the pre conversion tests. I made a note to talk about this with you :)

The essentials field is definitely in the pkg metadata though I checked. Maybe it is new because of `equivs-build` got updated somehow (which is likely since the author field had some changes as well)?

Either way the field doesn't seem to get published. It is a valid field though https://www.debian.org/doc/debian-policy/ch-controlfields.html#essential

pulp_deb/app/migrations/0025_merge_colliding_structure_content.py
```diff
@@ -0,0 +1,313 @@
...
+import logging
+
+from datetime import datetime
+
+from django.db import migrations, models
+from django.core.exceptions import ObjectDoesNotExist
+
+BATCH_SIZE = 1000
+
+log = logging.getLogger(__name__)
```
quba42, line 12: This currently results in the following logger name: `pulp_deb.app.migrations.0025_merge_colliding_structure_content`

If we want to make it easy to configure a logger that controls where "DB migration logging" goes, we might want to use a generic name for that in all our migrations: In the internet we found `django.db.migrations`. Possible alternative: `pulp_deb.app.migrations`.

quba42: Using the relevant logger name, logging can be re-directed by configuring a logger as described here: https://github.com/pulp/pulp_deb/pull/760#discussion_r1227716457

PR #776, opened 2023-05-25
Update publish tests to use pytest

### LINKED ISSUE #828: Convert publish tests to pytest.
Same as with previous tests the publishing tests should also be using the pytest framework.

hstct:
closes #828


pulp_deb/tests/functional/values.py
hstct, line 1: Not sure if `values.py` is a good name for this. Open for suggestions :)

mdellweg: You think constants.py does not fit for it?

hstct: Uhm, true I could put it in there. In my mind I was thinking having a distinct file for expected values. But now that you mention it they are just constants and other tests use values from `constants.py` for their assertions.

Gonna change it :)

quba42 (approved):
Lets merge.

PR #781, opened 2023-06-02
Add tests for related content filters

### LINKED ISSUE #780: Add a test for the new related content filters
#646 was added with #647, but still needs a "happy path" test.

It should be possible to extend an existing sync test by just applying the filter to the sync result, and checking that it finds the expected things.

sdherr:
And make filters return empty list if content not in RepoVersion instead of raising a ValidationError, which I believe to be more correct behavior.

closes #780

### DIFF
--- a/pulp_deb/tests/functional/api/test_sync.py
+++ b/pulp_deb/tests/functional/api/test_sync.py
@@ -252,6 +252,9 @@ def test_sync_optimize_skip_unchanged_package_index(
     deb_repository_factory,
     deb_get_repository_by_href,
     deb_sync_repository,
+    apt_package_api,
+    apt_release_api,
+    apt_release_component_api,
 ):
     """Test whether package synchronization is skipped when a package has not been changed."""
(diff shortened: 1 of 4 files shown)


CHANGES/780.feature
hstct, line 1: Rather than a feature can we change this to `.misc` to match the issue?

pulp_deb/tests/functional/api/test_sync.py
hstct, line 1: The test is looking good :+1:

hstct (changes requested):

CHANGES/780.feature
sdherr: Sure, done.

sdherr:
@hstct  I didn't change anything other than rename the CHANGES file and I don't think these test failures can be related to me.

quba42 (approved):
LGTM. I have seen it go green before pulpcore broke a unit test so feel free to merge.

PR #782, opened 2023-06-05
Add version, origin, label, and description fields to releases

### LINKED ISSUE #449: Make various release file fields user configurable

quba42:
fixes #449 

Note that this PR replaces #656. This PR is based on the previous one, but with some changes:

* We chose not to add the new fields to the ReleaseFile model for now, since the feature only really needs them on the `Release` model, and there are concerns around the `ReleaseFile` model's uniqueness constraints we would rather address in a separate PR. (Re-working the ReleaseFile model is not an immediate priority).
* We chose not to have a `sync_extra_release_fields` option on the remote. Instead we want to create a `publish_upstream_release_fields` on the repository/publish side. This should be done in a follow up PR that is to be released together with this one.
* The publish code was somewhat refactored in a bid to simplify it.

Follow up TODOs:

This PR is ready for final review and merge, but will require some follow up tasks once merged:

* Implement the `publish_upstream_release_fields` option.
* Extend an existing test, to ensure the new fields are properly synced and published.
* Consider removing the `repository.description` field, since it adds a lot of complexity and does not add much value beyond the new approach of storing extra release file fields on the Release model.
* Consider handling the existing `codename` and `suite` fields more like the new fields added by this PR.


CHANGES/449.feature
```diff
@@ -0,0 +1,3 @@
+Added ``version``, ``origin``, ``label``, and ``description`` fields to Releases. These fields can
+be set when creating new Releases via the API or when syncing if the Remote's
+``sync_extra_release_fields`` option is set to true.
```
daviddavis, line 3: I think this line should be dropped.

daviddavis, line 3:
```suggestion
Added ``version``, ``origin``, ``label``, and ``description`` fields to Releases. These fields can
be set when creating new Releases via the API.
```

pulp_deb/app/migrations/0024_add_release_fields.py
```diff
@@ -0,0 +1,41 @@
...
+
+    dependencies = [
+        ('deb', '0023_add_default_signing_services'),
+    ]
+
+    operations = [
+        migrations.AlterUniqueTogether(
+            name='release',
+            unique_together=set(),
+        ),
```
daviddavis, line 16: Is this itentional?

quba42: I am also puzzled by this. This was automatically generated by django. Since the migration had changed so much from before, I decided to remove and re-generate it, and this is the result.

CHANGES/449.feature
quba42: You are absolutely correct of course.

pulp_deb/app/migrations/0024_add_release_fields.py
daviddavis: Interesting. I guess we should just defer to Django's judgement then.

PR #791, opened 2023-06-06
Updated the scoping of the pytest fixtures.

### LINKED ISSUE #790: Use better scoping for pytest fixtures
Add more scoping to the fixtures in `conftest.py` similar to other pulp projects.

- API fixtures should be scoped to `session`
- While factory fixtures should be scoped to `class`

hstct:
closes #790

### DIFF
--- a/CHANGES/790.misc
+++ b/CHANGES/790.misc
@@ -1,0 +1,1 @@
+Added better scoping for the pytest fixtures.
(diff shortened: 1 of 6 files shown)


quba42 (approved):
Looks Sane, and tests are green.

Reviewed in pair session.

PR #794, opened 2023-06-07
Add publish_upstream_release_fields option

### LINKED ISSUE #793: Create publish_upstream_release_fields repo/publication option
The idea is to make the new behaviour from https://github.com/pulp/pulp_deb/pull/782 opt-in on the publication side.

That way we can prevent new syncs + publish in existing repos from suddenly changing the release file fields post pulp_deb upgrade.

This would be undesirable since it requires all APT clients consuming the publication to confirm that the changed release file fields are acceptable.

Therefore existing repos should initially be seeded with `publish_upstream_release_fields = False` while the default value for new repos should be `True`.

hstct:
closes #793


hstct:
Still needs tests.

pulp_deb/app/migrations/0025_aptrepository_publish_upstream_release.py
```diff
@@ -0,0 +1,25 @@
...
+
+    operations = [
+         # create the field with a default of False so that the field is populated as False for
+         # existing models but then set the default to True for new repositories going forward
+         migrations.AddField(
+            model_name='aptrepository',
+            name='publish_upstream_release_fields',
+            field=models.BooleanField(default=False),
+         ),
+        migrations.AlterField(
```
quba42, line 20: The indentation looks wrong here. Not sure how I managed that.

quba42 (approved):
I am tempted to just merge. The docs job did not start, but the docs have not changed, and the azure job is the same recurring Azure race condition as always...

PR #803, opened 2023-06-19
Fixed package matching not being caseless

hstct:
[noissue]


hstct:
This was ultimately an issue with the test packages and has been fixed here instead: https://github.com/pulp/pulp-fixtures/pull/240

PR #804, opened 2023-06-19
Fix content creation when unapplying migration 0021

quba42:
[noissue]

The previous version cannot work, but no one noticed, since it is not released yet, and is only designed to catch a rare edge case when unapplying the migration 0021.


quba42:
Never mind, this is released as part of pulp_deb 2.21

hstct (approved):

PR #810, opened 2023-06-20
Normal gettext import

quba42:
Not sure we really want to do this, so this is just for evaluation.


hstct:
Do we even use internationalization or localization? My general stance on this is we should support it so it will be easier in the future if someone wants to actually work on a translated version and it is low cost to do so now. We should however use this renaming of the import to make sure it is used consistently meaning double check if we only use `gettext` in all log statements.

On the topic of renaming the import from `_` to `gettext` I'm okay with leaving it at that but maybe we can even find something shorter that is not using the special character `_`. E.g. `gt` or something more descriptive `i18n`?

mdellweg:
Whatever we do, be aware that extracting the strings for translation is not done by executing the code, but statically analyzing it. The gettext util itself will look for `_("...")` (maybe even with a regex). I think it can be told to look for something different though.

quba42:
Thanks @mdellweg, that was the explanation I was looking for, for why things currently are the way they are.
May well be easiest to just forget about this then and leave everything as is. :smile:

PR #827, opened 2023-07-04
Update Changelog

pulpbot:
[noissue]


quba42:
This is clearly a mistake in automation.

quba42:
Ok, looks like this correctly added a beta version that was missed way back when to the changelog...

PR #831, opened 2023-07-13
Check `done` instead of `total` in sync test

mdellweg:
[noissue]

### DIFF
--- a/pulp_deb/tests/functional/api/test_sync.py
+++ b/pulp_deb/tests/functional/api/test_sync.py
@@ -343,6 +343,7 @@ def test_sync_orphan_cleanup_fail(
     deb_sync_repository,
     orphans_cleanup_api_client,
     monitor_task,
+    delete_orphans_pre,
 ):
     """Test whether an orphan cleanup is possible after syncing where only some PackageIndices got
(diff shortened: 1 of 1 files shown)


hstct:
Not sure why the amount of the content change again 😕

mdellweg:
Seems like the artifact count can differ. Would it be ok to only check the content count?

quba42:
I guess the idea is to assert that certain orphans were removed by the orphan cleanup. Maybe there is a more direct way to assert this. (Like record the actual artifact href that we think is an orphan and then assert that it no longer exists). This would avoid "counts" altogether.

mdellweg:
> I guess the idea is to assert that certain orphans were removed by the orphan cleanup. Maybe there is a more direct way to assert this. (Like record the actual artifact href that we think is an orphan and then assert that it no longer exists). This would avoid "counts" altogether.

My thinking is, we know about the content we expect to be removed. Let pulpcore handle the artifacts...
But we can also issue a list of hrefs to orphan cleanup. And if we requested 2 items to remove and the done is 2 that would be very precise. I'd rather not do that in this PR...

hstct (approved):

PR #833, opened 2023-07-13
Generating checksum named files for aptbyhash

### LINKED ISSUE #795: AptByHash
**The Problem.**

**Solution**

adamsanaglo:
Mitigating the Hash Sum Mismatch error by generating additional metadata files named by checksum for AptByHash.
closes #795


daviddavis:
@adamsanaglo the number of files for this PR seems quite large (84 files). A number of them should be ignored by the .gitignore so I am not sure how you have checked them in. If you run this, it should remove them from git:

```bash
git ls-files -c --ignored --exclude-standard | xargs git rm -r --cached
```

pulp_deb/app/constants.py
```diff
@@ -17,3 +17,4 @@
 
 # Represents null values since nulls can't be used in unique indexes in postgres < 15
 NULL_VALUE = "__!!!NULL VALUE!!!__"
+APT_BY_HASH = True
```
daviddavis, line 20: I think this should probably be a setting (see pulp_deb/app/settings.py).

CHANGES/795.feature
```diff
@@ -0,0 +1 @@
+Mitigating the Hash Sum Mismatch error by generating additional metadata files named by checksum
```
daviddavis, line 1: This probably needs some more information for users about the feature and maybe a link to https://wiki.ubuntu.com/AptByHash.

pulp_deb/app/tasks/publishing.py
```diff
@@ -277,10 +280,38 @@ def finish(self):
                 publication=self.parent.publication, file=File(open(package_index_path, "rb"))
             )
             package_index.save()
+
+            # Generating metadata files using checksum
+            if APT_BY_HASH:
+                by_hash_path = os.path.join(os.path.dirname(package_index_path), 'byhash')
+                if not os.path.exists(by_hash_path):
+                    os.makedirs(by_hash_path)
+                package_hashed_index_path = os.path.join(by_hash_path, _sha256sum(package_index_path))
```
daviddavis, line 289: I think the sha256 will already be available on the artifact for package_index:

```python
package_index.contentartifact_set.first().artifact.sha256
```

daviddavis:
A couple other things:

* There was a discussion on the issue you filed about how to persist the byhash files across publications. Have you given any thought to that? 
* This probably could use some user docs to at least document the config setting.

pulp_deb/app/tasks/publishing.py
daviddavis: Also, the checksums will probably need to be configurable. And it will likely need to be validated as a subset of [`ALLOWED_CONTENT_CHECKSUMS`](https://github.com/pulp/pulpcore/blob/bbb30b7a5de3cab3ac13e1913facb33f1292e2fe/pulpcore/app/settings.py#L283-L287).

pulp_deb/app/tasks/publishing.py
```diff
@@ -217,7 +219,7 @@ def publish(repository_version_pk, simple=False, structured=False, signing_servi
                 for release_helper in release_helpers:
                     release_helper.save_signed_metadata()
 
-    log.info(_("Publication: {publication} created").format(publication=publication.pk))
+    log.info(_("Publication: {publication} createddddd").format(publication=publication.pk))
```
daviddavis, line 222: image

adamsanaglo:
Thank you Mr. Davis.

Best regards,
Adams

pulp_deb/app/tasks/publishing.py
mdellweg: ... and sha256 is not allowed to be excluded from `ALLOWED_CONTENT_CHECKSUMS`.

hstct:
The linting stage needs pass before the actual tests are triggered. Besides the black formatting complain it is very likely that the stage will also complain because the commit message is missing a `closes #795` in the 2nd row of the commit message.

pulp_deb/app/tasks/publishing.py
sdherr, line 289: Using `os.path` functions certainly works, but the more modern-python way to do this is to [use pathlib](https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module):
```
from pathlib import Path

by_hash_path = Path(package_index_path) / "byhash"
by_hash_path.mkdir(parents=True, exists_ok=True)
package_hashed_index_path = by_hash_path / _sha256sum(package_index_path)
```

pulp_deb/app/tasks/publishing.py
```diff
@@ -277,10 +280,38 @@ def finish(self):
...
+                package_hashed_index.save()
+            # Done generating
+
             gz_package_index = PublishedMetadata.create_from_file(
                 publication=self.parent.publication, file=File(open(gz_package_index_path, "rb"))
             )
             gz_package_index.save()
+
+            # Generating metadata files using checksum
+            if APT_BY_HASH:
```
sdherr, line 303: There's no reason to have the APT_BY_HASH code split into two different if blocks; move the `package_index_path` code above down into this block too. Also there's a lot of duplication between the two, may be a good opportunity to de-duplicate by creating a helper function or wrapping in a loop.

pulp_deb/app/tasks/publishing.py
```diff
@@ -277,10 +280,38 @@ def finish(self):
...
+                gz_package_hashed_index_path = os.path.join(by_hash_path, _sha256sum(gz_package_index_path))
+                shutil.copyfile(gz_package_index_path, gz_package_hashed_index_path)
+                gz_package_hashed_index = PublishedMetadata.create_from_file(
+                    publication=self.parent.publication, file=File(open(gz_package_hashed_index_path, "rb"))
+                )
+                gz_package_hashed_index.save()
+            # Done generating
+
             self.parent.add_metadata(package_index)
             self.parent.add_metadata(gz_package_index)
```
sdherr, line 316: [Genuine question] Do you need to add the `package_hashed_index` and `gz_package_hashed_index` here too?

pulp_deb/app/tasks/publishing.py
```diff
@@ -394,3 +425,10 @@ def _zip_file(file_path):
         with GzipFile(gz_file_path, "wb") as f_out:
             shutil.copyfileobj(f_in, f_out)
     return gz_file_path
+
+def _sha256sum(path):
+    h = hashlib.sha256()
```
sdherr, line 430: FYI if you need to add support for other checksums the more generic way to do this is to call `hashlib.new()`:
```
h = hashlib.new("sha256")
h.update("whatever")
h.hexdigest()
```

pulp_deb/app/tasks/publishing.py
```diff
@@ -277,10 +280,38 @@ def finish(self):
...
+            # Generating metadata files using checksum
+            if APT_BY_HASH:
+                by_hash_path = os.path.join(os.path.dirname(package_index_path), 'byhash')
+                if not os.path.exists(by_hash_path):
+                    os.makedirs(by_hash_path)
+                package_hashed_index_path = os.path.join(by_hash_path, _sha256sum(package_index_path))
+                shutil.copyfile(package_index_path, package_hashed_index_path)
+                package_hashed_index = PublishedMetadata.create_from_file(
+                    publication=self.parent.publication, file=File(open(package_hashed_index_path, "rb"))
+                )
```
daviddavis, line 293: I looked at [the `create_from_file` method in pulpcore](https://github.com/pulp/pulpcore/blob/f7e62bb03948db8ae230a64d9642d7e144ce02ca/pulpcore/app/models/publication.py#L271) and I think you can just specify the relative_path instead of copying the file. Maybe something like this:

```python
package_hashed_index = PublishedMetadata.create_from_file(
    publication=self.parent.publication,
    file=package_index_path,
    relative_path=f"{package_index_path}/byhash/{package_index.contentartifact_set.first().artifact.sha256}"
)
```

pulp_deb/app/tasks/publishing.py
adamsanaglo: I found that by doing so the checksum named files would be added to the InRelease file, which would be cyclical based on this specification:
`A dists/<release>/by-hash/<algorithm>/<hash> URL MUST return the corresponding file for every hash and hash algorithm listed in dists/<release>[/InRelease](https://wiki.ubuntu.com/AptByHash/InRelease) for that file.`

pulp_deb/app/tasks/publishing.py
```diff
@@ -277,10 +279,30 @@ def finish(self):
...
+                package_hashed_index.save()
+
+                gz_package_hashed_index_path = _create_checksum_file(gz_package_index_path)
+                gz_package_hashed_index = PublishedMetadata.create_from_file(
+                    publication=self.parent.publication,
+                    file=File(open(gz_package_hashed_index_path, "rb")),
+                )
+
+                gz_package_hashed_index.save()
```
sdherr, line 303:
```suggestion
for path in (package_index_path, gz_package_index_path):
    hashed_index_path = _create_checksum_file(...
    hashed_index = PublishedMetadata...
    hashed_index.save()
```

sdherr:
> Besides the linting stage: ready-to-ship is complaining because your commit is missing either a `[noissue]` or better a `closes #795` (for the AptByHash issue) in the 2nd row of the commit message. This will not block tests I believe but it needs to happen at some point before the merge :)

The PR Description becomes the commit message when it's merged so he's asking you to fill in the Description and include a `closes #795` to link the two. This is what the `ready-to-ship` check is failing on if you click "details".

pulp_deb/app/tasks/publishing.py
```diff
@@ -394,3 +416,19 @@ def _zip_file(file_path):
...
+    h = hashlib.new(APT_BY_HASH_CHECKSUM_TYPE)
+    with open(path, "rb") as f:
+        for line in f:
+            h.update(line)
+    return h.hexdigest()
+
+
+def _create_checksum_file(file_path):
+    by_hash_path = Path(file_path).parents[0] / "byhash"
+    by_hash_path.mkdir(parents=True, exists_ok=True)
```
sdherr, line 431: `exist_ok`, singular. This is what the tests are failing on, although it's less obvious because this error message is mixed in with some others that don't look like they can be related to this PR.

hstct:
> The PR Description becomes the commit message when it's merged so he's asking you to fill in the Description and include a closes #795 to link the two. This is what the ready-to-ship check is failing on if you click "details".

Not exactly. It's the first commit that is missing the link to the issue. Which can be done with squashing all the commits and give it one proper commit message including the `closes #795`. This can be done now and add future commits with an amend commit / force push workflow or squash at the end once the feature is ready.

pulp_deb/app/tasks/publishing.py
```diff
@@ -413,3 +427,19 @@ def _zip_file(file_path):
...
+        for line in f:
+            h.update(line)
+    return h.hexdigest()
+
+
+def _create_checksum_file(file_path):
+    by_hash_path = Path(file_path).parents[0] / "byhash"
+    by_hash_path.mkdir(parents=True, exist_ok=True)
+    hashed_path = by_hash_path / _checksum_file(file_path)
+    shutil.copyfile(file_path, hashed_path)
```
daviddavis, line 444: You shouldn't need to copy the file at all if you just specify `relative_path` when creating `PublishedMetadata`. See my comment [here](https://github.com/pulp/pulp_deb/pull/833#discussion_r1265810805).

pulp_deb/app/settings.py
```diff
@@ -7,3 +7,8 @@
 
 FORBIDDEN_CHECKSUM_WARNINGS = True
 FORCE_IGNORE_MISSING_PACKAGE_INDICES = False
+
+PUBLISH_RELEASE_FILE_LABEL = False
+PUBLISH_RELEASE_FILE_VERSION = False
```
daviddavis, line 12: Are you adding these config variables? I don't see them referenced elsewhere in the PR.

pulp_deb/app/tasks/publishing.py
```diff
@@ -413,3 +427,19 @@ def _zip_file(file_path):
...
+def _checksum_file(path):
+    h = hashlib.new(APT_BY_HASH_CHECKSUM_TYPE)
+    with open(path, "rb") as f:
+        for line in f:
+            h.update(line)
+    return h.hexdigest()
+
+
+def _create_checksum_file(file_path):
+    by_hash_path = Path(file_path).parents[0] / "byhash"
```
daviddavis, line 441: I don't thin kthis path is quite right. From the wiki:

> A dists/<release>/by-hash/<algorithm>/<hash> URL MUST return the corresponding file for every hash and hash algorithm listed in dists/<release>[/InRelease](https://wiki.ubuntu.com/AptByHash/InRelease) for that file.

So you need "by-hash" instead of "byhash" and the algorithm type.

pulp_deb/app/settings.py
adamsanaglo: I left them in during a merge

pulp_deb/app/tasks/publishing.py
```diff
@@ -413,3 +430,9 @@ def _zip_file(file_path):
         with GzipFile(gz_file_path, "wb") as f_out:
             shutil.copyfileobj(f_in, f_out)
     return gz_file_path
+
+
+def _fetch_file_checksum(file_path, index):
+    h = getattr(index.contentartifact_set.first().artifact, APT_BY_HASH_CHECKSUM_TYPE)
+    hashed_path = Path(file_path).parents[0] / "by-hash" / APT_BY_HASH_CHECKSUM_TYPE / h
```
daviddavis, line 437:
```suggestion
    hashed_path = Path(file_path).parents[0] / "by-hash" / CHECKSUM_TYPE_MAP[APT_BY_HASH_CHECKSUM_TYPE] / h
```

functest_requirements.txt
```diff
@@ -1,3 +1,4 @@
 pulp-smash @ git+https://github.com/pulp/pulp-smash.git
 pytest
 python-debian>=0.1.36
+epdb
```
mdellweg, line 4: I don't think this is a test requirement.

hstct:
I have a few questions regarding this feature and I guess AptByHash in general. As I understand it this PR will only fetch the `/by-hash/` for the components (e.g. `/bullseye/main/binary-amd64/by-hash/`)? Will this be sufficient for utilizing this `apt` feature?

In the issue there is a specification (https://wiki.ubuntu.com/AptByHash) which only mentions the `/by-hash/` of a release - so that's where I'm confused.

I tested the feature branch with various debian repositories and it does fetch the component hashes if they exist on the remote and doesn't if the remote does not provide it.

NOTE: I realized unrelated to this feature that my test environment somehow doesn't sign Releases currently. So I will try to fix that today and have another look at it and maybe it answers some of my questions. I just wanted to give a status update.

daviddavis:
I was able to test out the feature without signing the release. Here's my Dockerfile:

```
FROM ubuntu:22.04

RUN echo "deb [trusted=yes] http://localhost:24816/pulp/content/test/ mydist mycomp" > /etc/apt/sources.list.d/test.list
RUN apt-get update
RUN apt-get install -y <package name>
```

functest_requirements.txt
```diff
@@ -1,3 +1,3 @@
 pulp-smash @ git+https://github.com/pulp/pulp-smash.git
 pytest
-python-debian>=0.1.36
+python-debian>=0.1.36
```
quba42, line 3: Nitpick: Try to avoid whitespace changes like this (`\ No newline at end of file`). I recommend configuring your editor to add newlines at the end of the file automatically.

pulp_deb/app/tasks/publishing.py
```diff
@@ -272,6 +271,7 @@ def add_package(self, package):
                 content_artifact=package.contentartifact_set.get(),
             )
             published_artifact.save()
+
```
quba42, line 274: Nitpick: Try to avoid whitespace changes where nothing else has changed (unless this was done by black?).

pulp_deb/app/tasks/publishing.py
```diff
@@ -296,10 +296,29 @@ def finish(self):
                 publication=self.parent.publication, file=File(open(package_index_path, "rb"))
             )
             package_index.save()
+
```
quba42, line 299: Same as the previous comment.

pulp_deb/app/tasks/publishing.py
```diff
@@ -296,10 +296,29 @@ def finish(self):
...
+                    for allowed_checksum in settings.ALLOWED_CONTENT_CHECKSUMS:
+                        if allowed_checksum in CHECKSUM_TYPE_MAP:
+                            hashed_index_path = _fetch_file_checksum(path, index, allowed_checksum)
+                            hashed_index = PublishedMetadata.create_from_file(
+                                publication=self.parent.publication,
+                                file=File(open(path, "rb")),
+                                relative_path=hashed_index_path,
+                            )
+                            hashed_index.save()
+            # Done generating
```
quba42, line 320: Nitpick: This comment is not needed. (It does not contain any information not already carried by the indentation).

quba42 (changes requested):
The below is just a first pass collecting nitpicks (small style improvements that are not that important).

I will collect any feedback about the feature itself in a separate review, but wanted to give you the chance to already fix these nitpicks in the meantime.

In addition to my comments below, I see the following small ToDos:

- [x] Once done with the nitpicks, it would be a good time to squash your commits together into a single commit. Right now there are a lot of intermediary commits that add things that are removed again in a later commit. This does not need to be kept in the history for ever.
- [ ] By convention, commit messages should leave the second line of the commit message blank, so there should be a blank line before the `closes #795` in your final commit message.

quba42:
It looks like most of my suggestions have been applied on the follow up PR #833 that replaces this one (since the work was taken over by a different developer).

As a result I am closing this PR in favor of #833

PR #839, opened 2023-07-18
Drop 2.18 and 2.19 support

quba42:
Also updates to latest CI and cleans up obsolete CI files!


template_config.yml
```diff
@@ -13,8 +13,6 @@ check_stray_pulpcore_imports: true
 ci_env: {}
 ci_trigger: '{pull_request: {branches: [''*'']}}'
 ci_update_branches:
-- '2.18'
-- '2.19'
 - '2.20'
 - '2.21'
```
quba42, line 17: We will still be updating the 2.20 branch used in Katello 4.7 and compatible with `pulpcore>=3.21.0,<3.25`.

That should be enough!

hstct:
> Also updates to latest CI and cleans up obsolete CI files!

Do you mean preemptively updating the CI ahead from the automatic (weekly) CI changes? Or where do they come from

quba42:
@hstct I noticed that the CI had "obsolete files" (which are not dropped automatically by the automatic CI update), so I wanted to clean those up, then I noticed that latest plugin template main branch had additional "obsolete files", so I as you put it "preemptively updated the CI" to latest main, so I could clean up all the latest "obsolete files".

That was a long way to say: "yes". 
:smile:

hstct:
Will the CI then create backports for 2.21 and 2.20 for this at the end of the week or do we need to do this manually?

quba42:
> Will the CI then create backports for 2.21 and 2.20 for this at the end of the week or do we need to do this manually?

The update CI job is independently run for each branch in the `ci_update_branches` list, so if `main` branch has already been preemptively updated by me, it will not open a PR for main (because there is nothing to do), but still open the automatic PRs for 2.21 and 2.21.

hstct:
Ok thanks for clarifying :)

hstct (approved):

PR #847, opened 2023-07-20
Convert import export tests

hstct:
closes #846

PR #851, opened 2023-07-24
Remove pulp smash

### LINKED ISSUE #796: Remove pulp-smash dependency
The switch to use local test data instead of the officially hosted ones at https://fixtures.pulpproject.org/ had some unintended side effects. Most tests are using some helper functions in in `pulp-smash` and they usually default on loading content from the hosted fixtures. This has already caused confusion multiple times during development.

Another reason is that `pulp-smash` is going to be deprecated at some point: https://discourse.pulpproject.org/t/retirement-of-pulp-smash/579

A lot of functions are already moved to `pulpcore` as pytest fixture but not everything. So it needs some refactoring.

hstct:
closes #796

### DIFF
--- a/functest_requirements.txt
+++ b/functest_requirements.txt
@@ -1,3 +1,6 @@
-pulp-smash @ git+https://github.com/pulp/pulp-smash.git
 pytest
 python-debian>=0.1.36
+pytest-custom_exit_code
+pytest-xdist
+trustme~=1.0.0
+proxy.py~=2.4.3
--- a/CHANGES/796.misc
+++ b/CHANGES/796.misc
@@ -1,0 +1,1 @@
+Removed the ``pulp-smash`` test dependency.
--- a/pulp_deb/tests/functional/api/test_crud_content_unit.py
+++ b/pulp_deb/tests/functional/api/test_crud_content_unit.py
@@ -1,5 +1,5 @@
 """Tests that perform actions over content unit."""
-import uuid
+from uuid import uuid4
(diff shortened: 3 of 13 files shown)


hstct:
depends on #776 and #847

functest_requirements.txt
```diff
@@ -1,3 +1,4 @@
+# TODO: REMOVE
 pulp-smash @ git+https://github.com/pulp/pulp-smash.git
```
hstct, line 2: Removing this dependency even though it is technically not used in `pulp_deb` has no impact when running it locally. However on github the test pipelines will fail because the `trustme` module cannot be imported which is a `pulpcore` dependency -> [see here](`https://github.com/pulp/pulp_deb/actions/runs/5647330466/job/15304403784#step:10:1678)

No clue why :/

hstct:
Depends on this pulpcore PR: https://github.com/pulp/pulpcore/pull/4177

quba42 (approved):
I did not fully understand the changes to `verify_publication_data`, and I don't fully understand why new pytest fixtures go in the files they go in as well as their naming scheme.

Putting aside the above caveats, these changes look sane to me, and the tests are green, and my gut feeling is that it makes more sense to get these refactors in and continue working on top, than it does to endlessly review them line by line, while continuing to work with the old test suit.

hstct:
On a second thought we can just merge it and if its causing too much confusion later on it is simple to change that wouldn't effect much.

PR #856, opened 2023-08-01
Update README badges

quba42:
[noissue]

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,9 @@
 # pulp_deb
 
-![release status](https://github.com/pulp/pulp_deb/workflows/Release%20Pipeline/badge.svg)
-![nightly status](https://github.com/pulp/pulp_deb/workflows/Deb%20Nightly%20CI/CD/badge.svg)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -29,6 +27,24 @@ The most important places:
 [5]: https://pypi.org/project/pulp-deb/
 [6]: https://pypi.org/project/pulp-deb-client/
 [7]: https://rubygems.org/gems/pulp_deb_client
```
hstct, line 29: It's weird to split the links up like that. Can we just put them all at the bottom?

hstct (changes requested):

README.md
quba42: Sure. I guess to be fully consistent, we should probably move the status update badge links into this list at the bottom as well...

hstct (approved):

quba42:
Squished the commits without changing anything.

quba42:
Note that this README is also used on [pypi.org](https://pypi.org/project/pulp-deb/).

PR #859, opened 2023-08-02
Change the publish behavior to use structured publish by default

### LINKED ISSUE #858: Change the default publish behavior to always use structured publish by default
Currently the default values for a publish are as follows: `structured=False` and `simple=False`. The serializer expects that at least one of the parameters is set to `True`.

Going forward with the recent structure changes we want to enable `structured` publish by default.

hstct:
closes #858


pulp_deb/app/tasks/publishing.py
```diff
@@ -70,8 +70,8 @@ def publish_verbatim(repository_version_pk):
 
 def publish(
     repository_version_pk,
-    simple=False,
-    structured=False,
```
hstct, line 74: Don't really know why the default value is set on so many locations. Seems to work without defining it here.

CHANGES/858.removal
```diff
@@ -0,0 +1 @@
+Changed the publish behavior to use structured publish by default.
```
quba42, line 1: I think we should be a little more descriptive about the before/after behaviour here, I came up with:
```
The ``pulp/api/v3/publications/deb/apt/`` API endpoint, used to require users to explicitly set at least one of ``simple`` or ``structured`` to ``True`` on the POST.
The new behavior is to default to ``structured=True`` and ``simple=False``.
```
I am unsure if we should also mention the effect this has on the corresponding Pulp CLI deb command. :thinking:

pulp_deb/app/serializers/publication_serializers.py
```diff
@@ -44,15 +44,6 @@ class AptPublicationSerializer(PublicationSerializer):
...
 
-    def validate(self, data):
-        """
-        Check that the publishing modes are compatible.
-        """
-        data = super().validate(data)
-        if not data["simple"] and not data["structured"]:
-            raise ValidationError("one of simple or structured publishing mode must be selected")
-        return data
-
```
quba42, line 55: There is no need to move this check into the `publish` function. I have tested all combinations with this block added back in and everything worked as expected.

The advantage of keeping it here, is that you will get the error as soon as you POST to the API, rather than starting a task that will then fail with the error.

pulp_deb/tests/functional/api/test_publish.py
```diff
@@ -143,10 +143,11 @@ def _verify_distribution(distribution, expected):
...
     [
-        DEB_PARAMS_PUB_SIMPLE,
-        DEB_PARAMS_PUB_STRUCTURED,
-        DEB_PARAMS_PUB_SIMPLE_AND_STRUCTURED,
-        DEB_PARAMS_PUB_ALL,
+        DEB_PUBLICATION_ARGS_ONLY_SIMPLE,
+        DEB_PUBLICATION_ARGS_ONLY_STRUCTURED,
+        DEB_PUBLICATION_ARGS_SIMPLE_AND_STRUCTURED,
+        DEB_PUBLICATION_ARGS_ALL,
+        {},
```
quba42, line 150: Do we somehow assert that using `{}` will produce the exact same result as using `DEB_PUBLICATION_ARGS_ONLY_STRUCTURED`?

quba42 (changes requested):
I have one changelog suggestion, one code suggestion, and one comprehension question about the tests.

Everything else LGTM.

Note: We should ideally also add a change to pulp-cli-deb to exchange `--structured` with `--no-structured` when pulp_deb >= 3.0. Otherwise it won't be possible to not publish structured using Pulp CLI with pulp_deb >= 3.0. However, this is not so critical and could be somewhat difficult to implement. I will have a go if I can figure out how.

pulp_deb/tests/functional/api/test_publish.py
hstct: No it's just testing if it works for now. Adding a check to see if the content is published structured (e.g. have it download content from a specific path) is possible and we can certainly add that at some point but maybe it's better to make it a separate test then.

pulp_deb/app/serializers/publication_serializers.py
hstct: Huh, when I tested it with this validate function it would raise the `ValidationError` every time I would publish without specifying either `structured` or `simple`. That's why I moved it to the publish task. Did you test this behavior as well?

hstct: Mh, seems like I did something wrong last time I tried that. Just double checked and you are correct it works with this validation just fine.

pulp_deb/tests/functional/api/test_publish.py
quba42: Ok.

CHANGES/858.removal
quba42: I don't think I want to mention Pulp CLI here, since we did not change Pulp CLI here and we described exactly what we changed.

quba42 (approved):
Latest version LGTM.

PR #880, opened 2023-08-30
Fix API root for the deb/copy/ endpoint for s3

### LINKED ISSUE #879: It is not possible to modify the API root for the advanced copy test case
We have discovered this via the s3 test when we tried to add test coverage here: https://github.com/pulp/pulp_deb/pull/878

After a long debugging session during pulp open floor, it was discovered that we had hard coded the API root for this endpoint, and the s3 test tries to modify the API root.

The solution is to set the API root for this endpoint as pulp_rpm already does: https://github.com/pulp/pulp_rpm/blob/main/pulp_rpm/app/urls.py

quba42:
closes #879

### DIFF
--- a/CHANGES/879.misc
+++ b/CHANGES/879.misc
@@ -1,0 +1,1 @@
+Fixed the deb/copy/ API path for scenarios that modify the API root.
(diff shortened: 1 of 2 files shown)


hstct (approved):
Very glad that this confusing error got sorted out!

PR #883, opened 2023-08-31
Continuing checksum named files for aptbyhash

### LINKED ISSUE #795: AptByHash

acheng-01:
Mitigating the Hash Sum Mismatch error by generating additional metadata files named by checksum for AptByHash. This PR continues the work of adamsanaglo and is meant to close out PR #833 

Closes #795


pulp_deb/app/__pycache__/constants.cpython-38.pyc
mdellweg, line 1: please do not add git ignored files

PR #884, opened 2023-09-01
Continuing #833 RE: Generating Checksum Named Files for AptByHash

### LINKED ISSUE #795: AptByHash

acheng-01:
This PR is to continue the work of adamsanaglo for AptByHash. It includes code changes for mitigating the Hash Sum Mismatch error by generating additional metadata files named by checksum for AptByHash, as well as changes advised by quba in the adamsanaglo's original PR.

Closes #795

### DIFF
--- a/CHANGES/795.feature
+++ b/CHANGES/795.feature
@@ -1,0 +1,2 @@
+Added an option to publish `by-hash/` files to mitigate the Hash Sum Mismatch error in debian repos as specified here: https://wiki.debian.org/DebianRepository/Format#indices_acquisition_via_hashsums_.28by-hash.29.
+Use the APT_BY_HASH setting to enable this feature.
--- a/docs/feature_overview.rst
+++ b/docs/feature_overview.rst
@@ -144,6 +144,22 @@ That way, the repository version created in the target repository, can be meanin
 We are also planning to expand the advanced copy feature with a :ref:`dependency solving <dependency_solving>` mechanism in the future.
 
 
+.. _apt_by_hash:
+
+ApyByHash
+--------------------------------------------------------------------------------
+
+AptByHash is a feature that mitigates commonplace 'Hash Sum Mismatch' errors during an 'apt-get-update'.
+It adds the checksum of the package metadata to the the packages' names.
+These files are then stored within a 'by-hash' directory within each release architecture in the specified debian repository.
+The client will then use the filename to identify the expected checksum and download a file whose name matches the checksum.
+
+Please note that this feature is disabled by default and should be enabled prior to use.
+To do this, set `APT_BY_HASH = True` in `/pulp_deb/app/settings.py`.
+
+In addition, you are responsible for setting up a reverse proxy with cache in order to cache the by-hash files.
+
+
 Roadmap and Experimental
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
--- a/pulp_deb/app/settings.py
+++ b/pulp_deb/app/settings.py
@@ -5,5 +5,6 @@ Check `Plugin Writer's Guide`_ for more details.
     http://docs.pulpproject.org/en/3.0/nightly/plugins/plugin-writer/index.html
 """
 
+APT_BY_HASH = False
 FORBIDDEN_CHECKSUM_WARNINGS = True
 FORCE_IGNORE_MISSING_PACKAGE_INDICES = False
--- a/pulp_deb/app/tasks/publishing.py
+++ b/pulp_deb/app/tasks/publishing.py
@@ -2,6 +2,7 @@ import asyncio
 import os
 import shutil
 from contextlib import suppress
+from pathlib import Path
 
 from datetime import datetime, timezone
 from debian import deb822
@@ -39,6 +40,7 @@ from pulp_deb.app.constants import (
     CHECKSUM_TYPE_MAP,
 )
 
+from pulp_deb.app.settings import APT_BY_HASH
 
 import logging
 from gettext import gettext as _
@@ -301,6 +303,23 @@ class _ComponentHelper:
                 publication=self.parent.publication, file=File(open(gz_package_index_path, "rb"))
             )
             gz_package_index.save()
+
+            # Generating metadata files using checksum
+            if APT_BY_HASH:
+                for path, index in (
+                    (package_index_path, package_index),
+                    (gz_package_index_path, gz_package_index),
+                ):
+                    for checksum in settings.ALLOWED_CONTENT_CHECKSUMS:
+                        if checksum in CHECKSUM_TYPE_MAP:
+                            hashed_index_path = _fetch_file_checksum(path, index, checksum)
+                            hashed_index = PublishedMetadata.create_from_file(
+                                publication=self.parent.publication,
+                                file=File(open(path, "rb")),
+                                relative_path=hashed_index_path,
+                            )
+                            hashed_index.save()
+
             self.parent.add_metadata(package_index)
             self.parent.add_metadata(gz_package_index)
 
@@ -340,6 +359,7 @@ class _ReleaseHelper:
         self.release["Components"] = ""  # Will be set later
         if release.description != NULL_VALUE:
             self.release["Description"] = release.description
+        self.release["Acquire-By-Hash"] = "yes" if APT_BY_HASH else "no"
 
         for checksum_type, deb_field in CHECKSUM_TYPE_MAP.items():
             if checksum_type in settings.ALLOWED_CONTENT_CHECKSUMS:
@@ -414,3 +434,10 @@ def _zip_file(file_path):
         with GzipFile(gz_file_path, "wb") as f_out:
             shutil.copyfileobj(f_in, f_out)
     return gz_file_path
+
+
+def _fetch_file_checksum(file_path, index, checksum):
+    digest = getattr(index.contentartifact_set.first().artifact, checksum)
+    checksum_type = CHECKSUM_TYPE_MAP[checksum]
+    hashed_path = Path(file_path).parents[0] / "by-hash" / checksum_type / digest
+    return hashed_path


quba42:
Regarding the failed "rady-to-ship" CI test, I believe the CI wants `closes #795` in the commit message instead of `closes pulp#795`. Automation is picky that way.

Apart from other small details, I believe the ball is back in our court to provide a final review for the feature itself. I am not sure I will get to it this week, but it is now near the top of our ToDo list.

daviddavis:
I think the commit message overall needs to be improved. It should describe the change/feature.

pulp_deb/app/tasks/publishing.py
```diff
@@ -34,11 +35,9 @@
 
 from pulp_deb.app.serializers import Package822Serializer
 
-from pulp_deb.app.constants import (
-    NO_MD5_WARNING_MESSAGE,
-    CHECKSUM_TYPE_MAP,
-)
+from pulp_deb.app.constants import NO_MD5_WARNING_MESSAGE, CHECKSUM_TYPE_MAP
```
quba42, line 38: Nitpick: Neither flake8, nor black seem to demand this change, so why are we changing it?

pulp_deb/app/tasks/publishing.py
```diff
@@ -296,10 +295,28 @@ def finish(self):
                 publication=self.parent.publication, file=File(open(package_index_path, "rb"))
             )
             package_index.save()
+
```
quba42, line 298: Nitpick: Let's not make random white space changes elsewhere in the code, that are not demanded by flake8 or black.

pulp_deb/app/settings.py
```diff
@@ -7,3 +7,5 @@
 
 FORBIDDEN_CHECKSUM_WARNINGS = True
 FORCE_IGNORE_MISSING_PACKAGE_INDICES = False
+
+APT_BY_HASH = True
```
quba42, line 11: Nitpick: For lack of a better organizing principle, let's put these in alphabetical order:
```suggestion
APT_BY_HASH = True
FORBIDDEN_CHECKSUM_WARNINGS = True
FORCE_IGNORE_MISSING_PACKAGE_INDICES = False
```

pulp_deb/app/tasks/publishing.py
```diff
@@ -413,3 +431,10 @@ def _zip_file(file_path):
         with GzipFile(gz_file_path, "wb") as f_out:
             shutil.copyfileobj(f_in, f_out)
     return gz_file_path
+
+
+def _fetch_file_checksum(file_path, index, allowed_checksum):
+    h = getattr(index.contentartifact_set.first().artifact, allowed_checksum)
+    checksum_type = CHECKSUM_TYPE_MAP[allowed_checksum]
```
quba42, line 438: Nitpick: This function really does not care if the checksum being passed in is an allowed checksum, so my naming preference would be:
```suggestion
def _fetch_file_checksum(file_path, index, checksum):
    h = getattr(index.contentartifact_set.first().artifact, checksum)
    checksum_type = CHECKSUM_TYPE_MAP[checksum]
```

pulp_deb/app/tasks/publishing.py
```diff
@@ -413,3 +431,10 @@ def _zip_file(file_path):
         with GzipFile(gz_file_path, "wb") as f_out:
             shutil.copyfileobj(f_in, f_out)
     return gz_file_path
+
+
+def _fetch_file_checksum(file_path, index, allowed_checksum):
+    h = getattr(index.contentartifact_set.first().artifact, allowed_checksum)
```
quba42, line 437: Nitpick: Can we find a better local variable name than `h`? Or am I missing something obvious that `h` stands for?

pulp_deb/app/tasks/publishing.py
```diff
@@ -296,10 +295,28 @@ def finish(self):
...
             )
             gz_package_index.save()
+
+            # Generating metadata files using checksum
+            if APT_BY_HASH:
+                for path, index in (
+                    (package_index_path, package_index),
+                    (gz_package_index_path, gz_package_index),
+                ):
+                    for allowed_checksum in settings.ALLOWED_CONTENT_CHECKSUMS:
```
quba42, line 310: Nitpick (and probably a matter of personal preference): I would probably just name the loop variable `checksum` instead of `allowed_checksum`. I feel like this reads more cleanly.

quba42 (changes requested):
First off some more nitpicks (small things of minor importance). I will post my thoughts on the feature itself separately.

Final Nitpick: The latest iteration of the commit message is now:
```
Added functionality to generate checksum named metadata files for AptByHash that was initially started by adamsanaglo.

Addressed git commit and code formatting issues pointed out by quba42

closes #795
```
- The title line should be shorter and use present tense.
- The fact that something from previous review iterations was addressed is not interesting two years from now.

My suggestion is something like:
```
Add ability to publish package indices using AptByHash format

closes #795
```

quba42:
Now my thoughts on the feature itself:

I have verified that this works as designed and adds a `by-hash/` folder with all the right files/paths next to the package indices.

It does not add the `by-hash/` folder next to the Release file.

So my final question is, how useful/complete is this feature in its current form? If I read https://wiki.ubuntu.com/AptByHash it just talks about the Release file `by-hash/` folder, and not the one we are adding here.

Is there a risk that advertising `Acquire-By-Hash: yes` in the release file, and not providing the release file `by-hash/` folder might confuse some clients?

Since I have little experience with this feature, I don't know the answer to these questions. However the answer is essential to my final question before merging:

Right now this behaviour can be enabled/disabled via the `APT_BY_HASH` setting which defaults to true. Is this feature in a state where we want to enable it for everyone by default? Is there a case for letting us toggle this feature for individual publications rather than just by global setting? (My feeling is that this would be overkill. If the feature can't hurt we should simply enable it by default, if we have doubts about the completeness, we should disable it by default, and revisit as the need arises).

@daviddavis I am guessing some of these big picture questions are as much for you as for @acheng-01  :wink:

daviddavis:
> So my final question is, how useful/complete is this feature in its current form? If I read https://wiki.ubuntu.com/AptByHash it just talks about the Release file by-hash/ folder, and not the one we are adding here.

That's interesting and I had not noticed that in the spec. If I look at the debian repo for example, it has the by-hash folder in [the directory containing the Packages](http://ftp.debian.org/debian/dists/bookworm/main/) but not [in the Release folder](http://ftp.debian.org/debian/dists/bookworm/). Ubuntu does have a by-hash folder [in the Release folder](http://archive.ubuntu.com/ubuntu/dists/lunar/). 

We should probably add these extra by-hash folders for completeness and to align with the spec but also, [I tested it out using Ubuntu](https://github.com/pulp/pulp_deb/pull/833#issuecomment-1670995800) and it seemed to work so I think this change is probably safe as is.

> Right now this behaviour can be enabled/disabled via the APT_BY_HASH setting which defaults to true. Is this feature in a state where we want to enable it for everyone by default?

I would argue that we should have this feature disabled by default. We're running Pulp behind a reverse proxy that caches various files and one of the important pieces of this feature is that the by-hash files should be cached and served by default for 3 days. Serving these files directly from Pulp doesn't provide for this so I think it makes sense to disable this feature for most users.

Moreover, I think we should probably provide accompanying docs with this feature in which we tell users that they are responsible for setting up a reverse proxy with cache in order to cache the by-hash files.

pulp_deb/app/tasks/publishing.py
acheng-01: I am not sure how this got changed (or whether I did it somehow). I will change it back.

pulp_deb/app/settings.py
acheng-01: Sure thing.

pulp_deb/app/tasks/publishing.py
acheng-01: Done.

pulp_deb/app/tasks/publishing.py
acheng-01: Done.

pulp_deb/app/tasks/publishing.py
quba42: All it takes is somebody running `black` with a different configuration, or else somebody adding another constant during development, and then removing it again manually at a later time. This PR has seen a fair amount of rebasing, squishing, changing things and then changing them back, handover to new developers, so there are lots of possibilities here. :wink:

CHANGES/795.feature
```diff
@@ -0,0 +1,2 @@
+Mitigating the Hash Sum Mismatch error in debian repos by generating additional metadata files named by checksum.
+Wiki of the feature: https://wiki.ubuntu.com/AptByHash
```
quba42, line 2: I would like to find some formulation here, that makes clear the new feature is disabled by default, and requires advanced configuration outside of Pulp. Perhaps something like:
```
Added an option to publish ``by-hash/`` files.
Use the ``APT_BY_HASH`` setting to enable this feature.
Requires additional configuration outside of Pulp, to comply with the feature specification found here:
https://wiki.ubuntu.com/AptByHash
```

quba42 (changes requested):
Everything looks very clean now.
Given what @daviddavis said, I made a suggestion regarding the changelog entry.

Given the feature is now disabled by default, I leave it up to you if you want me to merge the current state, or if you prefer to extend it first. I see the following possible follow up tasks (but I am fine if they do not materialize):

- Add the release file ``by-hash/`` folder as well.
- Add documentation on how to use this feature with reverse proxy.
- Extend a publication test to ensure that indices are in fact being published in the ``by-hash`` location.

acheng-01:
Regarding adding the release file `by-hash` folder, after taking a closer look at the specs and doing more thinking, I have decided that it is probably unnecessary to create it. To expand:

I think Adams's original PR was aimed at following [the specs for debian](https://wiki.debian.org/DebianRepository/Format#indices_acquisition_via_hashsums_.28by-hash.29) rather than for ubuntu as @quba42 had linked. In that regard, the code works and now there is a by-hash generated for each architecture's packages within the repo.

In addition to the difference in spec, I believe pulp reads directly from the Release file that already contains all the hash digests for the repository's packages; it doesn't touch the by-hash folder at all. This being the case, a by-hash folder in release wouldn't add much value other than caching.

@daviddavis do you have anything to add?

daviddavis:
I think that makes sense. When I tested this out with Ubuntu [here](https://github.com/pulp/pulp_deb/pull/833#issuecomment-1670995800), apt pulled the Packages files from the by-hash folder in the Packages folder. I'm not sure what value the Releases by-hash folder serves. At least for us, the problem was that clients were pulling Packages files that didn't match the Release so providing the Release by-hash folder seems pointless.

It might be worth stating in the docs and changelog entry that this implementation conforms to the debian spec that you linked to (and not the Ubuntu although it does appear that apt on Ubuntu does properly handle the hash files).

CHANGES/795.feature
daviddavis: I would link to this spec instead since we're implementing debian's spec.

https://wiki.debian.org/DebianRepository/Format#indices_acquisition_via_hashsums_.28by-hash.29

quba42:
@acheng-01 So if you update the release notes you can set this to "ready for review" and we can merge it.

acheng-01:
@quba42 Sorry for the delays here. I've been trying to extend one of the publishing tests to include verification that the by-hash folder is created when APT_BY_HASH is enabled without much success. Because that hashed_index_path is being saved to hashed_index, I feel that I might not be able to access it in the test without changing the APIs in pulp_deb-client or modifying some pre-existing model also within that plugin. (Please let me know if I overlooked something, overthough something, or if there is a way to test for this without altering stuff in pulp_deb-client...)

I have tested this outside of the testing environment though, and I can confirm that the by-hash folder is generated upon publishing a repo. It seems like you would be ok with me just setting this as ready to merge though, so I'll go ahead and mark this as ready for now.

quba42 (changes requested):
You need to squish all your commits together for the "Deb CI / ready-to-ship (pull_request)" test to turn green.

Apart from that this looks ready to me. The additional "Feature Overview" docs look good to me.

Regarding tests: I am not sure how simple it is to retrieve the checksum of the publication artifact, though if worst came to worst, one could simply download the published/hosted Release file, retrieve a Package index from that, download it, calculate its checksum, and then check for the presence of the relevant APT by hash path. This would work completely without use of Pulp client, but would not be very elegant. I think we can postpone this and live without that test for now.

docs/feature_overview.rst
```diff
@@ -144,6 +144,23 @@ That way, the repository version created in the target repository, can be meanin
...
+AptByHash is a feature that mitigates commonplace 'Hash Sum Mismatch' errors during an 'apt-get-update'.
+It adds the checksum of the package metadata to the the packages' names.
+These files are then stored within a 'by-hash' directory within each release architecture in the specified debian repository.
+The client will then use the filename to identify the expected checksum and download a file whose name matches the checksum.
+This follows the specification outlined in `the Debian repository wiki <https://wiki.debian.org/DebianRepository/Format#indices_acquisition_via_hashsums_.28by-hash.29>`_.
+
+Please note that this feature is disabled by default and should be enabled prior to use.
+To do this, set `APT_BY_HASH = True` in `/pulp_deb/app/settings.py`.
+
+In addition, you are responsible for setting up a reverse proxy with cache in order to cache the by-hash files.
```
mdellweg, line 161: This sounds like it's not going to cache files that were never requested before getting old.
Can we not tell pulp which of the by-hash files it needs to carry from publication to publication?

daviddavis: I mentioned persisting by-hash files in the issue but there didn’t seem much interest so I thought we’d go with the simplest solution at least initially. But we could look at persisting by-hash across publications.

There will probably be some caveats though. One that I can think of is if a publisher is distributing individual publications, we don’t have a history of which publications have been published or not. I think we’ll just have to publish all by-hash files for a given date range and potentially publish some by-hash files for publications that were never distributed. I think we can just note this in the docs though?

Also, I am assuming that we’ll want to persist by-hash files for publications in a given date range? The Ubuntu spec for instance recommends keeping by-hash files for 3 days.

mdellweg: In general, i don't think an additional by-hash file introduces any issues.
Keeping for 3 days sounds a bit like we'd want the distribution and not the publication to take care of this. That is a publication would provide its files only and the distribution would keep track on which by-hash-artifacts were active in the last 3 days.

daviddavis: To me, having the distribution handle this doesn’t make sense. If you have retain_repo_versions set for example, it’s possible the publication and its files will be cleaned up before the 3 day period ends.

mdellweg: Right. So if the distribution kept a list of recently served publications (delete cascade for the join model), we would have the proper data as long as it's available?

daviddavis: Doesn’t publication have a foreign key to repo version? So wouldn’t that also prevent the repo version and its other artifacts from being cleaned up? Also, how would the distribution no longer have the associated publication after 3 days?

quba42 (approved):
Ongoing discussions about how to develop this feature further not withstanding, I am fine with merging the current state.

docs/feature_overview.rst
daviddavis: I’m still not following. Suppose I have a retain_repo_versions set to 1 and I publish a new repo version. The distribution keeps the old repo version (N-1) since it’s still serving its files. Then how do old repo version/publication get cleaned up after 3 days?

mdellweg: It won't you'd need to retain _something_ for this feature to work. But you need to anyway, because serving the by-hash files without the pool (of the same version) is not gaining you much. So let's say, you have a retain count of X, you add new versions no more than once a day, and the distribution serves files from the current and if not found yet publications it was attached to (by increasing age up to X days).
Now you may identify one bad file in the second to latest publication, you definitely need to stop serving right away, you can still delete that publication, and the distribution would immediately forget it (because the join model cascade deletes with the publication).
If you delete the repository, all you are left with is an empty distribution with no faint memory about any publication (as today).

daviddavis: > But you need to anyway, because serving the by-hash files without the pool (of the same version) is not gaining you much.

I don’t know that this is necessarily true (at least for us). Removing packages from a repository happens rarely so usually older publications are some subset of the current publication. Moreover, Pulp is behind a caching layer which is caching packages for at least 3 days.

Your proposal sounds good to me but I don’t see any benefits for us. I think this PR as it is would be very useful for us though. The main problem we had is that when caching repo metadata, the Release/InRelease/Package files were being cached at different times which caused them to become out sync, which caused our users to hit a nasty hash mismatch error when using our repos, and in turn we basically had to not cache them (which isn’t ideal). This isn’t a problem for pulp_rpm though since it publishes it metadata files with hash digests in the file name so we can cache them for days/weeks/months/etc. Simply having pulp_deb publish the by-hash files which we can cache solves our problem.

From what I understand of your proposal, it seems like maybe it could be filed as a follow-up task? It seems like the work here would be a necessary step towards it.

daviddavis: > So let's say, you have a retain count of X

By the way, I’ve found that in practice setting the retain repo count to be very tricky if not impossible. We have some repos that are published once a year and others that are published multiple times per day. So it’s impossible to use a value for X that will guarantee that you don’t clean up a repo version/publication that was published in the last 3 days.

See my comment here for more info about this problem: https://github.com/pulp/pulpcore/issues/2705#issuecomment-1351364685

mdellweg: Sure it's possibly a second step.
I only wonder how you can be sure all the by-hash files are cached in the proxy before you switch to the next version.

daviddavis: @mdellweg can you open an issue outlining your proposal and then we can go from there?

PR #889, opened 2023-09-05
Add 3.0 to ci_update_branches

quba42:
[noissue]


quba42:
I believe this should be fine to merge just the `template_config.yml` change, and let automatic CI update to take care of actually applying it.

If this causes issues, we will not do this again in the future and apply CI changes immediately with each config change.

PR #891, opened 2023-09-06
Improve advanced copy performance

quba42:


pulp_deb/app/tasks/copy.py
```diff
@@ -27,43 +27,42 @@ def find_structured_publish_content(content, src_repo_version):
...
+    ).select_related("release_component")
+
+    # ReleaseComponents:
+    release_component_ids = set()
+    distributions = set()
+    for prc in prc_qs.iterator():
+        release_component_ids.add(prc.release_component.pk)
+        distributions.add(prc.release_component.distribution)
+
+    release_component_content_qs = source_repo_version.content.filter(pk__in=release_component_ids)
```
quba42, line 50: Alternative way to retrieve the distributions:
```
distributions = list(release_component_content_qs.values_list("distribution", flat=True).distinct())
```
Not sure if this is faster...

pulp_deb/app/tasks/copy.py
```diff
@@ -27,43 +27,42 @@ def find_structured_publish_content(content, src_repo_version):
...
+    prc_qs = PackageReleaseComponent.objects.filter(
+        pk__in=prc_content_qs.only("pk")
+    ).select_related("release_component")
+
+    # ReleaseComponents:
+    release_component_ids = set()
+    distributions = set()
+    for prc in prc_qs.iterator():
+        release_component_ids.add(prc.release_component.pk)
+        distributions.add(prc.release_component.distribution)
```
quba42, line 48: Alternative way to retrieve the distributions once we have the `release_component_content_qs`:
```
distributions = list(release_component_content_qs.values_list("distribution", flat=True).distinct())
```
Not sure if this is faster.

m-bucher (approved):
There might still be potential for further optimization

This moves querying logic into the database and that is probably a big performance improvement (verification: pending :grin: ).
So :+1: from me

hstct (approved):

PR #902, opened 2023-09-26
Improve optimize sync conditions

quba42:
This PR achieves the following:

- Do not disable optimize if we are switching from mirror to not mirror between two syncs. We can optmize in this case.
- Log additional reasons why we did not optimize in certain cases.
- Improve the readability of optimize conditions.

### DIFF
--- a/pulp_deb/app/tasks/synchronizing.py
+++ b/pulp_deb/app/tasks/synchronizing.py
@@ -176,12 +176,6 @@ def synchronize(remote_pk, repository_pk, mirror, optimize):
     repository = AptRepository.objects.get(pk=repository_pk)
     previous_repo_version = repository.latest_version()
 
-    # TODO: The optimize feature currently does not work in combination with mirror=True! We fall
-    # back to full sync for now, until a proper fix is ready:
-    if mirror:
-        log.info(_("Falling back to optimize=False behaviour since mirror=True is set!"))
-        optimize = False
-
     if not remote.url:
         raise ValueError(_("A remote must have a url specified to synchronize."))
 
@@ -546,7 +540,6 @@ class DebFirstStage(Stage):
         self.remote = remote
         self.optimize = optimize
         self.previous_repo_version = previous_repo_version
-        self.previous_sync_info = defaultdict(dict, previous_repo_version.info)
         self.sync_info = defaultdict()
         self.sync_info["remote_options"] = self._gen_remote_options()
         self.sync_info["sync_options"] = {
@@ -554,11 +547,23 @@ class DebFirstStage(Stage):
             "mirror": mirror,
         }
         self.parsed_url = urlparse(remote.url)
-        self.sync_options_unchanged = (
-            self.previous_sync_info["remote_options"] == self.sync_info["remote_options"]
-            and self.previous_sync_info["sync_options"]["mirror"]
-            == self.sync_info["sync_options"]["mirror"]
-        )
+        if self.optimize:
+            previous_sync_info = defaultdict(dict, self.previous_repo_version.info)
+            if not previous_sync_info:
+                log.info(_("Setting optimize=False since there is no previous_sync_info."))
+                self.optimize = False
+            elif not previous_sync_info["remote_options"] == self.sync_info["remote_options"]:
+                log.info(_("Setting optimize=False since the remote options have changed."))
+                self.optimize = False
+            elif mirror and not previous_sync_info["sync_options"]["mirror"]:
+                log.info(_("Setting optimize=False since this sync switches to mirror=True."))
+                self.optimize = False
+            # TODO: https://github.com/pulp/pulp_deb/issues/631
+            if mirror:
+                log.info(_("Falling back to optimize=False behaviour since mirror=True is set!"))
+                log.info(_("See https://github.com/pulp/pulp_deb/issues/631 for more information."))
+                self.optimize = False
+                self.sync_info["sync_options"]["optimize"] = False
 
     async def run(self):
         """
@@ -618,7 +623,7 @@ class DebFirstStage(Stage):
         release_file = await self._create_unit(release_file_dc)
         if release_file is None:
             return
-        if self.optimize and self.sync_options_unchanged:
+        if self.optimize:
             previous_release_file = await _get_previous_release_file(
                 self.previous_repo_version, distribution
             )
@@ -899,7 +904,7 @@ class DebFirstStage(Stage):
             else:
                 raise NoPackageIndexFile(relative_dir=package_index_dir)
 
-        if self.optimize and self.sync_options_unchanged:
+        if self.optimize:
             previous_package_index = await _get_previous_package_index(
                 self.previous_repo_version, relative_path
             )
--- a/pulp_deb/tests/functional/api/test_sync.py
+++ b/pulp_deb/tests/functional/api/test_sync.py
@@ -303,6 +303,27 @@ def test_sync_optimize_skip_unchanged_package_index(
     assert apt_release_component_api.list(package=f"{package_href},{repo.pulp_href}").count == 1
 
 
+@pytest.mark.parallel
+def test_sync_optimize_switch_to_no_mirror(
+    deb_init_and_sync,
+):
+    """
+    Test that when syncing a repo with mirror=True, and then re-syncing that repo with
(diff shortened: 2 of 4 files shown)


hstct (commented):
This should be a test case as well. To test it we need to allow passing arguments to the `deb_sync_repository` fixture and adapt the `deb_init_and_sync` fixture to use the arguments as well. The actual test case should be relatively simple then.

quba42:
@hstct I wonder if there is some simple way to assert that certain log statements have or have not happened.

hstct:
Checking for specific log statements is somewhat messy. We can just check the task if content has been skipped or not.

hstct:
In `test_sync.py` is a helper function that checks this provided a task and the correct report code.

Here we check if the release file and the package indices have not been skipped in the sync: https://github.com/pulp/pulp_deb/blob/main/pulp_deb/tests/functional/api/test_sync.py#L226-L227

pulp_deb/tests/functional/api/test_sync.py
```diff
@@ -303,6 +303,27 @@ def test_sync_optimize_skip_unchanged_package_index(
...
+@pytest.mark.parallel
+def test_sync_optimize_switch_to_no_mirror(
+    deb_init_and_sync,
+):
+    """
+    Test that when syncing a repo with mirror=True, and then re-syncing that repo with
+    mirror=False, optimize=True, the releases will be skipped by optimize mode.
+    """
+
+    sync_args = {"mirror": True}
```
hstct, line 315: If we set the value here explicitly to `sync_args = {"optimize": False, "mirror": True}` the tests will run locally.

Otherwise it will fail. Here is the full log of the error:

```
c1149a7f71f9 pulp [1d43f02f-8e4a-4424-be4f-5c709066384b]: pulpcore.tasking.tasks:INFO: Starting task 018adab3-a8f9-706f-92f3-97d12a13348b
c1149a7f71f9 pulp [1d43f02f-8e4a-4424-be4f-5c709066384b]: pulp_deb.app.tasks.synchronizing:INFO: Setting optimize=False since the remote options have changed.
c1149a7f71f9 pulp [1d43f02f-8e4a-4424-be4f-5c709066384b]: pulpcore.tasking.tasks:INFO: Task 018adab3-a8f9-706f-92f3-97d12a13348b failed ('mirror')
c1149a7f71f9 pulp [1d43f02f-8e4a-4424-be4f-5c709066384b]: pulpcore.tasking.tasks:INFO:   File "/src/pulpcore/pulpcore/tasking/tasks.py", line 64, in _execute_task
c1149a7f71f9     result = func(*args, **kwargs)
c1149a7f71f9
c1149a7f71f9   File "/src/pulp_deb/pulp_deb/app/tasks/synchronizing.py", line 182, in synchronize
c1149a7f71f9     first_stage = DebFirstStage(remote, optimize, mirror, previous_repo_version)
c1149a7f71f9
c1149a7f71f9   File "/src/pulp_deb/pulp_deb/app/tasks/synchronizing.py", line 555, in __init__
c1149a7f71f9     if mirror and not previous_sync_info["sync_options"]["mirror"]:
c1149a7f71f9
```

pulp_deb/tests/functional/conftest.py
```diff
@@ -580,7 +582,13 @@ def deb_init_and_sync(
     """Initialize a new repository and remote and sync the content from the passed URL."""
 
     def _deb_init_and_sync(
-        repository=None, remote=None, url=None, remote_args={}, repo_args={}, return_task=False
+        repository=None,
+        remote=None,
+        url=None,
+        remote_args={},
+        repo_args={},
+        sync_args={},
```
hstct, line 590: Nitpick: this new parameter should be included in the docstring below as well.

pulp_deb/tests/functional/conftest.py
```diff
@@ -447,14 +447,16 @@ def deb_sync_repository(apt_repository_api, monitor_task):
     and returns the monitored task.
     """
 
-    def _deb_sync_repository(remote, repo):
+    def _deb_sync_repository(remote, repo, mirror=False, optimize=True):
```
hstct, line 450: Nitpick: The new parameters should be included in the docstring below as well.

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -546,19 +540,27 @@ def __init__(self, remote, optimize, mirror, previous_repo_version, *args, **kwa
...
+                self.optimize = False
+            if mirror and not previous_sync_info["sync_options"]["mirror"]:
+                log.info(_("Setting optimize=False since this sync switches to mirror=True."))
+                self.optimize = False
+            # TODO: https://github.com/pulp/pulp_deb/issues/631
+            if mirror:
+                log.info(_("Falling back to optimize=False behaviour since mirror=True is set!"))
+                log.info(_("See https://github.com/pulp/pulp_deb/issues/631 for more information."))
+                self.optimize = False
+                self.sync_info["sync_options"]["optimize"] = False
```
hstct, line 563: In the current implementation this will likely fail in the edge case that it is the first sync of the repository. See above.

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -546,19 +540,27 @@ def __init__(self, remote, optimize, mirror, previous_repo_version, *args, **kwa
...
-            self.previous_sync_info["remote_options"] == self.sync_info["remote_options"]
-            and self.previous_sync_info["sync_options"]["mirror"]
-            == self.sync_info["sync_options"]["mirror"]
-        )
+        if self.optimize:
+            previous_sync_info = defaultdict(dict, self.previous_repo_version.info)
+            if not previous_sync_info["remote_options"] == self.sync_info["remote_options"]:
+                log.info(_("Setting optimize=False since the remote options have changed."))
+                self.optimize = False
+            if mirror and not previous_sync_info["sync_options"]["mirror"]:
```
hstct, line 555: There is an edge case when a repository is synced for the first time and it is set to `mirror=True` because the `previous_sync_info` is empty and cannot resolve the nested dict condition without error.

One way to resolve this is to explicitly set the `sync_options` before in this case. Something like:

```python
if "sync_options" not in previous_sync_info:
    previous_sync_info["sync_options"] = {"mirror": None, "optimize": None}
```

Another solution would be nesting the if statements to check for this case.

pulp_deb/tests/functional/api/test_sync.py
hstct: On another look this test as it is should stay because it showed that there is an edge case that will run into an error if a repository is synced for the first time and has the mirror and optimize option enabled. See below.

pulp_deb/app/tasks/synchronizing.py
hstct, line 563: This will likely run into the same edge case as above if we manage to get here on the repositories first sync.

pulp_deb/app/tasks/synchronizing.py
quba42: What I don't understand if is why this was not a problem before? We had the same dict access before...

pulp_deb/tests/functional/conftest.py
quba42: Added.

pulp_deb/tests/functional/conftest.py
quba42: Added.

pulp_deb/tests/functional/api/test_sync.py
quba42: Agreed: The test caught a real bug in my changes. I was able to reproduce the bug via manual syncing using my branch.

pulp_deb/app/tasks/synchronizing.py
quba42: This one should be fine, since this is `self.sync_info` and not `self.previous_sync_info` and `self.sync_info` was explicitly seeded with the field a few lines further up.

pulp_deb/app/tasks/synchronizing.py
quba42: I now understood why this error was not possible before. It was only because we had the dangerous dict access hidden behind a `and` condition where the thing before the `and` would always evaluate `False` on the initial sync, thus preventing the error.

I have amended my changes to fix the issue on the PR version as well.

hstct (approved):

PR #906, opened 2023-09-28
Fix ci tests

hstct:
Due to some recent pulpcore changes we needed to adapt some tests in order to fix the CI pipelines.

See: https://github.com/pulp/pulpcore/pull/4477 https://github.com/pulp/pulpcore/pull/4494

### DIFF
--- a/pulp_deb/tests/functional/api/test_download_policies.py
+++ b/pulp_deb/tests/functional/api/test_download_policies.py
@@ -48,7 +48,7 @@ def test_lazy_sync_immediate_download_test(
 ):
     """Test whether a immediate sync after a lazy one has all artifacts available."""
     # Cleanup artifacts
-    NON_LAZY_ARTIFACT_COUNT = 14
+    NON_LAZY_ARTIFACT_COUNT = 17
 
     # Create repository and remote and sync them
     repo, remote = deb_init_and_sync(remote_args=remote_args)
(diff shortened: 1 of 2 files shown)


pulp_deb/tests/functional/api/test_download_policies.py
```diff
@@ -48,7 +48,7 @@ def test_lazy_sync_immediate_download_test(
 ):
     """Test whether a immediate sync after a lazy one has all artifacts available."""
     # Cleanup artifacts
-    NON_LAZY_ARTIFACT_COUNT = 14
+    NON_LAZY_ARTIFACT_COUNT = 17
```
quba42, line 51: Why does this change an artifact count?

hstct: I don't for certain. All I can say is it was previously 17 as well but changed it when we ditched `pulp-smash` and now that this https://github.com/pulp/pulpcore/pull/4494 change is now also adopted by `pulpcore` from `pulp-smash` it needs to be 17 again. Best guess that it is now including the three `Packages.gz` files that are not empty and have been previously not downloaded - see the pulpcore PR for a log output.

quba42: I guess I will just accept this for now. :shrug:

quba42 (approved):

quba42:
I added a backport 3.0 label, do we need further backports?

quba42:
I decided to go ahead and merge this, since broken tests are blocking everything.

hstct:
We are still using the old tests for `2.21` therefore we can't cherry-pick this commit. Here is a manual change #917

PR #912, opened 2023-10-12
Add a dedicated content filter test

hstct:
[noissue]


quba42:
Also seems related: https://github.com/pulp/pulp_deb/pull/927

hstct:
> Also seems related: #927

To test this we would need to update the `debian_update/` test data to include an updated source package and a new source package.

If we want to include this with this PR I can do that.

PR #922, opened 2023-10-26
Fix package repo uniqueness

quba42:


quba42:
I have now tested that this code works for at least some of the cases that prompted this change.
I will now start working on systematic test coverage. In particular we need to cover the following cases:

- Sync and/or upload of a package set that contains forbidden duplicates => Must raise `ValueError`
- Sync and/or upload of a package set that contains allowed duplicates => Must complete and must not discard any duplicates
- Sync and/or upload of a package that collides with a already present package => previously added colliding package must be removed from the repo version.
- Sync and/or upload of a package that is a allowed duplicate of a  already present package => both packages must remain.

pulp_deb/app/models/repository.py
```diff
@@ -117,3 +121,95 @@ class AptRepositoryReleaseServiceOverride(BaseModel):
...
+        # Now remove existing packages that are duplicates of any packages added to new_version
+        if package_qs_added.count() and content_qs_existing.count():
+            log.debug(_("Removing duplicates for type: {}".format(pulp_type)))
+
+            for batch in batch_qs(package_qs_added.values(*repo_key_fields, "sha256")):
+                find_dup_qs = models.Q()
+
+                for content_dict in batch:
+                    sha256 = content_dict.pop("sha256")
+                    item_query = models.Q(**content_dict) & ~models.Q(sha256=sha256)
```
quba42, line 182: This is most likely the critical line for performance. This change must not cause a significant performance degradation!

pulp_deb/app/models/repository.py
```diff
@@ -117,3 +121,95 @@ class AptRepositoryReleaseServiceOverride(BaseModel):
...
+                )
+                for package_fields in package_qs_added_duplicates.values(*repo_key_fields):
+                    log.debug(message.format(package_fields))
+
+            message = _(
+                "Cannot create repository version since there are newly added packages with the "
+                "same name, version, and architecture, but a different checksum. If the log level "
+                "is DEBUG, you can find a list of affected packages in the Pulp log."
+            )
+            raise ValueError(message)
```
quba42, line 171: Should pulp_deb try to recover from incoming forbidden duplicates?
Personally I say know. In the case of a sync the upstream repo is definitely not compliant with the APT repo spec. In the case of uploads, the user is trying to shoot themselves in the foot.

However, at least in the sync case, I can already hear the laments: "But this repo works for us, and we just want to sync it, can't you handle the error?" => The best thing pulp_deb could possibly do, is keep one of the forbidden duplicates at random and discard the rest or perhaps simply discard all the forbidden duplicates but sync the rest of the repo which is still a behaviour that could easily lead to hard to debug user complaints, so I still prefer the hard fail.

daviddavis:
I'm grateful that you tested out the use case where a package replaces an existing existing package and removes the PackageReleaesComponent. Thank you for that.

It sounds like the issue is that you have two packages with the same NVA but different paths. Is it not possible to solve this problem by adding relative_path to repo_key_fields?

quba42:
> It sounds like the issue is that you have two packages with the same NVA but different paths. Is it not possible to solve this problem by adding relative_path to repo_key_fields?

The problem is, that a duplicate package (same package, version, architecture fields, but different relative_path) are only allowed, if they have an identical sha256 (that is, they really are the same package, that is stored in two different pool folder locations for "reasons"). If they have differing sha256, they are not allowed to exist in the same repo twice. (The spec is pretty clear on that). So the key difference between what I built here, and the pulpcore's `remove_duplicates` is the following line:
```
item_query = models.Q(**content_dict) & ~models.Q(sha256=sha256)  # For pulp_deb vs.:
item_query = Q(**content_dict) # In remove_duplicates from pulpcore
```

The same issue exists for pulpcore's `validate_duplicate_content` (used for checking the incoming content for duplicates). Some time a go I dealt with that part of the problem by not using `validate_duplicate_content` for `pulp_deb`, because it was preventing synchronization of clearly legal repositories. With hindsight, it would have been better if I had used that occasion to fully understand the problem (rather than trial and error fixing the symptoms). Not using `validate_duplicate_content` at all solved one problem but caused/masked others. This PR essentially adds a deb specific version of `validate_duplicate_content` back in. So in that way this PR makes duplicate checking more strict.

quba42:
@daviddavis I have been wondering if there is a case to be made, to be stricter for uploads than for sync. For example: It is legal for the pool folder to contain an identical copy of some package in two places, but I would not exactly call it good practice. For the sync we must be able to sync such legal repositories. However, we could choose to protect our upload users from shooting themselves in the foot, by making this forbidden for uploads. Of course, we also support mixing uploaded and synced content in a single Pulp repository so such discrimination could quickly get complicated. For now I am leaning towards not doing anything like this, but I would be interested to hear your thoughts.

daviddavis:
@quba42 in terms of our use case, we don't allow users to set the relative_path when they upload. So it's not a problem for us I don't think. But I think being stricter for uploads makes sense. And I think postponing that work until it becomes an issue also makes sense.

pulp_deb/app/models/repository.py
```diff
@@ -117,3 +121,96 @@ class AptRepositoryReleaseServiceOverride(BaseModel):
...
+                    find_dup_qs |= item_query
+
+                package_qs_duplicates = (
+                    package_obj.objects.filter(pk__in=content_qs_existing)
+                    .filter(find_dup_qs)
+                    .only("pk")
+                )
+                prc_qs_duplicates = (
+                    PackageReleaseComponent.objects.filter(pk__in=content_qs_existing)
+                    .filter(package__in=package_qs_duplicates)
```
daviddavis, line 190: So currently the code will remove any duplicate PRCs but now the code will only remove a duplicate PRC if it's attached to a duplicate package? I think reproducing the current way of cleaning up PRCs here would be preferrable for us since we are adding PRCs to repos independent of packages.

daviddavis: Actually, I think we're always adding package + PRC at the same time. However, I still think that maybe checking for dupe PRCs independently of dupe packages might be safer.

quba42: I am not sure it is possible in practice, but I am always slightly worried about the following scenario (if PRC duplicates are handled independently of package duplicates):

- Package A has PRC A
- Package B has PRC B
- Package A kicks out Package B as a duplicate.
- PRC B kicks out PRC A as a duplicate.

=> Now things are inconsistent.

Like I said, I don't think this is likely (and maybe not even possible) in practice, but I have a bad feeling about it.

I am also leaning more and more towards getting rid of PRCs entirely at some point. [My thursday PulpCon session](https://hackmd.io/KxYx73feSVeQkGD2xLNubg) is about evaluating a possible alternative to PRCs.

In the mean time I am genuinely unsure what is better here.

hstct (approved):

PR #925, opened 2023-11-01
Serve previously published artifacts for up to 3 days

### LINKED ISSUE #911: [DESIGN] A phasing effect for publications that were recently served by a distribution.
Apt-by-hash is supposed to serve files at known locations for some time (order of days) after the corresponding publication has been replaced by a newer version. This can be accomplished by a distribution that can serve multiple publications like overlays, as long as it can be assured that the publications exist sufficiently long.

> It won't you'd need to retain _something_ for this feature to work. But you need to anyway, because serving the by-hash files without the pool (of the same version) is not gaining you much. So let's say, you have a retain count of X, you add new versions no more than once a day, and the distribution serves files from the current and if not found yet publications it was attached to (by increasing age up to X days).
Now you may identify one bad file in the second to latest publication, you definitely need to stop serving right away, you can still delete that publication, and the distribution would immediately forget it (because the join model cascade deletes with the publication).
If you delete the repository, all you are left with is an empty distribution with no faint memory about any publication (as today).

_Originally posted by @mdellweg in https://github.com/pulp/pulp_deb/pull/884#discussion_r1352291358_

daviddavis:
This code requires https://github.com/pulp/pulpcore/pull/4636.

fixes #911


daviddavis:
I got the code to use a single query.

I'm still thinking we'd only like to serve the by-hash files (but not other old artifacts) for 3 days. Any objection to me adding a `PUBLICATION_CACHE_PATH_REGEX` setting?

pulp_deb/app/models/publication.py
```diff
@@ -46,6 +60,40 @@ class AptDistribution(Distribution):
...
+            )
+            .order_by("-publication__distributedpublication__pulp_created")
+            .select_related(
+                "content_artifact",
+                "content_artifact__artifact",
+            )
+        ).first()
+
+        if pa:
+            return pa.content_artifact
```
mdellweg, line 86: Will this serve the apt-by-hash metadata files from the recent publication now?
(No need to add them to the current anymore, right?)

pulp_deb/app/models/publication.py
```diff
@@ -46,6 +60,40 @@ class AptDistribution(Distribution):
     """
 
     TYPE = "apt-distribution"
+    PUBLICATION_CACHE_DURATION = timedelta(days=3)
```
ipanova, line 63: does it make sense to make this setting configurable? not sure if global setting but at least per distribution would make sense?

pulp_deb/app/models/publication.py
daviddavis: Yes exactly. The publishing code is only adding the current by-hash metadata files to the publication so once this PR is merged, we should be good to go.

pulp_deb/app/models/publication.py
daviddavis: I think some sort of setting makes sense. It was discussed a bit on the issue. For our use case, a global setting would be easier since we'll just want to set it globally and not per distro but I don't have a strong opinion. Maybe a global setting for the default with ability to override per distro would solve all use cases.

I think though just using 3 days for this first cut makes sense. We can change it later when a user requests it. Also, the future of this feature is unclear (ie sounds like we might import it into other plugins or pulpcore) so just hardcoding it would make that easier so we wouldn't have to deal with a data migration.

pulp_deb/app/models/publication.py
```diff
@@ -46,6 +60,40 @@ class AptDistribution(Distribution):
...
+
+    @hook(AFTER_SAVE, when="publication", has_changed=True)
+    def set_distributed_publication(self):
+        if self.publication:
+            DistributedPublication(distribution=self, publication=self.publication).save()
```
mdellweg, line 68: So this adds the current distribution too, right?
And the content handler is adjusted to only look for DistributedPublication. That sounds like a nice and concise logic.
Only downside, we cannot put the expiration date on the DistributedPublication itself.
Should we use this hook to also clean records that are already older than 3 days?

mdellweg: > if as a user I increased the duration, I would probably expect Pulp to start serving older publications.

That is an interesting question. Cleanup would only be needed if that table is in danger of growing out of bounds.

daviddavis: I wouldn't think that a table growing out of bounds would be likely but I suppose anything's possible. If we cleanup records here, what's a good timeframe for retention? 30 days?

daviddavis: I went ahead and added code to cleanup DistributedPublications older than 3 days. Seems like that should work for now and we can revisit later if/when the duration can be user configured.

hstct:
Due to recent CI changes this PR needs to be rebased on top of the current main branch

daviddavis:
@hstct done.

daviddavis:
@quba42 when you say you looked for the content, I assume you were looking at the directory listing pages? The old cached content does not show up there. Rather you have to try to request it directly. One thing you could test is noting the path to a particular by-hash file, change the content, republish, and then request the by-hash file.

quba42:
@daviddavis I just keep getting 404's for everything I try (regardless if I am using curl, wget, or browser). Tried for both metadata and packages in a clean oci-env. I must be missing something obvious.

Example I tried:
```
curl http://localhost:5001/pulp/content/apt_by_hash_test/dists/pulp/upload/binary-amd64/by-hash/SHA256/74ae5140c9e318f851c9168f13dcbb2a3a66db06e65da40943203e44e4f1caf4
404: Not Found
```

daviddavis:
So I went through an example and it seemed to work for me. I started with the a repo using [the quickstart example from the docs](https://docs.pulpproject.org/pulp_deb/workflows/sync.html#quickstart-example) and then ran:

```
$ http :24816/pulp/content/quickstart-nginx-bookworm-amd64/pool/nginx/n/nginx/

<html>
<head><title>Index of /pulp/content/quickstart-nginx-bookworm-amd64/pool/nginx/n/nginx/</title></head>
<body bgcolor="white">
<h1>Index of /pulp/content/quickstart-nginx-bookworm-amd64/pool/nginx/n/nginx/</h1>
<hr><pre><a href="../">../</a>
<a href="nginx-dbg_1.24.0-1~bookworm_amd64.deb">nginx-dbg_1.24.0-1~bookworm_amd64.deb</a>                                                               02-Jan-2024 16:55  5.3 MB
<a href="nginx_1.24.0-1~bookworm_amd64.deb">nginx_1.24.0-1~bookworm_amd64.deb</a>                                                                   02-Jan-2024 16:55  898.2 kB
</pre><hr></body>
</html>

$ pulp deb content list --version 1.24.0-1~bookworm --package nginx | jq '.[].pulp_href'
"/pulp/api/v3/content/deb/packages/018ccb19-b5b9-7d9c-a28a-1d29229c3939/"

$ http :24817/pulp/api/v3/content/deb/package_release_components/ package==/pulp/api/v3/content/deb/packages/018ccb19-b5b9-7d9c-a28a-1d29229c3939/ | jq '.results[].pulp_href'
"/pulp/api/v3/content/deb/package_release_components/018ccb19-bc6d-7598-a159-9cea255062c0/"

$ pulp deb repository content modify --repository quickstart-nginx-bookworm-amd64 --remove-content '[{"pulp_href": "/pulp/api/v3/content/deb/packages/018ccb19-b5b9-7d9c-a28a-1d29229c3939/", "pulp_href": "/pulp/api/v3/content/deb/package_release_components/018ccb19-bc6d-7598-a159-9cea255062c0/"}]'
Started background task /pulp/api/v3/tasks/018ccb23-14f8-7f69-bf14-679ab7dcf903/
Done.                                                                                                                                                 

$ pulp deb publication create --repository quickstart-nginx-bookworm-amd64 

$ http :24816/pulp/content/quickstart-nginx-bookworm-amd64/pool/nginx/n/nginx/

<html>
<head><title>Index of /pulp/content/quickstart-nginx-bookworm-amd64/pool/nginx/n/nginx/</title></head>
<body bgcolor="white">
<h1>Index of /pulp/content/quickstart-nginx-bookworm-amd64/pool/nginx/n/nginx/</h1>
<hr><pre><a href="../">../</a>
<a href="nginx-dbg_1.24.0-1~bookworm_amd64.deb">nginx-dbg_1.24.0-1~bookworm_amd64.deb</a>                                                               02-Jan-2024 16:55  5.3 MB
</pre><hr></body>
</html>

$ http :24816/pulp/content/quickstart-nginx-bookworm-amd64/pool/nginx/n/nginx/nginx_1.24.0-1~bookworm_amd64.deb
HTTP/1.1 200 OK
Accept-Ranges: bytes
Content-Length: 898250
Content-Type: application/octet-stream
Date: Tue, 02 Jan 2024 17:06:39 GMT
Etag: "17a695e797582a03-db4ca"
Last-Modified: Tue, 02 Jan 2024 16:52:11 GMT
Server: Python/3.10 aiohttp/3.8.1

+-----------------------------------------+
| NOTE: binary data not shown in terminal |
+-----------------------------------------+
```

A couple things I would check. First you could check that DistributedPublications are getting populated correctly:

```
$ pulpcore-manager shell_plus
In [1]: DistributedPublication.objects.all()
Out[1]: <QuerySet [<DistributedPublication: pk=018ccb1e-48c5-7d0b-b98b-39715ea51e00>, <DistributedPublication: pk=018ccb23-6622-7a46-bde9-a74ce965bf08>]>
```

Secondly, if you can debug, I would create a breakpoint here and then try to request the package or metadata file:

<https://github.com/pulp/pulp_deb/pull/925/files#diff-44986adc414526615e0e2c0e14aad998fcebc6667efadbbfcb8e9ddff846568bR86>

It should hit the breakpoint, and `recent_dp ` and `pa` should both be not null.

pulp_deb/app/models/publication.py
```diff
@@ -47,6 +61,47 @@ class AptDistribution(Distribution):
 
     TYPE = "apt-distribution"
     SERVE_FROM_PUBLICATION = True
+    PUBLICATION_CACHE_DURATION = timedelta(days=3)
+
+    @hook(AFTER_SAVE, when="publication", has_changed=True)
```
quba42, line 66: It looks like the condition for this hook is insufficient to ensure that a `DistributedPublication` is created for all possible workflows.

In my case, when I run [the quickstart example from the docs](https://docs.pulpproject.org/pulp_deb/workflows/sync.html#quickstart-example), no `DistributedPublication` is created. If I modify the workflow to create the distribution _before_ the publication, then everything works as expected (the `DistributedPublication` is created). The problem is that I create the publication first (at which time the publication is not yet distributed, so the hook on the publication does not create a `DistributedPublication`), but the hook on the distribution only fires when publication has changed, which is not the case, because I am relying on the `repository` link, rather than a direct `publication` link. See:
```
$ pulp deb distribution list
[
  {
    "pulp_href": "/pulp/api/v3/distributions/deb/apt/018ccf09-46a2-7589-b318-9094a2b6bbe8/",
    "pulp_created": "2024-01-03T11:15:40.835141Z",
    "base_path": "quickstart-nginx-bookworm-amd642",
    "base_url": "http://localhost:5001/pulp/content/quickstart-nginx-bookworm-amd642/",
    "content_guard": null,
    "hidden": false,
    "pulp_labels": {},
    "name": "quickstart-nginx-bookworm-amd642",
    "repository": "/pulp/api/v3/repositories/deb/apt/018ccf09-03be-7e77-ad7b-00e23902aeed/",
    "publication": null
  },
]
```

quba42 (changes requested):
I think I found the problem why my workflow has not been working, see the comment below:

pulp_deb/app/models/publication.py
quba42: I am fine with adding settings once there is actually a request for them.

pulp_deb/app/models/publication.py
```diff
@@ -47,6 +61,47 @@ class AptDistribution(Distribution):
...
+    """
+    A history of previously distributed publications.
+    """
+
+    distribution = models.ForeignKey(Distribution, on_delete=models.CASCADE)
+    publication = models.ForeignKey(Publication, on_delete=models.CASCADE)
+
+    @hook(AFTER_SAVE)
+    def cleanup(self):
+        """Clean up DistributedPublications older than PUBLICATION_CACHE_DURATION."""
```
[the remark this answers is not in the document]
daviddavis: The options are [save, create, update, and delete](https://rsinger86.github.io/django-lifecycle/hooks_and_conditions/#lifecycle-moments). The last two don't make sense since DistributedPublications aren't updated and they won't normally be deleted outside this cleanup. The difference between save and create is that create only triggers if pk is null. And I generally tend to default to "after" instead of "before" but I don't see a huge benefit of either.

pulp_deb/app/models/publication.py
```diff
@@ -47,6 +61,47 @@ class AptDistribution(Distribution):
...
     SERVE_FROM_PUBLICATION = True
+    PUBLICATION_CACHE_DURATION = timedelta(days=3)
+
+    @hook(AFTER_SAVE, when="publication", has_changed=True)
+    def set_distributed_publication(self):
+        if self.publication:
+            DistributedPublication(distribution=self, publication=self.publication).save()
+
+    def content_handler(self, path):
+        recent_dp = self.distributedpublication_set.filter(
```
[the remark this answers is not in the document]
daviddavis: Yes, it's defined by django:

https://docs.djangoproject.com/en/5.0/topics/db/examples/many_to_one/

pulp_deb/app/models/publication.py
daviddavis: Ah interesting find. I think I need to add another hook then to handle this case.

pulp_deb/app/models/publication.py
```diff
@@ -47,6 +61,47 @@ class AptDistribution(Distribution):
...
     TYPE = "apt-distribution"
     SERVE_FROM_PUBLICATION = True
+    PUBLICATION_CACHE_DURATION = timedelta(days=3)
+
+    @hook(AFTER_SAVE, when="publication", has_changed=True)
+    def set_distributed_publication(self):
+        if self.publication:
+            DistributedPublication(distribution=self, publication=self.publication).save()
+
+    def content_handler(self, path):
```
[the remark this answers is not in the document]
daviddavis: pulpcore:

https://github.com/pulp/pulpcore/blob/d8239beacde38fd73fa9a1d335d14d07db8a3e3c/pulpcore/content/handler.py#L578

daviddavis:
While updating the hooks, I realized there was a problem with the design: the life of a DistributedPublication is based on when its created and not when it was last used to serve content. Which is a problem because it'll get cleaned up early. 

So I've updated the code to create the DistributedPublication so it gets created with the previous publication when a new publication is used to serve content. But this creates a new problem: the list of DistributedPublications doesn't include current publication so it serves the old content over the new/current publication content. I need to look more into this.

I also tried to look at simplifying the code since having it create DistributedPublication with the previous publication makes it quite unwieldy. My first thought was to continue creating DistributedPublication with the current publication and add an `expires_at` field but this has the same problem: the life of the DistributedPublication is based on when its created and it'll get deleted too early. So I think I need to stick with creating DistributedPublication with the previous publication.

daviddavis:
Ok, I think I've resolved the issue in the most practical/simplest way. Instead of creating a DistributedPublication for the previous publication, I create a DistributedPublication for the current publication being served by the Distribution. Then when a new publication is created (or set on the Distribution), older DistributedPublications will have their `expires_at` field set.

daviddavis:
Lint failure is unrelated to this PR. Fix is here: https://github.com/pulp/pulp_deb/pull/998.

quba42 (approved):
Everything looks pretty clean and reduced, and for the first time everything worked as expected using my chosen test workflow.

Am I forgetting anything or are we finally ready to merge? :wink:

daviddavis:
I think this is ready to merge unless anyone else has any other last minute feedback.

PR #951, opened 2023-11-19
Update CI files for branch 2.21

pulpbot:
[noissue]


quba42:
Perhaps we should simply stop updating this branch?: https://github.com/pulp/pulp_deb/pull/955

hstct:
We won't support this branch any longer.

PR #955, opened 2023-11-20
Stop supporting the 2.* branches

quba42:
[noissue]

Do we want to continue updating these?


hstct (approved):
I'm in favor since we should focus on the new structure updates introduced in 3.0. It's a clean break imo

PR #956, opened 2023-11-20
Update signing key name

hstct:
[noissue]


hstct:
I used this occasion to update the test data with the new signing key as well. Since it probably will cause some confusion later on if we don't :)

hstct:
We need to backport this as well probably best to just update alle the branches with it

pulp_deb/tests/functional/sign_deb_release.sh
```diff
@@ -6,7 +6,7 @@ RELEASE_FILE="$(/usr/bin/readlink -f $1)"
 OUTPUT_DIR="$(/usr/bin/mktemp -d)"
 DETACHED_SIGNATURE_PATH="${OUTPUT_DIR}/Release.gpg"
 INLINE_SIGNATURE_PATH="${OUTPUT_DIR}/InRelease"
-GPG_KEY_ID="Pulp QE"
+GPG_KEY_ID="pulp-fixture-signing-key"
```
mdellweg, line 9: Are you using any of the pulpcore gpgkey related fixtures? Then both names can happen to appear depending on the installed version of pulpcore.

quba42: So, for example, we would need to use one for regular tests and a different one for lowerbounds?

hstct: More likely double check onto which branches this got backported to in pulpcore. I will double check this once we are ready to merge this.

mdellweg: I would prefer extracting the nae from the same fixture providing the key.

quba42: I don't think it is worth doing:

- Others don't have a better solution either, e.g.: https://github.com/pulp/pulp_rpm/pull/3327
- We need to unbreak the tests ASAP.
- Hopefully, the need to swap out test signing keys is something that only happens once every couple of years.

PR #961, opened 2023-11-20
Have the CI manage the docs

quba42:
`ci_update_docs: true` in `template_config.yml`


docs/changes.rst
```diff
@@ -1,3 +1,5 @@
 .. _pulp_deb-changes:
 
 .. include:: ../CHANGES.rst
+
+.. include:: ../HISTORY.rst
```
quba42, line 5: We don't have this file in the pulp_deb repo, but the plugin template wants to add it here. Will this cause problems?

mdellweg: hmmm. Probably.

mdellweg: An empty file seems to be fine though.

quba42: Added an empty, appears to work.

hstct (approved):

PR #971, opened 2023-11-29
Add performance tests for pulp deb

### LINKED ISSUE #970: Include performance tests for the nightly CI
Should start as a simple implementation so we can gather some data on how long different tasks take over time.

Similar to pulp_rpm these tests should include the following operations on real world repositories:

- sync
- publish
- pulp2pulp

hstct:
closes #970


hstct:
Will just merge it and see if the nightly pipeline succeeds and adjust afterwards if needed.

PR #975, opened 2023-12-06
Revert a ci change regarding cloud storage

hstct:
[noissue]


hstct:
This is just a investigation on to why the s3 test pipeline takes 5 times as long as the other ones.

hstct:
Got fixed in plugin_template

PR #977, opened 2023-12-06
Set SERVE_FROM_PUBLICATION on AptDistribution

### LINKED ISSUE #976: pulp_deb serves unpublished content
Steps to reproduce:

1. Create a repo
2. Create a distro that points to your repo
3. Add a content unit to the repo
4. Observe by hitting the content app that the content is being distributed even though it has not been published.

daviddavis:
This prevents the content app from serving unpublished content when a repo is set on a distribution:

https://github.com/pulp/pulpcore/commit/27a0dc7424769b2d5d7e98ebc4cf9cdba6cdc149

Also, this setting will be used to protect repo versions once the repo version protection work is merged:

https://github.com/pulp/pulpcore/pull/4750

fixes #976

### DIFF
--- a/CHANGES/976.bugfix
+++ b/CHANGES/976.bugfix
@@ -1,0 +1,1 @@
+Fixed a bug where pulp_deb was serving unpublished content when distributing a repository that has content but no publications.
--- a/pulp_deb/app/models/publication.py
+++ b/pulp_deb/app/models/publication.py
@@ -46,6 +46,7 @@ class AptDistribution(Distribution):
     """
 
     TYPE = "apt-distribution"
+    SERVE_FROM_PUBLICATION = True
 
     class Meta:
         default_related_name = "%(app_label)s_%(model_name)s"


ipanova (approved):

CHANGES/976.bugfix
```diff
@@ -0,0 +1 @@
+Fixed bug where pulp_deb was serving unpublished content if repository was set for a distribution.
```
quba42, line 1: I think this describes the bug more clearly:
```suggestion
Fixed a bug where pulp_deb was serving unpublished content when distributing a repository that has content but no publications.
```

quba42 (changes requested):
Nitpick on the changelog formulation. Apart from this I am happy to merge this.

According to my local tests the fix causes affected Pulp distributions to return 404 instead of unpublished content, just like when creating a distribution that does not reference a repo (nor a publication) at all. This strikes me as the correct behavior.

PR #985, opened 2023-12-18
Fix APT_BY_HASH setting

### LINKED ISSUE #984: Setting APT_BY_HASH doesn't enable the apt-by-hash feature

daviddavis:
fixes #984

### DIFF
--- a/CHANGES/984.bugfix
+++ b/CHANGES/984.bugfix
@@ -1,0 +1,1 @@
+Fixed a bug where enabling the ``APT_BY_HASH`` setting did not enable the feature.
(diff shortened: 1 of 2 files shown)


quba42 (approved):
Not sure I 100% understand why the duplicate/separate import would cause it to not work, but it is certainly cleaner with these changes, and I have full faith that you tested both versions so I am going to approve this.

quba42:
Backporting only to 3.1, since that is the release that adds this feature.

daviddavis:
The existing code imports the setting directly from the file which means that it’s always False. This changes it so that it uses django.conf.setting object which allows settings to be overridden by the user. 

https://docs.djangoproject.com/en/5.0/topics/settings/#using-settings-in-python-code

PR #998, opened 2024-01-26
Fix black errors

daviddavis:
A new version of black was released today (24.1.0) and it looks like it changed some formatting. pulp_deb doesn't pin black so the CI lint tests are failing. Here, I upgraded my local version of black and ran it to reformat the code.

### DIFF
--- a/.ci/scripts/collect_changes.py
+++ b/.ci/scripts/collect_changes.py
@@ -19,9 +19,11 @@ tc_settings = toml.load("pyproject.toml")["tool"]["towncrier"]
 CHANGELOG_FILE = tc_settings.get("filename", "NEWS.rst")
 START_STRING = tc_settings.get(
     "start_string",
-    "<!-- towncrier release notes start -->\n"
-    if CHANGELOG_FILE.endswith(".md")
-    else ".. towncrier release notes start\n",
+    (
+        "<!-- towncrier release notes start -->\n"
+        if CHANGELOG_FILE.endswith(".md")
+        else ".. towncrier release notes start\n"
+    ),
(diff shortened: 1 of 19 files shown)


.ci/scripts/collect_changes.py
```diff
@@ -19,9 +19,11 @@
...
 START_STRING = tc_settings.get(
     "start_string",
-    "<!-- towncrier release notes start -->\n"
-    if CHANGELOG_FILE.endswith(".md")
-    else ".. towncrier release notes start\n",
+    (
+        "<!-- towncrier release notes start -->\n"
+        if CHANGELOG_FILE.endswith(".md")
+        else ".. towncrier release notes start\n"
+    ),
```
mdellweg, line 26: This change can only come via the plugin template.

.ci/scripts/schema.py
```diff
@@ -7,6 +7,7 @@
 But some pulp paths start with curly brackets e.g. {artifact_href}
 This script modifies drf-spectacular schema validation to accept slashes and curly brackets.
 """
+
```
mdellweg, line 10: Maybe this also.

mdellweg: Nope, this is not in the template...

quba42:
@daviddavis Yes, I imagine that pin will eventually be bumped and we will then need to actually fix the black errors, but until then we will use the pin from the plugin template. Unfortunately there is currently another CI error, that is preventing us from adding the pin of black. I am hoping it will soon be fixed (I think so since it appears to affect every Pulp plugin.)

PR #1006, opened 2024-02-05
Improve performance test to save test results in csv files

hstct:
[noissue]


pulp_deb/tests/performance/utils.py
```diff
@@ -0,0 +1,19 @@
...
+from datetime import datetime
+
+
+def write_csv_to_tmp(filename, header_row, content_row):
+    """Write a CSV file with a header and a single content row to the /tmp/ folder.
+
+    :param filename: The name of the csv file.
+    :param header_row: The column header names.
+    :param content_row: The line of data for the content columns."""
+    with open(f"/tmp/{filename}.csv", "w", encoding="UTF-8", newline="") as f:
```
dosas, line 14: If the parameter is named filename I would expect that I had to pass `name.csv` not `name`
maybe `filestem` or `basename` is a better name?

dosas (approved):

PR #1007, opened 2024-02-12
Fix no artifact uploads

quba42:


quba42:
I have purposely omitted this fix from the `PackageReleaseComponentSerializer` and the `SourcePackageReleaseComponentSerializer`, since those should only be created via package uploads.

quba42:
I can nolonger reproduce the issue I was trying to solve here. I think it was only present in the old Pulp version that was installed in my Forklift instance...

PR #1015, opened 2024-02-18
Update CI files for branch main

pulpbot:


quba42 (approved):
Apparently we are testing with Python 3.11 now.

mdellweg:
No, but that's an expected misconception. 3.11 is the version of python running all the ci scripts now. With that logic, we removed the option from the template-config. Pulp under test runs with whatever python version the container base image brings along. And We currently (to my knowledge) have one with 3.8 and one with 3.9. Choosing these accordingly to the declared python compatibiliy claim is the new challenge. I hope, we can soon run different base images in the 4 ci scenarios to cover more ground.

mdellweg (approved):

PR #1018, opened 2024-02-19
Add retrieve functionality

quba42:


quba42:
This currently does not work, because the UniqueTogetherValidator says no before retrieve is ever called.

quba42:
Requieres https://github.com/pulp/pulpcore/pull/5084

quba42:
The alternative to the pulpcore change appears to be to switch the serializer from `NoArtifactContentSerializer` to `NoArtifactContentUploadSerializer`. But if I do that, then the create API wants to have exactly one of `upload` or `file` parameter, and I don't want the API to need those parameters.

requirements.txt
```diff
@@ -1,6 +1,6 @@
 # All things django and asyncio are deliberately left to pulpcore
 # Example transitive requirements: asgiref, asyncio, aiohttp
-pulpcore>=3.45.1,<3.55
+pulpcore>=3.49.0,<3.55
```
quba42, line 3: We may want to create a version of this PR where all changes are within pulp_deb for the benefit of Katello and merge that first.

quba42: Creating a pulp_deb changes only version of this first, will switch to the pulpcore version in a later PR.

quba42:
I should add some tests for this!

pulp_deb/tests/functional/conftest.py
```diff
@@ -70,6 +71,12 @@ def apt_release_component_api(apt_client):
     return ContentReleaseComponentsApi(apt_client)
 
 
+@pytest.fixture(scope="session")
+def deb_release_component():
+    """Fixture for APT release API."""
+    return DebReleaseComponent
```
hstct, line 77: Ah sorry, I may have given you false info because I thought this was a separate API. Since it's just an object let's just call it directly in the `deb_release_component_factory`

See `DebAptPublication` for reference

pulp_deb/tests/functional/conftest.py
```diff
@@ -104,6 +111,31 @@ def _deb_package_factory(**kwargs):
     return _deb_package_factory
 
 
+@pytest.fixture(scope="class")
+def deb_release_component_factory(
+    deb_release_component, apt_release_component_api, gen_object_with_cleanup
+):
+    """Fixture that generates deb package with cleanup."""
+
+    def _deb_release_component_factory(**kwargs):
```
hstct, line 120: If we are calling the `DebReleaseComponent` directly here than it's best to add `component` and `distribution` to the parameters here, because they are necessary to create that object.

pulp_deb/tests/functional/conftest.py
```diff
@@ -104,6 +111,31 @@ def _deb_package_factory(**kwargs):
...
+
+        :returns: The created ReleaseComponent.
+        """
+        args = {
+            "component": kwargs.pop("component"),
+            "distribution": kwargs.pop("distribution"),
+        }
+        if "repository" in kwargs:
+            args["repository"] = kwargs.pop("repository")
+        release_component_object = deb_release_component(**args)
```
hstct, line 131: We can remove the part where we pop arguments from the kwargs and call the `DebReleaseComponent` directly here. Since repository is optional for it we can either:

- let `kwargs` handle it and don't add `kwargs` to the `gen_object_with_cleanup` 
- or we add `repository` as an optional parameter in `_deb_release_component_factory`

quba42: So, I ended up adding an optional `repository` parameter and getting rid of the `kwargs` from `gen_object_with_cleanup`.

My logic is that we don't currently use them, and we can always add them if ever we do decide to use them.

pulp_deb/tests/functional/conftest.py
```diff
@@ -104,6 +105,27 @@ def _deb_package_factory(**kwargs):
...
+
+        :returns: The created ReleaseComponent.
+        """
+        args = {
+            "component": component,
+            "distribution": distribution,
+        }
+        if repository:
+            args["repository"] = repository
+        release_component_object = DebReleaseComponent(**args)
```
hstct, line 123:
```suggestion
        release_component_object = DebReleaseComponent(
            component=component,
            distribution=distribution,
            **kwargs
        )
```

This way we can remove the args and `if repository:` block. We can also remove `repository=None` in this case from the function call and let it be handled by `**kwargs` since the latter is not really used anyway.

quba42: Done.

hstct (approved):

PR #1020, opened 2024-02-26
Suppress the warnings about python3-apt not being installed

### LINKED ISSUE #1019: Suppress the warnings about python3-apt not being installed
**Describe the bug**
All non trivial syncs cause log warnings like the following:
```
“pulpcore-worker-4[59559]: /usr/lib/python3.9/site-packages/debian/deb822.py:716: UserWarning: Parsing of Deb822 data with python3-apt’s apt_pkg was requested but this package is not importable. Is python3-apt installed?
```

This is true but irrelevant, since we can't use python3-apt unless Pulp was installed on an APT based distribution.
It often leads to user confusion when users see the warning and think it is related to their problems (it never is).

**To Reproduce**
Sync any repo, check the logs.

**Expected behavior**
No warnings about things we expect to happen always!

quba42:
closes #1019

We no longer want to see warnings like the following in the log:

pulpcore-worker: /usr/lib/python3.9/site-packages/debian/deb822.py:716: UserWarning: Parsing of Deb822 data with python3-apt’s apt_pkg was requested but this package is not importable. Is python3-apt installed?

### DIFF
--- a/CHANGES/1019.bugfix
+++ b/CHANGES/1019.bugfix
@@ -1,0 +1,1 @@
+Suppressed deb822's confusing log warning about python-apt not being installed.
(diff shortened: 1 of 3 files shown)


hstct (approved):

hstct:
That's great because the message was annoying

mdellweg (approved):

PR #1026, opened 2024-02-27
Add mirror option to optimize sync

hstct:
[noissue]


pulp_deb/app/tasks/synchronizing.py
```diff
@@ -185,6 +235,37 @@ def synchronize(remote_pk, repository_pk, mirror, optimize):
...
+    should_skip = all(skip_count) and not diff_remote_options
+
+    if should_skip:
+        log.info("No change in ReleaseFiles detected. Skipping sync.")
+        with ProgressReport(
+            message="Skipping sync (no changes for any ReleaseFile)",
+            code="sync.complete_skip.was_skipped",
+        ) as pb:
+            asyncio.run(pb.aincrement())
+        return
```
quba42, line 267: This entire block needs to move behind an `if optimize:` :wink:

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -564,12 +645,6 @@ def __init__(self, remote, optimize, mirror, previous_repo_version, *args, **kwa
             elif mirror and not previous_sync_info["sync_options"]["mirror"]:
                 log.info(_("Setting optimize=False since this sync switches to mirror=True."))
                 self.optimize = False
-            # TODO: https://github.com/pulp/pulp_deb/issues/631
-            if mirror:
-                log.info(_("Falling back to optimize=False behaviour since mirror=True is set!"))
-                log.info(_("See https://github.com/pulp/pulp_deb/issues/631 for more information."))
-                self.optimize = False
-                self.sync_info["sync_options"]["optimize"] = False
```
quba42, line 572: I guess we still want to keep this.

If we don't skip the entire sync because nothing has changed, and we are using mirror and optimize, we need to continue to not use the partial optimize.

pulp_deb/app/tasks/synchronizing.py
quba42: Actually should we make it `if optimize and mirror:`?

i.e. should the no-mirror optimize also use this new initial check, or just rely on the per release file optimize checks as before?

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -185,6 +235,37 @@ def synchronize(remote_pk, repository_pk, mirror, optimize):
...
+        else:
+            skip_count.append(False)
+
+    remote_options = gen_remote_options(remote)
+    if not previous_repo_version.info:
+        diff_remote_options = False
+    elif previous_repo_version.info["remote_options"] == remote_options:
+        diff_remote_options = False
+    else:
+        diff_remote_options = True
```
quba42, line 256: elif mirror and not previous_sync_info["sync_options"]["mirror"]:

I think we need one more check here, namely `if not previous_repo_version.info["sync_options"]["mirror"]`, then we should not skip.

Let's say we have synced a repo with `--no-mirror` several times. This means we may have accumulated packages from past upstream repo versions. We now want to get rid of those by switiching to `--mirror` sync. We don't want the optimize sync to skip our sync just because the upstream repo has not changed since our last sync.

quba42 (changes requested):
Found a few small conditional checks we still need, but most of the main use cases are working as expected. We should be able to finish this soon.

Also, feel free to add final touches like an issue and a changelog note.

CHANGES/1027.feature
```diff
@@ -0,0 +1 @@
+Allow optimize with mirror mode if nothing has changed in a distribution.
```
quba42, line 1:
```suggestion
Allow optimize with mirror mode if nothing has changed in any distribution.
```
Or maybe something like "if nothing at all has changed in the repo being synced." would be clearer for end users?

hstct: either way is fine by me

quba42 (approved):
Looking good!

PR #1054, opened 2024-03-25
Fix source package publish error

### LINKED ISSUE #1053: Publishing a source package that belongs to two dists raises IntegrityError
Steps to reproduce:
1. Create a source package
2. Add it to two dists in a repo
3. Publish the repo.

You'll get an IntegrityError. Basically the source package publish code needs something similar to this line:

https://github.com/pulp/pulp_deb/blob/106f1489120bebff4bfe012356bdbc8065d8d17b/pulp_deb/app/tasks/publishing.py#L302

daviddavis:
Handle a source package publish IntegrityError when a source package belongs to multiple dists.

fixes #1053

### DIFF
--- a/pulp_deb/tests/functional/api/test_source_package.py
+++ b/pulp_deb/tests/functional/api/test_source_package.py
@@ -1,0 +1,123 @@
+"""Tests related to source packages."""
+
+import json
+import pytest
+import re
+from uuid import uuid4
+
+from pulp_deb.tests.functional.utils import (
+    get_counts_from_content_summary,
+    get_local_package_absolute_path,
+)
+from pulpcore.client.pulpcore.exceptions import ApiException
+
+SOURCE_PACKAGE_RELPATH = "mimir_1.0.dsc"
+SOURCE_PACKAGE_SOURCE = "mimir_1.0.tar.xz"
+SOURCE_PACKAGE_PATH = "data/debian/pool/asgard/m/mimir"
+
+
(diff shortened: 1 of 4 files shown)


hstct (approved):

quba42:
@daviddavis Can you add a regression test for this? Something like:

1. Upload a binary package to two dists
2. Publish the repo

daviddavis:
@quba42 just any binary package or do you mean a source package?

quba42:
> @quba42 just any binary package or do you mean a source package?

I meant source package of course!

I found for example this one in our current test data set to work with: https://github.com/ATIX-AG/pulp_deb/tree/main/pulp_deb/tests/functional/data/debian/pool/asgard/m/mimir

pulp_deb/tests/functional/api/test_source_package.py
```diff
@@ -0,0 +1,122 @@
...
+            if artifact := _find_existing_artifact(artifacts_api_client, exc):
+                return artifact
+            else:
+                raise
+        return artifact
+
+    return _artifact_factory
+
+
+def test_upload_source_package_and_publish(
```
hstct, line 56: I think this test can be run in parallel. At least it didn't complain when I tried this locally.

hstct:
I reviewed the changes and the test. The test will fail as expected when the changes are not applied. Everything looks good to me. If we can change the test to run in parallel I'm happy to merge it.

PR #1061, opened 2024-04-04
Allow source files to publish at by-hash paths

### LINKED ISSUE #1059: Allow source files to publish at by-hash paths
Less than a year ago, Adams implemented #795 (AptByHash) that would mitigate checksum mismatches for deb packages by having those packages be published with their hash as their metadata file name. There should also be support for deb source packages being published at those by-hash paths as well. There is a [debian example](https://ftp.debian.org/debian/dists/bookworm/contrib/source/) where this is done.

acheng-01:
Refactored the previous AptByHash code to extend publishing at by-hash paths for source files.

Closes #1059

### DIFF
--- a/CHANGES/1059.feature
+++ b/CHANGES/1059.feature
@@ -1,0 +1,1 @@
+Extend publishing at by-hash paths to source files.
--- a/.github/workflows/scripts/install.sh
+++ b/.github/workflows/scripts/install.sh
@@ -82,7 +82,7 @@ VARSYAML
 
 cat >> vars/main.yaml << VARSYAML
 pulp_env: {}
-pulp_settings: {"allowed_content_checksums": ["md5", "sha1", "sha256", "sha512"], "allowed_export_paths": ["/tmp"], "allowed_import_paths": ["/tmp"]}
+pulp_settings: {"allowed_content_checksums": ["md5", "sha1", "sha256", "sha512"], "allowed_export_paths": ["/tmp"], "allowed_import_paths": ["/tmp"], "apt_by_hash": true}
 pulp_scheme: https
 pulp_default_container: ghcr.io/pulp/pulp-ci-centos9:latest
 VARSYAML
--- a/pulp_deb/app/tasks/publishing.py
+++ b/pulp_deb/app/tasks/publishing.py
@@ -365,19 +365,9 @@ class _ComponentHelper:
 
             # Generating metadata files using checksum
             if settings.APT_BY_HASH:
-                for path, index in (
-                    (package_index_path, package_index),
-                    (gz_package_index_path, gz_package_index),
-                ):
-                    for checksum in settings.ALLOWED_CONTENT_CHECKSUMS:
-                        if checksum in CHECKSUM_TYPE_MAP:
-                            hashed_index_path = _fetch_file_checksum(path, index, checksum)
-                            hashed_index = PublishedMetadata.create_from_file(
-                                publication=self.parent.publication,
-                                file=File(open(path, "rb")),
-                                relative_path=hashed_index_path,
-                            )
-                            hashed_index.save()
+                self.generate_by_hash(
+                    package_index_path, package_index, gz_package_index_path, gz_package_index
+                )
 
             self.parent.add_metadata(package_index)
             self.parent.add_metadata(gz_package_index)
@@ -394,8 +384,30 @@ class _ComponentHelper:
                 publication=self.parent.publication, file=File(open(gz_source_index_path, "rb"))
             )
             gz_source_index.save()
+
+            # Generating metadata files using checksum
+            if settings.APT_BY_HASH:
+                self.generate_by_hash(
+                    source_index_path, source_index, gz_source_index_path, gz_source_index
+                )
+
             self.parent.add_metadata(source_index)
             self.parent.add_metadata(gz_source_index)
+
+    def generate_by_hash(self, index_path, index, gz_index_path, gz_index):
+        for path, index in (
+            (index_path, index),
+            (gz_index_path, gz_index),
+        ):
+            for checksum in settings.ALLOWED_CONTENT_CHECKSUMS:
+                if checksum in CHECKSUM_TYPE_MAP:
+                    hashed_index_path = _fetch_file_checksum(path, index, checksum)
+                    hashed_index = PublishedMetadata.create_from_file(
+                        publication=self.parent.publication,
+                        file=File(open(path, "rb")),
+                        relative_path=hashed_index_path,
+                    )
+                    hashed_index.save()
 
 
 class _ReleaseHelper:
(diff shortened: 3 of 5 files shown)


acheng-01:
I've confirmed that source files have their metadata published to /source/by-hash.
In my testing I have also confirmed that the by-hash file is used upon downloading the package with apt-get.

hstct (dismissed):
Verified that this works. I'm inclined to merge this. In the end though it would be great to have tests for `apt-by-hash` in general not necessary source but all packages. But I can create an extra issue for this for now.

CHANGES/1059.feature
```diff
@@ -0,0 +1 @@
+Extend publishing at by-hash paths to source files.
```
quba42, line 1: Nitpick: Consider configuring your text editor to add that trailing newline.

I don't think it strictly matters here, but it makes for ugly diffs, cats, etc. and in most places it is more correct for a text file to end with a final trailing newline. It is a matter of setting your text editor up once.

quba42 (commented):
@acheng-01 I am going to second what @hstct has said.

Since I do not naturally use/manually test the APT by hash story, I will not notice if anything breaks unless it has test coverage, so it is really in the interest of those using the story to add that test coverage.

That being said, I am willing to merge this as is (especially since we did not have the common curtsy to give an initial review for this in a timely manner).

@acheng-01, @daviddavis (if you are involved with this): Please let us know how you want to proceed.

daviddavis:
Yea, I thought we had tests for apt-by-hash but it looks like we just added one for [caching content to be served for apt-by-hash](https://github.com/pulp/pulp_deb/pull/925).

I think adding a test for apt-by-hash makes sense. I don't have a preference if we add it in this PR or a followup PR. Also, I might suggest we add a general apt-by-hash test and not one for the source files since source files are a pain to test with.

Also, how do we enable APT_BY_HASH in the test environment. Just set it [here](https://github.com/pulp/pulp_deb/blob/e4c0e3d956d089ddb9d74ee749ed2f41f832c237/template_config.yml#L50)?

hstct:
> Also, how do we enable APT_BY_HASH in the test environment. Just set it [here](https://github.com/pulp/pulp_deb/blob/e4c0e3d956d089ddb9d74ee749ed2f41f832c237/template_config.yml#L50)?

@daviddavis Yea I think that should do it.

It might be sufficient to add a test [here](https://github.com/pulp/pulp_deb/blob/main/pulp_deb/tests/functional/api/test_download_content.py) and see if we can just download the correct files from the `by-hash` folders? That way it should be pretty simple to test both standard and source packages. The default repository in the test fixtures provides both types of packages.

CHANGES/1059.feature
acheng-01: Thanks for the suggestion!

acheng-01:
Thanks for chiming in everyone. I'll look into adding tests in within the next week or so as part of this PR!

acheng-01:
Apologies that getting this test in has taken so long, but it's finally here! Please let me know if there is anything I should change with this test - otherwise it is passing on my end. Thanks everyone for you patience :)

daviddavis:
@quba42 @hstct The new test here is passing locally for me and @acheng-01 but it seems to fail in the CI. Any ideas?

daviddavis:
My guess is that the CI scripts need to be regenerated given the change to template_config.yml but I have no idea how to do that anymore.

quba42:
> My guess is that the CI scripts need to be regenerated given the change to template_config.yml but I have no idea how to do that anymore.

What you need to do is have both pulp_deb and plugin template cloned locally. Make sure you check out this PR branch in pulp_deb. In the [pugin_template repo](https://github.com/pulp/plugin_template) make sure you check out the commit referenced [here](https://github.com/acheng-01/pulp_deb/blob/ac-by-hash-for-source-files/.github/template_gitref) (e.g.: `git checkout  2021.08.26-322-g8700c1c`). Then in the plugin_template folder run: `./plugin-template --github pulp_deb`. Go back to the pulp_deb repo and check what the plugin template has done. Add it to the PR and hopefully everything will work.

acheng-01:
@quba42 I did the following:
1. rebased my PR branch against main
2. ran the plugin-template script as you had instructed
3. pushed whatever changed into this PR

There are a lot of files changed after those steps and I'm not sure if they're correct. For some reason the gitref says unknown now. Is that supposed to happen?

acheng-01:
Never mind, @daviddavis helped me identify what I did wrong in the process. I did the template thing again and things look right now. Could use a second set of eyes to verify this though.

acheng-01:
I'm not entirely sure why this test is not working in the CI pipeline. It appears that the publication isn't created successfully somehow? The `gen_object_with_cleanup` fixture uses the `monitor_task` fixture that seems to be triggering the `PulpTaskError` for the task not being in a completed state. However, I can't tell if this means the task is skipped, canceled, or failed...

I haven't been able to repro this failure locally so it's been difficult trying to understand why this is happening. The way that this test initiates the repo, publication, and distribution should be relatively similar to the other two content tests, with the only difference being that I'm passing in "sync_sources": True for remote args. Maybe that's what's causing this test to behave differently?

Would greatly appreciate if anyone might know why this is happening :)

acheng-01:
One more find in the logs:
```
pulp [None]: aiohttp.access:INFO: 127.0.0.1 [11/Jul/2024:21:22:37 +0000] "GET /debian/dists/ragnarok/jotunheimr/source/Sources HTTP/1.1" 404 172 "-" "pulpcore/3.55.1 (cpython 3.9.19-final0, Linux x86_64) (aiohttp 3.9.5)"
pulp [None]: aiohttp.access:INFO: 127.0.0.1 [11/Jul/2024:21:22:37 +0000] "GET /debian/dists/ragnarok/jotunheimr/source/Release HTTP/1.1" 200 247 "-" "pulpcore/3.55.1 (cpython 3.9.19-final0, Linux x86_64) (aiohttp 3.9.5)"
```
I was expecting `source/Sources` to be available instead of `source/Release`. For some reason some step is pulling for both. Is there a verification somewhere that checks that all these metadata files exist? It's not immediately obvious to me how these endpoints are being checked.

daviddavis:
@acheng-01 I was able to reproduce the error locally. If you rebase your branch against main, your test fails. I believe something about [this change that was merged yesterday](https://github.com/pulp/pulp_deb/commit/b29542e8ddb547fec1c8607148fda51e65d9b99e) is causing your change to fail.

acheng-01:
[Bug fix](https://github.com/pulp/pulp_deb/pull/1116) for awareness. I implemented @daviddavis's change so that the ci tests are passing for this PR, but let me know if I should revert it.

Again, huge thanks for the assist!

hstct (approved):
Ah yea, sorry I missed that when adding the change last week. I merged @daviddavis's PR.

Thanks for implementing the test. Reviewed your changes again, they are looking good! You can revert that patch from yesterday and everything should be good to go!

acheng-01:
@hstct I did another rebase on main and it looks like all the tests are passing. Ready for merge!

hstct:
Awesome! Thanks for this contribution!

hstct (approved):

PR #1068, opened 2024-04-11
Backport 3.1: add retrieve functionality / fix source package publish

hstct:
Because of the performance tests which split the `conftest.py` and the fact that the source package publish fix uses fixtures introduced in the "add retrieve functionality" there were a lot of merge conflicts which I had to fix manually.

If we don't want to backport the "add retrieve functionality" feature we can just close this PR

### DIFF
--- a/CHANGES/1053.bugfix
+++ b/CHANGES/1053.bugfix
@@ -1,0 +1,2 @@
+Fixed a bug where an ``IntegrityError`` was raised during publish when a source package belonged to
+two dists.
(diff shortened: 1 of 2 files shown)


quba42:
Given retrieve changes the API behavior we should not add it in a bugfix release.

Does backporting the fix without the test fix our cherry-pick merge conflicts? If not, I am fine with not adding this particular bugfix to `3.1`.

hstct:
If we don't want the tests it shouldn't create a merge conflict last time I checked, so we could do it without.

hstct:
As discussed I removed the source package tests and just backport the fix. The tests are still available in 3.2 and main so I'm confident it should be fine here as well and doesn't require the retrieve feature due to some changes in the test setup

quba42 (approved):
I think it makes sense to backport this without the new tests.

PR #1069, opened 2024-04-19
Create architectures and components when creating a release

### LINKED ISSUE #1038: Retrieve functionality breaks semver, requires task monitoring
[A recent change](https://github.com/pulp/pulp_deb/commit/599104f6d9a26979317e760d93cdad4a54ab215f) was introduced that made a backward incompatible change to the API. I would think this violates [semver](https://semver.org/) as it says you must "increment MAJOR version when you make incompatible API changes". The only exception it calls out are bug fixes.

That aside, the change is problematic for us. In our system, users make a single web request to add a new release to a repo which in turn finds or creates a Release along with the necessary ReleaseArchitectures and ReleaseComponents, and then adds all the necessary objects to the repo. We don't have a background job system nor a way to monitor Pulp tasks. So getting back tasks to monitor is problematic for us.

I wonder if it would be possible to solve both these problems by either:
1. Introducing a new parameter to the create actions that would switch between the old functionality or the new retrieve functionality. Or maybe only return a task if `repository` is set and otherwise return the new architecture/component.
2. Creating new endpoints to handle this new retrieve functionality

adrianabedon:
This PR adds the following: 
1. Allows for a single api call to create release, components, and architectures. 
2. Adds retrieve functionality to the release creation endpoint. If a release already exists, then the existing release will be retrieved. 
3. The release creation endpoint did not originally return a task, which is a bug since it can add content to a repository. This PR updates the endpoint to return a task.

closes #1038

### DIFF
--- a/CHANGES/1038.feature
+++ b/CHANGES/1038.feature
@@ -1,0 +1,2 @@
+Added the ``architectures`` and ``components`` parameters to the release API in order to create releases with accompanying architectures and components in a single API call.
+Also added retrieve functionality to the release API.
--- a/docs/_scripts/structured_repo.sh
+++ b/docs/_scripts/structured_repo.sh
@@ -26,15 +26,13 @@ REPO_HREF=$(http ${PULP_URL}/pulp/api/v3/repositories/deb/apt/ name=${NAME} | jq
 TASK_HREF=$(http ${PULP_URL}/pulp/api/v3/distributions/deb/apt/ name=${NAME} base_path=${NAME} repository=${REPO_HREF} | jq -r .task)
 wait_until_task_finished ${PULP_URL}${TASK_HREF}
 
-# upload the package to create Package, ReleaseComponent, PackageReleaseComponent, and Architecture content and add it to the repo in a single action
-TASK_HREF=$(http --form ${PULP_URL}/pulp/api/v3/content/deb/packages/ file@frigg_1.0_ppc64.deb repository=${REPO_HREF} distribution=${DIST} component=${COMP} | jq -r .task)
+# Create a Release, ReleaseArchitecture, and ReleaseComponent content to set various release file fields and add them to the repo in a single action
+TASK_HREF=$(http ${PULP_URL}/pulp/api/v3/content/deb/releases/ repository=${REPO_HREF} distribution=${DIST} ${RELEASE_FILE_FIELDS[@]} architectures:='["amd64", "ppc64"]' components:=['"'${COMP}'"'] | jq -r .task)
 wait_until_task_finished ${PULP_URL}${TASK_HREF}
 
-# Also create a Release content to set various release file fields
-RELEASE_HREF=$(http ${PULP_URL}/pulp/api/v3/content/deb/releases/ distribution=${DIST} ${RELEASE_FILE_FIELDS[@]} | jq -r .pulp_href)
-
-# add our content to the repository
-TASK_HREF=$(http ${PULP_URL}${REPO_HREF}modify/ add_content_units:="[\"${RELEASE_HREF}\"]" | jq -r .task)
+# upload the package to create Package and PackageReleaseComponent content and add it to the repo in a single action
+# the ReleaseComponent and ReleaseArchitecture were created in the previous step but could have been created in this step
+TASK_HREF=$(http --form ${PULP_URL}/pulp/api/v3/content/deb/packages/ file@frigg_1.0_ppc64.deb repository=${REPO_HREF} distribution=${DIST} component=${COMP} | jq -r .task)
 wait_until_task_finished ${PULP_URL}${TASK_HREF}
 
 # publish our repo
--- a/pulp_deb/app/serializers/content_serializers.py
+++ b/pulp_deb/app/serializers/content_serializers.py
@@ -5,8 +5,15 @@ from django.conf import settings
 
 from debian import deb822, debfile
 
-from rest_framework.serializers import CharField, DictField, Field, ValidationError, Serializer
-from pulpcore.plugin.models import Artifact, CreatedResource, RemoteArtifact
+from rest_framework.serializers import (
+    CharField,
+    DictField,
+    ListField,
+    Field,
+    ValidationError,
+    Serializer,
+)
+from pulpcore.plugin.models import Artifact, Content, CreatedResource, RemoteArtifact
 from pulpcore.plugin.serializers import (
     ContentChecksumSerializer,
     MultipleArtifactContentSerializer,
@@ -721,6 +728,12 @@ class ReleaseSerializer(NoArtifactContentSerializer):
     A Serializer for Release.
     """
 
+    def get_unique_together_validators(self):
+        """
+        We do not want UniqueTogetherValidator since we have retrieve logic!
+        """
+        return []
+
     codename = CharField()
     suite = CharField()
     distribution = CharField()
@@ -728,6 +741,59 @@ class ReleaseSerializer(NoArtifactContentSerializer):
     origin = NullableCharField(required=False, allow_null=True, default=None)
     label = NullableCharField(required=False, allow_null=True, default=None)
     description = NullableCharField(required=False, allow_null=True, default=None)
+    architectures = ListField(child=CharField(), required=False)
+    components = ListField(child=CharField(), required=False)
+
+    @staticmethod
+    def _get_or_create_content_pk(model, **data):
+        content, created = model.objects.get_or_create(**data)
+        CreatedResource(content_object=content).save()
+        if not created:
+            content.touch()
+        return content.pk
+
+    def retrieve(self, validated_data):
+        """
+        If the Release already exists, retrieve it!
+        """
+        return Release.objects.filter(
+            codename=validated_data["codename"],
+            suite=validated_data["suite"],
+            distribution=validated_data["distribution"],
+            version=validated_data.get("version", NULL_VALUE),
+            origin=validated_data.get("origin", NULL_VALUE),
+            label=validated_data.get("label", NULL_VALUE),
+            description=validated_data.get("description", NULL_VALUE),
+        ).first()
+
+    def create(self, validated_data):
+        architectures = validated_data.pop("architectures", [])
+        components = validated_data.pop("components", [])
+        repository = validated_data.pop("repository", None)
+
+        content_pks = []
+
+        release = super().create(validated_data)
+        content_pks.append(release.pk)
+
+        for arch in architectures:
+            architecture_pk = self._get_or_create_content_pk(
+                ReleaseArchitecture, distribution=release.distribution, architecture=arch
+            )
+            content_pks.append(architecture_pk)
+
+        for comp in components:
+            component_pk = self._get_or_create_content_pk(
+                ReleaseComponent, distribution=release.distribution, component=comp
+            )
+            content_pks.append(component_pk)
+
+        if repository:
+            repository.cast()
+            with repository.new_version() as new_version:
+                new_version.add_content(Content.objects.filter(pk__in=content_pks))
+
+        return release
 
     class Meta(NoArtifactContentSerializer.Meta):
         model = Release
@@ -739,6 +805,8 @@ class ReleaseSerializer(NoArtifactContentSerializer):
             "origin",
             "label",
             "description",
+            "architectures",
+            "components",
         )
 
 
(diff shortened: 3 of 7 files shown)


pulp_deb/app/serializers/content_serializers.py
```diff
@@ -721,13 +728,71 @@ class ReleaseSerializer(NoArtifactContentSerializer):
...
     label = NullableCharField(required=False, allow_null=True, default=None)
     description = NullableCharField(required=False, allow_null=True, default=None)
+    architectures = ListField(child=CharField(), required=False)
+    components = ListField(child=CharField(), required=False)
+
+    @staticmethod
+    def _get_or_create_content_pk(model, **data):
+        content, created = model.objects.get_or_create(**data)
+        if created:
+            CreatedResource(content_object=content).save()
```
mdellweg, line 751:
```suggestion
        if created:
            CreatedResource(content_object=content).save()
        else:
            content.touch()
```
We may need to protect this from cleanup.

daviddavis:
The docs test failure is from this script in the docs which is expecting the release endpoint to return a release object instead of a task:

https://github.com/pulp/pulp_deb/blob/3a0ef81f6cbf876abdaf53559a3d96c7010485b9/docs/_scripts/structured_repo.sh#L34

This script is on this page:

https://docs.pulpproject.org/pulp_deb/workflows/upload.html#create-a-structured-repo-manually

daviddavis:
With your change, these lines shouldn't even be needed anymore:

https://github.com/pulp/pulp_deb/blob/3a0ef81f6cbf876abdaf53559a3d96c7010485b9/docs/_scripts/structured_repo.sh#L37C1-L39C1

CHANGES/1038.feature
```diff
@@ -0,0 +1 @@
+Create architectures and components with a release and return a task when creating a release
```
quba42, line 1: Very minor request: Could you configure your text editor to add a final trailing newline `\n` at the end of the file?

I don't think this really matters for anything, other than that it bothers my sense of order in the universe. :sweat_smile:

quba42:
@daviddavis So, since we are adding retrieve functionality (changing API behavior), if we were to adhere to semantic versioning, then we would need to call the next release after merging this 4.0.0, right?

Maybe I have wrong associations regarding X version bumps, but this feels wrong to me. I feel like an X version bump also creates the expectation that this is some kind of major release with many new features (rather than one small feature that happens to change some API behavior).

@mdellweg Do you have an opinion or intuition on this?

quba42:
Regarding the failed docs test, this should turn green once https://github.com/pulp/pulp_deb/pull/1074 is merged (and this PR rebased on top of it).

mdellweg:
> @daviddavis So, since we are adding retrieve functionality (changing API behavior), if we were to adhere to semantic versioning, then we would need to call the next release after merging this 4.0.0, right?
> 
> Maybe I have wrong associations regarding X version bumps, but this feels wrong to me. I feel like an X version bump also creates the expectation that this is some kind of major release with many new features (rather than one small feature that happens to change some API behavior).
> 
> @mdellweg Do you have an opinion or intuition on this?

I'm not an expert here. To some extend every change you do (even just bugfixes) are changing some behaviour. That's why we call it software. Because it is not completely rigid. Even "just" adding a new endpoint is normally not considered a breaking change, but if you had someone relying on /api/foo/ returning 404 you'd break them. I think it's really hard to tell where to draw the line.
In particular, adding the retrieve functionality is like stopping to return an error condition and actually doing what the user intended the server to do. Is this a breaking change? I'd say no.

mdellweg:
Having said all that, semver is not about "big changes", but only about "did you break a user?". OTOH I believe semver is designed for shared libraries. Applying it to a REST API is already quite a stretch.

adrianabedon:
I went ahead and added a test case to the PR.

docs/_scripts/structured_repo.sh
```diff
@@ -30,11 +30,8 @@ wait_until_task_finished ${PULP_URL}${TASK_HREF}
 TASK_HREF=$(http --form ${PULP_URL}/pulp/api/v3/content/deb/packages/ file@frigg_1.0_ppc64.deb repository=${REPO_HREF} distribution=${DIST} component=${COMP} | jq -r .task)
 wait_until_task_finished ${PULP_URL}${TASK_HREF}
 
-# Also create a Release content to set various release file fields
-RELEASE_HREF=$(http ${PULP_URL}/pulp/api/v3/content/deb/releases/ distribution=${DIST} ${RELEASE_FILE_FIELDS[@]} | jq -r .pulp_href)
-
-# add our content to the repository
-TASK_HREF=$(http ${PULP_URL}${REPO_HREF}modify/ add_content_units:="[\"${RELEASE_HREF}\"]" | jq -r .task)
+# Also create a Release content to set various release file fields and add it to the repo
+TASK_HREF=$(http ${PULP_URL}/pulp/api/v3/content/deb/releases/ repository=${REPO_HREF} distribution=${DIST} ${RELEASE_FILE_FIELDS[@]} | jq -r .pulp_href)
```
daviddavis, line 34:
```suggestion
TASK_HREF=$(http ${PULP_URL}/pulp/api/v3/content/deb/releases/ repository=${REPO_HREF} distribution=${DIST} ${RELEASE_FILE_FIELDS[@]} | jq -r .task)
```

adrianabedon: Good catch

quba42 (changes requested):
The code changes and test look good to me. My only suggestions relate to changelog and documentation. I will try my best to test this later today. Should fit well with some Pulp CLI work I need to do.

My apologies for the delays in getting this review written down.

quba42:
I just tested using something like the following workflow:
```
NAME='empty-metadata-repo'
REPO_HREF=$(pulp deb repository create --name=${NAME} | jq -r '.pulp_href')

RELEASE_OPTIONS=(
  distribution=custom
  codename=custom
  suite=stable
  version=1
  origin=myorigin
  label=mylabel
  description=mydescription
)

http ${PULP_URL}/pulp/api/v3/content/deb/releases/ repository=${REPO_HREF} architectures:='["amd64", "i386"]' components:='["main", "contrib", "non-free"]' "${RELEASE_OPTIONS[@]}"

pulp deb publication create --repository=${NAME}
pulp deb distribution create --name=${NAME} --base-path=${NAME} --repository=${NAME}
```
I then went and destroyed the repo again and re-added the release with only some of the components and architectures, adding the rest in a separate second API call. I noticed the following slightly odd behavior:

- The Release will always be listed in the `created_resources` of the task (regardless of whether it already exists in Pulp or not).
- Any relevant ReleaseComponents and ReleaseArchitectures will be listed in the `created_resources` of the task only if they were actually created by the API call. If they already existed within Pulp and were retrieved, they are still added to any resultant new repository version, but they are not listed in `created_resources`.

I believe this is undesirable since (in the future when we have RBAC) we don't want to leak any information about what ReleaseComponents and ReleaseArchitectures already exist within Pulp. So Ideally they should always be listed in `created_resources` irrespective of whether they already existed or were just created. Just like the Release.

To achieve this we probably need to update the task manually when we go through the retrieve case. That being said, I am not sure how easy it is to access the task, given it is created by pulpcore, and not by us.

I think we already have the same wonky behavior when uploading packages as well, so we could leave this for a follow up PR.

pulp_deb/app/serializers/content_serializers.py
```diff
@@ -721,13 +728,73 @@ class ReleaseSerializer(NoArtifactContentSerializer):
...
+        else:
+            content.touch()
```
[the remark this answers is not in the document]
quba42: @mdellweg I believe we do want to always list all relevant entities in `created_resources`, regardless of whether they were in fact created or retrieved. Is that correct? Just want to confirm with you.

CHANGES/1038.feature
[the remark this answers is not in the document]
adrianabedon: I applied these suggestions to the changelog

docs/_scripts/structured_repo.sh
```diff
@@ -30,11 +30,8 @@ wait_until_task_finished ${PULP_URL}${TASK_HREF}
 TASK_HREF=$(http --form ${PULP_URL}/pulp/api/v3/content/deb/packages/ file@frigg_1.0_ppc64.deb repository=${REPO_HREF} distribution=${DIST} component=${COMP} | jq -r .task)
 wait_until_task_finished ${PULP_URL}${TASK_HREF}
 
-# Also create a Release content to set various release file fields
-RELEASE_HREF=$(http ${PULP_URL}/pulp/api/v3/content/deb/releases/ distribution=${DIST} ${RELEASE_FILE_FIELDS[@]} | jq -r .pulp_href)
-
-# add our content to the repository
-TASK_HREF=$(http ${PULP_URL}${REPO_HREF}modify/ add_content_units:="[\"${RELEASE_HREF}\"]" | jq -r .task)
+# Also create a Release content to set various release file fields and add it to the repo
+TASK_HREF=$(http ${PULP_URL}/pulp/api/v3/content/deb/releases/ repository=${REPO_HREF} distribution=${DIST} ${RELEASE_FILE_FIELDS[@]} | jq -r .task)
```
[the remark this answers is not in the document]
adrianabedon: I changed the script to create a Release with architectures and the components before uploading the package.

pulp_deb/app/serializers/content_serializers.py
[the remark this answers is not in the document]
adrianabedon: I implemented the suggestion. I wasn't sure what the correct behavior was when originally making changes. I was basing the logic off of the package upload serializer which has a [similar function](https://github.com/pulp/pulp_deb/blob/3699b9f97ea75b06ebdcd465345c47847e20f8a3/pulp_deb/app/serializers/content_serializers.py#L250C5-L254C60).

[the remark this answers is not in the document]
mdellweg: I don't know the right way to do it.

daviddavis (approved):

docs/_scripts/structured_repo.sh
```diff
@@ -26,15 +26,13 @@ REPO_HREF=$(http ${PULP_URL}/pulp/api/v3/repositories/deb/apt/ name=${NAME} | jq
 TASK_HREF=$(http ${PULP_URL}/pulp/api/v3/distributions/deb/apt/ name=${NAME} base_path=${NAME} repository=${REPO_HREF} | jq -r .task)
 wait_until_task_finished ${PULP_URL}${TASK_HREF}
 
-# upload the package to create Package, ReleaseComponent, PackageReleaseComponent, and Architecture content and add it to the repo in a single action
-TASK_HREF=$(http --form ${PULP_URL}/pulp/api/v3/content/deb/packages/ file@frigg_1.0_ppc64.deb repository=${REPO_HREF} distribution=${DIST} component=${COMP} | jq -r .task)
+# Create a Release, ReleaseArchitecture, and ReleaseComponent content to set various release file fields and add them to the repo in a single action
+TASK_HREF=$(http ${PULP_URL}/pulp/api/v3/content/deb/releases/ repository=${REPO_HREF} distribution=${DIST} ${RELEASE_FILE_FIELDS[@]} architectures:=\'["amd64", "ppc64"]\' components:=\'["${COMP}"]\' | jq -r .task)
```
daviddavis, line 30: I think there is something wrong with this line that's causing the docs CI job to fail.

quba42: From the error:
```
error:
    argument REQUEST_ITEM: "ppc64]'" is not a valid value
```
It looks like the escapes for `architectures:=\'["amd64", "ppc64"]\'` are not quite right. Since there is no bash variable in there, maybe the `'` does not need to be escaped? Or maybe the `[` and `]` do need to be escaped? Welcome to bash escape hell!

CHANGES/1038.feature
```diff
@@ -0,0 +1,2 @@
+Added the ``architectures`` and ``components`` parameters to the release API in order to create releases with accompanying architectures and components in a single API call.
+Also added retrieve functionality to the release API.
```
quba42, line 2: Really minor nitpick, but I recommend configuring your text editor to add final trailing newline characters at the end of files. Most places want a trailing newline at the end of the file. And for most editors it is easy to configure this behavior once and then never have to think about it again.

`CHANGES/1038.removal` below also lacks the trailing newline.

adrianabedon: I added the newlines. I'll remember this for future contributions!

docs/_scripts/structured_repo.sh
adrianabedon: I definitely confused myself with the escape characters. Everything should be working now.

daviddavis (approved):

quba42 (approved):
Had one final look over the latest state and re-tested my workflow. => LGTM!

Also tested that I can make the test fail by deliberately breaking the code. => check!

Thanks for the good work, and being patient while waiting for my reviews.

PR #1083, opened 2024-05-13
Update apt-by-hash docs

daviddavis:
[noissue]

### DIFF
--- a/docs/feature_overview.rst
+++ b/docs/feature_overview.rst
@@ -153,11 +153,16 @@ AptByHash is a feature that mitigates commonplace 'Hash Sum Mismatch' errors dur
 It adds the checksum of the package metadata to the the packages' names.
 These files are then stored within a 'by-hash' directory within each release architecture in the specified debian repository.
 The client will then use the filename to identify the expected checksum and download a file whose name matches the checksum.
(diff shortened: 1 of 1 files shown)


docs/feature_overview.rst
```diff
@@ -153,11 +153,13 @@ AptByHash is a feature that mitigates commonplace 'Hash Sum Mismatch' errors dur
 It adds the checksum of the package metadata to the the packages' names.
 These files are then stored within a 'by-hash' directory within each release architecture in the specified debian repository.
 The client will then use the filename to identify the expected checksum and download a file whose name matches the checksum.
+The pulp_deb plugin will cache and continue to serve previous publications' published artifacts
+including the by-hash files for up to 3 days so that users can still fetch these files after a new
+publication is created.
```
mdellweg, line 158: That is if the publication is not removed by anything, right?

daviddavis: Yes, I'll add a note about that.

quba42 (approved):
I think all of this is correct for the latest version of pulp_deb.

hstct (approved):
LGTM we can merge this together with #1061

quba42:
I think we can merge this now, since it is already true for the current level of apt-by-hash support (without source files).

PR #1090, opened 2024-05-28
Add test for uploading the same source package twice.

hstct:
[noissue]

### DIFF
--- a/pulp_deb/tests/functional/api/test_source_package.py
+++ b/pulp_deb/tests/functional/api/test_source_package.py
@@ -121,3 +121,31 @@ def test_upload_source_package_and_publish(
     # attempt to publish the repo
     publication = deb_publication_factory(repository, structured=True)
     assert publication.repository_version == repository.latest_version_href
+
+
+def test_upload_same_source_package(
+    artifact_factory,
+    apt_source_package_api,
+    deb_source_package_factory,
+    delete_orphans_pre,
+):
+    """Test whether uploading the same source package works and that it stays unique."""
+    artifact_factory(SOURCE_PACKAGE_SOURCE)
+    artifact = artifact_factory(SOURCE_PACKAGE_RELPATH)
+    attrs = {
+        "artifact": artifact.pulp_href,
+        "relative_path": SOURCE_PACKAGE_RELPATH,
+    }
+
+    # Create the first source package and verify its attributes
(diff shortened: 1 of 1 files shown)


daviddavis (approved):

pulp_deb/tests/functional/api/test_source_package.py
```diff
@@ -121,3 +121,31 @@ def test_upload_source_package_and_publish(
...
+
+
+def test_upload_same_source_package(
+    artifact_factory,
+    apt_source_package_api,
+    deb_source_package_factory,
+    delete_orphans_pre,
+):
+    """Test whether uploading the same source package works and that it stays unique."""
+    artifact_factory(SOURCE_PACKAGE_SOURCE)
```
mdellweg, line 133: Are we using this?

pulp_deb/tests/functional/api/test_source_package.py
```diff
@@ -121,3 +121,31 @@ def test_upload_source_package_and_publish(
     # attempt to publish the repo
     publication = deb_publication_factory(repository, structured=True)
     assert publication.repository_version == repository.latest_version_href
+
+
+def test_upload_same_source_package(
+    artifact_factory,
+    apt_source_package_api,
+    deb_source_package_factory,
+    delete_orphans_pre,
```
mdellweg, line 130: If you created the package into a repository, there'd be no need for orphan cleanup and the test should be able to run with "@pytest.mark.parallel".

mdellweg (commented):
Fell free to ignore my comments and merge anyway.

pulp_deb/tests/functional/api/test_source_package.py
daviddavis: Yes, the way it works is that you upload the source(s) and then the source control file (dsc) which has the digests of the sources. The multi-artifact source package then gets created from the dsc and any source artifacts it points to.

pulp_deb/tests/functional/api/test_source_package.py
hstct: Huh, I might try to do that!

hstct: It's actually more trouble than it's worth. The testing time is still reasonable once it gets out of hand again one can look at making more tests run in parallel again.

PR #1098, opened 2024-06-18
Fix changelog after conversion from RST

mdellweg:
The warning need to be marked differently in Markdown.

[noissue]


hstct (approved):
I added this to the docs PR yesterday as well. This will unblock the CI. However I just removed the block, so this is the better solution. Will merge this and drop my commit.

PR #1101, opened 2024-06-19
Raise pulpcore requirement to 3.55+

### LINKED ISSUE #1100: Make plugin compatible with with pulpcore 3.55+

hstct:
closes #1100

### DIFF
--- a/CHANGES/1100.feature
+++ b/CHANGES/1100.feature
@@ -1,0 +1,1 @@
+Made the plugin compatible with pulpcore 3.55.0+.
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 # All things django and asyncio are deliberately left to pulpcore
 # Example transitive requirements: asgiref, asyncio, aiohttp
-pulpcore>=3.49.0,<3.55
+pulpcore>=3.49.0,<3.70
 python-debian>=0.1.44,<0.2.0
 python-gnupg>=0.5,<0.6
 jsonschema>=4.6,<5.0
--- a/pulp_deb/tests/functional/api/test_download_policies.py
+++ b/pulp_deb/tests/functional/api/test_download_policies.py
@@ -42,7 +42,7 @@ def test_download_policy(
 
 @pytest.mark.parametrize("remote_args", [{"policy": "on_demand"}, {"policy": "streamed"}])
 def test_lazy_sync_immediate_download_test(
-    artifacts_api_client,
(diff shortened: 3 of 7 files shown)


requirements.txt
```diff
@@ -1,6 +1,6 @@
 # All things django and asyncio are deliberately left to pulpcore
 # Example transitive requirements: asgiref, asyncio, aiohttp
-pulpcore>=3.49.0,<3.55
+pulpcore>=3.55.0,<3.70
```
quba42, line 3: I don't think we need to raise the floor, just the ceiling?
```suggestion
pulpcore>=3.49.0,<3.70
```

CHANGES/1100.feature
```diff
@@ -0,0 +1 @@
+Made the plugin compatible with pulpcore 3.55.0+.
```
quba42, line 1: I had a look at other plugins, and the formulations for this are all over the place.
Let's use this formulation this time (consistent with our last upper bound bump), but maybe start a followup discussion in the Pulp discourse.

mdellweg:
Oh, I think you are in for a search and replace session.
e.g. `artifacts_api_client` is now `pulpcore_bindings.ArtifactsApi`.

If used this scriptlet:
```bash
find . -name "*.py" -exec sed -i \
-e 's/content_api_client/pulpcore_bindings.ContentApi/' \
-e 's/artifacts_api_client/pulpcore_bindings.ArtifactsApi/' \
-e 's/upstream_pulp_api_client/pulpcore_bindings.UpstreamPulpsApi/' \
\{\} +
```

But you need to adjust the fixture calling in the function headers by hand.

quba42 (changes requested):

requirements.txt
quba42: pulp_rpm just raised the upper bound as well.

CHANGES/1100.feature
hstct: I just copied it from the last time we did this.

quba42: I think that is the best we can do for now. I added a comment on Discourse about this: https://discourse.pulpproject.org/t/pulpcore-3-55-is-coming-soon/1259

mdellweg: If you want it consistently, you need to automate it.

quba42 (commented):
I am good with the changes as soon as the tests turn green.

quba42 (approved):

PR #1106, opened 2024-06-23
Update CI files for branch main

pulpbot:


.github/workflows/update_ci.yml
```diff
@@ -74,7 +74,7 @@ jobs:
         with:
           fetch-depth: 0
           path: "pulp_deb"
-          ref: "3.3"
+          ref: "3.2"
```
quba42, line 77: Did the automation flip the order of entries in this file?

Should we perhaps flip the order here, and re-run the workflow to get rid of this ugly diff?
https://github.com/pulp/pulp_deb/blob/main/template_config.yml#L70

mdellweg: Well, that is a list (inherently ordered). I have no idea, why it is switching order now...
But i agree it's probably best to have it sorted from newest to oldest, because we have the "last_release_branch" first.

mdellweg: BTW, if you knew 3.3 was not needed to be supported any longer as when 3.4 emerges, you'd need not have it listed here.

quba42 (approved):

PR #1108, opened 2024-06-23
Update CI files for branch 3.2

pulpbot:


quba42:
I think for these CI-changes to work on this branch, we would need to bump the pulpcore floor to a version that has the new `pulpcore-manager openapi` command.

mdellweg:
Yes, those are all the "supported_release_branches" in pulpcore.
The only one in the supported range of this plugin is 3.49.
Maybe we should however keep 3.54 a bit longer, because that's the last before a breaking change. (no easy upgrade path.)

quba42:
Not sure how best to continue with this.

mdellweg:
I bumped the python version to 3.9. Please approve and merge only if you agree with that approach.

quba42 (approved):
Since we can't really release without an up to date CI (and the up to date CI implies raising the pulpcore floor), I am in favor of merging this.

I also want to do a dot release.

PR #1114, opened 2024-07-08
Improve queries in publish task to utilize select_related

### LINKED ISSUE #1115: Improve publish task database queries
The get queries are used in a sub optimal way which impacts the performance.

E.g. instead of looping of PackageReleaseComponents which we fetch each iteration we can get them all at once previously reducing the API calls for the same result.

hstct:
closes #1115

### DIFF
--- a/pulp_deb/app/tasks/publishing.py
+++ b/pulp_deb/app/tasks/publishing.py
@@ -95,10 +95,9 @@ def publish(
         log.warning(_(NO_MD5_WARNING_MESSAGE))
 
     repo_version = RepositoryVersion.objects.get(pk=repository_version_pk)
-    if signing_service_pk:
-        signing_service = AptReleaseSigningService.objects.get(pk=signing_service_pk)
-    else:
-        signing_service = None
+    signing_service = (
+        AptReleaseSigningService.objects.get(pk=signing_service_pk) if signing_service_pk else None
+    )
 
     log.info(
         _(
@@ -127,14 +126,13 @@ def publish(
                     release.description = repository.description
 
                 component = "all"
-                architectures = (
+                architectures = list(
                     Package.objects.filter(
                         pk__in=repo_version.content.order_by("-pulp_created"),
                     )
                     .distinct("architecture")
                     .values_list("architecture", flat=True)
                 )
-                architectures = list(architectures)
                 if "all" not in architectures:
                     architectures.append("all")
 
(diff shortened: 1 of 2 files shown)


pulp_deb/app/tasks/publishing.py
```diff
@@ -146,25 +144,29 @@ def publish(
...
+                )
+                for source_package in source_packages:
                     release_helper.components[component].add_source_package(source_package)
 
                 release_helper.finish()
 
             if structured:
+                release_components = ReleaseComponent.objects.filter(
+                    pk__in=repo_version.content.order_by("-pulp_created")
+                ).select_related("release", "distribution")
```
quba42, line 164: I am not sure this select_related does anything since release_components don't actually have any foreign keys, they just have a field. They just have two text fields `distribution` and `component`.

pulp_deb/app/tasks/publishing.py
```diff
@@ -217,34 +219,41 @@ def publish(
...
+                        signing_service=signing_service,
                     )
 
-                    for prc in PackageReleaseComponent.objects.filter(
+                    package_release_components = PackageReleaseComponent.objects.filter(
                         pk__in=repo_version.content.order_by("-pulp_created"),
-                        release_component__in=release_components,
-                    ):
+                        release_component__in=release_components_filtered,
+                    ).select_related("release_component", "package")
```
quba42, line 244: I am guessing this is where the performance improvement comes from. I am guessing before it had to do another query for each prc to retrieve the accompanying package and component!

pulp_deb/app/tasks/publishing.py
```diff
@@ -217,34 +219,41 @@ def publish(
...
-                        pk__in=repo_version.content.order_by("-pulp_created"),
-                        release_component__in=release_components,
-                    ):
+                    source_package_release_components = (
+                        SourcePackageReleaseComponent.objects.filter(
+                            pk__in=repo_version.content.order_by("-pulp_created"),
+                            release_component__in=release_components,
+                        ).select_related("release_component", "source_package")
+                    )
+                    for drc in source_package_release_components:
```
quba42, line 256: Unrelated to this PR, but do you know what "drc" is short for? wouldn't "src" or perhaps "sprc" make more sense?

quba42 (changes requested):
The changes look very clean to me, and I can see how they would dramatically improve query performance! The only thing I am not sure about is the `select_related("release", "distribution")` on ReleaseComponents. Can we double check if that actually does anything? If not it should be dropped.

Also, using `release_components_filtered` to create the list of components and the prcs looks like a bugfix to me. Was this correct before?!

pulp_deb/app/tasks/publishing.py
hstct: I'll will double check this

pulp_deb/app/tasks/publishing.py
hstct: Exactly

pulp_deb/app/tasks/publishing.py
hstct: I'm guessing it's because source packages have this file format https://wiki.debian.org/dsc

pulp_deb/app/tasks/publishing.py
hstct: technically it doesn't seem to be a performance boost. At least I couldn't find a significant difference with the select_related here or without.

However we do filter for `ReleaseComponent` two times in this conditional block without select_related.

Since it doesn't seem to have any other benefit I'm happy to revert this back

quba42 (approved):

PR #1116, opened 2024-07-12
Fix source publish bug

daviddavis:
Fix a source publishing bug that was introduced in https://github.com/pulp/pulp_deb/commit/b29542e8ddb547fec1c8607148fda51e65d9b99e where the code was looping through all release components instead of just the release components for the distribution when attempting to publish source files for a specific distribution component. This raised a key error that was found in @acheng-01's PR #1061 which added a test for source publishing.


acheng-01:
Thanks for identifying this!

hstct (approved):
Uh, oh. I missed that! Sorry for the inconvenience! Good catch!

PR #1124, opened 2024-07-25
Rename staging_docs to docs

pedro-psb:
This does:

- Remove legacy docs
- Rename staging_docs to docs

[noissue]


staging_docs/user/guides/publish.svg
quba42, line 1: Don't we still need the images for the new docs?

quba42:
I don't really know how to review this. I presume it all makes sense?

Would it perhaps make sense to put the removal and the rename into two separate commits? Might make it easier to understand the two commits individually.

mdellweg:
You can run `pulp-docs serve` or similar, when you want to preview docs from a local checkout.

staging_docs/user/guides/publish.svg
pedro-psb: Yes, but the staging_docs did had copies of the images (those were just renamed).
I will make remove/rename into two commits so its easier to see what's happening.

quba42 (approved):
Ok, the two separate commits really helped me unblock my brain: This is just doing exactly what it says in the description: Remove the still present legacy docs, and then move the new docs to the now empty location. :laughing: I also see this has already happened for other plugins.

If I want to check something in the legacy docs I will just need to go back in the history to before this change. That all LGTM.

PR #1142, opened 2024-08-26
Fixed cleanup of temporary files from tasks

### LINKED ISSUE #1141: Clean tmp files after tasks are done
**Describe the bug**
pulp_deb creates a lots of tmp files in `/tmp/` which are results of signing service data, i.e. `Release.gpg`, `InRelease` and tmp files which represent archives. They are not cleaned up properly even though it's used in context.

**To Reproduce**
Publish anything with a signing service and check `/tmp/` there are multiple dirs and files still there like this:

```
[root@a090ae583dd2 /]# ls /tmp/tmp.*
tmp.MwDk6o1WR9:
InRelease  Release.gpg

tmp.WFuoBxtKJo:
InRelease  Release.gpg
```

**Expected behavior**
The temp files do be gone after tasks are complete.

hstct:
closes #1141


pulp_deb/app/models/signing_service.py
```diff
@@ -92,6 +96,10 @@ def validate(self):
                             "service script, must end with the 'InRelease' file name!"
                         )
                         raise RuntimeError(message)
+                    files_to_cleanup.append(inline_path)  # Track file for cleanup
+                    directories_to_cleanup.append(
+                        os.path.dirname(inline_path)
```
sbernhard, line 101: should we really clean up the directory AND the file?
would it be better to create allt the temp files one specific temp directory and then ONLY remove this temp directory?

pulp_deb/app/models/signing_service.py
```diff
@@ -137,3 +147,19 @@ def validate(self):
                         if verified.pubkey_fingerprint != self.pubkey_fingerprint:
                             message = "'{}' appears to have been signed using the wrong key!"
                             raise RuntimeError(message.format(detached_path))
+
+                # Clean up the signature files
+                for file_path in files_to_cleanup:
```
sbernhard, line 152: what happens, if a exceptions is thrown. Are the files then cleaned up, too?

hstct: The exception shows that this `file_path` cannot be deleted for whatever reason. Maybe it can't find it. If there is in fact a legit tmp file than it's not deleted yes. It then moves to the next `file_path` and attempts to delete that.

pulp_deb/app/models/signing_service.py
hstct: Each task will create their own `tmp.{task-id-slug}` directory in `/tmp/`. If I don't clean them up than you end up with a lot of empty directories.

Changing it that they _all_ get created into a specific directory seems more involved.

The `inline_path` gets taken from the input:

```
        {
          "signatures": {
            "inline": "<relative_path>/InRelease",
            "detached": "<relative_path>/Release.gpg",
          }
        }
```

sbernhard: If we have one directory for each task, how about saving the tmp directories path for each task and after the task finished (or an exception occurs), remove the directory completely?

pulp_deb/app/models/signing_service.py
sbernhard: No, I mean. In the above section, where a lot of tmp files are created, there are also a lot of exceptions which will maybe thrown. In case of an exception, is then the tmp file/directory removed?

mdellweg:
To my knowledge, each task is performed in a temporary directory:
https://github.com/pulp/pulpcore/blob/main/pulpcore/tasking/_util.py#L126

Is that mechanism somehow not working?

hstct:
> To my knowledge, each task is performed in a temporary directory: https://github.com/pulp/pulpcore/blob/main/pulpcore/tasking/_util.py#L126
> 
> Is that mechanism somehow not working?

Not sure but the tmp files that are created for the signing service stuff looks like this:

```
[root@1664e82d3482 tmp]# tree tmp*
tmp.1Yg9JNMlfs
├── InRelease
└── Release.gpg
tmp.4CDioeKNzT
├── InRelease
└── Release.gpg
tmp.5w3Hljn54v
├── InRelease
└── Release.gpg
tmp.6oIwBihpBI
├── InRelease
└── Release.gpg
```

pulp_deb/app/models/signing_service.py
hstct: It should work like this now? It doesn't use the pulpcore task but creates its own temp directory. But I'm not sure I follow exactly what you mean. It's better to just delete it after the validation is done because we won't need it anymore.

mdellweg:
I do not see any new temporary directory being created here:
https://github.com/pulp/pulpcore/blob/main/pulpcore/app/models/content.py#L788

So I assume at this point the process should "be" in the temporary directory of the task.
Does the signing script itself create these new directories?

quba42:
> Does the signing script itself create these new directories?

I think so. The signing service script we use on production instances has:
```
OUTPUT_DIR="$(/usr/bin/mktemp -d)"
DETACHED_SIGNATURE_PATH="${OUTPUT_DIR}/Release.gpg"
INLINE_SIGNATURE_PATH="${OUTPUT_DIR}/InRelease"
```
It then passes the paths of the created files back to Pulp via the return value. I presume pulp_deb then turns those files into publication artifacts, and once that is done, it should cleanup the files (as well as the containing temp directory)?

hstct:
Yes the signing script does create tmpfiles this was an issue with the files created by the tests. The workaround here is that I added a log for the signing script that will be used during the cleanup in the signing service fixture

mdellweg:
I'd prefer if we can teach the signing script to write files in the current directory that then is already a temporary directory.
From the perspective of the script itself, it's not actually writing temporary files but output files in the first place they should only be considered temporary in the context of the task code.

sbernhard:
> I'd prefer if we can teach the signing script to write files in the current directory that then is already a temporary directory. From the perspective of the script itself, it's not actually writing temporary files but output files in the first place they should only be considered temporary in the context of the task code.

Agree, or the signing service should be responsible to cleanup temp files itself.

pulp_deb/app/models/signing_service.py
```diff
@@ -34,106 +34,116 @@ def validate(self):
...
+                            if test_data not in inline_data:
+                                message = (
+                                    "The inline file '{}' contains different data from the "
+                                    "original file."
+                                )
+                                raise RuntimeError(message.format(inline_path))
+
+                    # Verify Release.gpg file
+                    detached_path = signatures.get("detached")
+                    if detached_path:
```
bastian-src, line 124: Do we have to throw an error in the `else` case here, too?

pulp_deb/app/models/signing_service.py
```diff
@@ -34,106 +34,116 @@ def validate(self):
...
-                            raise RuntimeError(message.format(detached_path))
+
+                    if imported_keys[0]["fingerprint"] != self.pubkey_fingerprint:
+                        raise RuntimeError(
+                            "The signing service fingerprint does not match its public key!"
+                        )
+
+                    # Verify InRelease file
+                    inline_path = signatures.get("inline")
+                    if inline_path:
```
bastian-src, line 88: Do we have to throw an error in the `else` case here?

bastian-src (commented):
Left some comments :)
~ mostly nitpicking as I'm not so familiar with the whole process :crossed_fingers:

quba42:
> I'd prefer if we can teach the signing script to write files in the current directory that then is already a temporary directory.

Would this imply teaching pulpcore to send in an extra parameter to the script? Does [https://github.com/pulp/pulpcore/blob/6a2d120a10b7a696a1282087eb51c5112cbb8351/pulpcore/app/models/content.py#L788](the pulpcore sign method) even have a way of knowing it is being called from a task? How can it receive the temp folder path?

> Agree, or the signing service should be responsible to cleanup temp files itself.

The signing script returns what files it created as the return value. I guess manually cleaning those up, is what this PR is trying to do. I don't think the signing script could be made to clean up behind itself, since it has no way of knowing when Pulp is done with the files.

mdellweg:
> > I'd prefer if we can teach the signing script to write files in the current directory that then is already a temporary directory.
> 
> Would this imply teaching pulpcore to send in an extra parameter to the script? Does [https://github.com/pulp/pulpcore/blob/6a2d120a10b7a696a1282087eb51c5112cbb8351/pulpcore/app/models/content.py#L788](the pulpcore sign method) even have a way of knowing it is being called from a task? How can it receive the temp folder path?

As far as i see, this script should only be called in two circumstances:
1. When validating a new singing service (see [0,1])
2. When being called from a task

In both cases, I expect the CWD (current working directory) to be a temporary one. If that is not true, that's the thing we should seek to fix.

[0] https://github.com/pulp/pulpcore/blob/main/pulpcore/app/models/content.py#L903
[1] https://github.com/pulp/pulp_deb/blob/main/pulp_deb/app/models/signing_service.py#L35

quba42:
@mdellweg I will have a go at testing a signing service script that just creates the file in the working directory.

mdellweg:
> Still pondering if there might be another solution that doesn't involve a change in pulpcore.

Honestly, at this point, i think it _is_ a bug in pulpcore and should be fixed there. It must be possible to call the signing script inside of a specific directory.

hstct:
This PR now requires this pulpcore PR https://github.com/pulp/pulpcore/pull/5924

What changed is that now the `sign` and `asign` methods allow for a an optional output path which the signing script then uses. Here we let the publish create a directory inside that temp folder and this now appears to solve the issue where the content scope did not clean it up afterwards.

The sync task still unzips the Package archive from upstream and saves it in the /tmp/ directory as well. This one needs manual cleanup since the file gets used out of content scope.

With the changes all relevant tmp files should now properly be cleaned up.

hstct:
Thanks to @mdellweg I was able to achieve the same thing without relying on a pulpcore by using environment variables change see: https://github.com/pulp/pulpcore/pull/5924#issuecomment-2431187924

hstct:
@mdellweg I'm not sure, really. If I call `pwd` in the script it will also try to use the project root e.g. `/src/pulp_deb/` in the oci-env. Unless I'm forgetting something?

quba42 (approved):
I have one minor suggestion. It is a take it or leave it kind of thing, so I am not waiting with the ACK.

pulp_deb/app/tasks/publishing.py
```diff
@@ -586,3 +594,10 @@ def _batch_fetch_artifacts(packages):
     remote_artifact_dict = {artifact.sha256: artifact for artifact in remote_artifacts}
 
     return artifact_dict, remote_artifact_dict
+
+
+def _create_random_directory(path):
+    dir_name = "".join(random.choices(string.ascii_letters + string.digits, k=10))
```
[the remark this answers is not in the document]
hstct: I think it's alright to leave it. I believe this helper method can be dropped entirely when we remove simple publish so it should be alright.

PR #1153, opened 2024-09-11
Prefetch artifacts for publish

### LINKED ISSUE #1148: Prefetch Artifacts and RemoteArtifacts related to packages during publish.
The publish task experiences performance degradation due to frequent database I/O operations. For each package, it retrieves the associated artifacts and often fetches the related `RemoteArtifact` as well. In large repositories, these repeated queries can significantly slow down the process.

To improve performance, artifacts should be prefetched in bulk before adding them to `PublishedArtifact`. The artifacts, including `RemoteArtifact`, should be fetched in a single operation, minimizing the number of database queries and improving overall efficiency.

hstct:
closes #1148


pulp_deb/app/serializers/content_serializers.py
```diff
@@ -479,11 +479,11 @@ def to822(self, component=""):
...
-        try:
-            artifact = self.instance._artifacts.get()
-            artifact.touch()  # Orphan cleanup protection until we are done!
-        except Artifact.DoesNotExist:
-            artifact = RemoteArtifact.objects.filter(sha256=self.instance.sha256).first()
+        artifact = None
+        if artifact_dict and self.instance.sha256 in artifact_dict:
+            artifact = artifact_dict[self.instance.sha256]
+        elif remote_artifact_dict and self.instance.sha256 in remote_artifact_dict:
+            artifact = remote_artifact_dict[self.instance.sha256]
```
quba42, line 486: So, in the unlikely event that something exists both in artifact_dict and remote_artifact_dict, we prefer the proper artifact. If we don't have either, then artifact is `None`, which was always a possibility. Sounds good!

quba42:
The top commit LGTM. Now we just need to finish #1146 first, then this should be good to go as well.

quba42 (approved):

PR #1163, opened 2024-10-02
Remove legacy config for old docs causing issue with bumpversion script

hstct:
[noissue]


quba42 (approved):

quba42:
Do we also need this for any of our supported release branches?

PR #1166, opened 2024-10-07
Expose the sha256 for source package dsc files

daviddavis:
We'd like to be able to have a sha256 column for Source Packages like we do for packages. This would save us an extra call to the artifact endpoint.

[noissue]

### DIFF
--- a/pulp_deb/app/models/content/content.py
+++ b/pulp_deb/app/models/content/content.py
@@ -197,6 +197,11 @@ class SourcePackage(Content):
                 kwargs.pop(kw)
         super().__init__(*args, **kwargs)
 
+    @property
+    def sha256(self):
+        """Return the sha256 of the dsc file."""
+        return self.contentartifact_set.get(relative_path=self.relative_path).artifact.sha256
+
     def derived_dsc_filename(self):
         """Print a nice name for the Dsc file."""
         return "{}_{}.{}".format(self.source, self.version, self.SUFFIX)
--- a/CHANGES/+add-dsc-sha256.feature
+++ b/CHANGES/+add-dsc-sha256.feature
@@ -1,0 +1,1 @@
+Expose a sha256 column for source packages that contains the digest of the dsc file.
--- a/pulp_deb/app/serializers/content_serializers.py
+++ b/pulp_deb/app/serializers/content_serializers.py
@@ -1146,6 +1146,7 @@ class SourcePackageSerializer(MultipleArtifactContentSerializer):
         ),
         required=False,
     )
+    sha256 = CharField(help_text=_("sha256 digest of the dsc file."), read_only=True)
     format = CharField(read_only=True)
     source = CharField(read_only=True)
     binary = CharField(read_only=True)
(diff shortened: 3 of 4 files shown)


hstct (approved):
Sounds like a reasonable request. Changes look good to me!

Is there any reason why we shouldn't do this @quba42 ?

pulp_deb/app/models/content/content.py
```diff
@@ -197,6 +197,11 @@ def __init__(self, *args, **kwargs):
                 kwargs.pop(kw)
         super().__init__(*args, **kwargs)
 
+    @property
+    def sha256(self):
+        """Return the sha256 of the dsc file."""
+        return self.contentartifact_set.get(relative_path=self.relative_path).artifact.sha256
```
quba42, line 203: Not sure if this is important, but I tried to look up where `contentartifact_set` is defined, and could not find it. Am I just being blind?

quba42 (commented):
@daviddavis I do have a question: It looks like the implementation for source packages in this PR is different, from the implementation for other packages. Here you define a property that accesses the `sha256` on the associated artifact. While on the [BasePackage model](https://github.com/pulp/pulp_deb/blob/main/pulp_deb/app/models/content/content.py#L68), there is an actual sha256 DB column, that is set during sync (but not for source packages): https://github.com/pulp/pulp_deb/blob/main/pulp_deb/app/tasks/synchronizing.py#L1011.

As far as I can tell, this means for Package and InstallerPackage, sha256 is present in the DB both on the content table and the associated artifact, essentially duplicating this data. The implementation in this PR may well be the better approach, but I would like to hear your thoughts on this choice? Might it even be worth to change how Package and InstallerPackage work to be more like the implementation from this PR (not as part of this PR but at some point)?

pulp_deb/app/models/content/content.py
daviddavis: It's a method that django defines when you have a [many-to-one relationship](https://docs.djangoproject.com/en/5.1/topics/db/examples/many_to_one/).

daviddavis:
@quba42 I don't know the answer to your questions. I think maybe SourcePackage doesn't have a sha256 db column since unlike Package and InstallerPackage, SourcePackage is a multi-artifact content type so there's not a single sha256 that's directly associated with the source package itself. I think the main factor driving the design of this field though is the sync code and it's been years since I've used or worked with the sync code in Pulp.

I would say though that this change doesn't preclude adding a sha256 db column to SourcePackage or removing the sha256 db column from Package/InstallerPackage though.

daviddavis:
It looks like maybe Package has a sha256 field as it's part of [its uniqueness constraint](https://github.com/pulp/pulp_deb/blob/d98e115945eacb9a3b3919121dc45b3999bf789e/pulp_deb/app/models/content/content.py#L95). Whereas it doesn't look like SourcePackage has a uniqueness constraint (maybe it should?).

quba42 (approved):
Any possible long-term design discussions notwithstanding, I am fine with merging this quality of life improvement as is.

daviddavis:
Thanks. I've filed an issue about the missing source package index.

https://github.com/pulp/pulp_deb/issues/1169

PR #1191, opened 2024-11-20
Add utility script to detect dangling PRCs in structured content

hstct:
[noissue]


quba42:
I see the following naming issue: "orphaned content" has a well defined meaning in Pulp (content that is not in any repository version). The PRCs we are looking for here, do not meet that definition, so the term "orphan" is likely to lead people astray.

I am not sure what a better term might be, but here are some possibilities: "Repo orphaned PRC", "dangling PRC", "dangling structured content". I am open to better suggestions.

quba42 (approved):
Reviewed and merged downstream (identical change).

PR #1209, opened 2024-12-11
Rebase migrations in preparation for 3.70

### LINKED ISSUE #1204: Rebase migrations to core/0091
Plugins should update any migrations that depend on pulpcore to depend on core/0091.

See https://discourse.pulpproject.org/t/action-required-plugins-should-rebase-their-migrations/1814 for details on the process.

This action can be taken at any time, but needs to happen no later than the plugin declaring core/3.70 compatibility.

mdellweg:
fixes #1204


quba42 (approved):
LGTM.

PR #1217, opened 2025-01-16
Add permissive sync setting for less lenient validation of Release files

hstct:


hstct:
Ideally we want this to be tested. Some refactoring of the relevant class will need to happen to make unit testing possible. Which is long overdue anyway and doesn't require us to create extra test data or extensive mocking.

hstct:
oops didn't push it on the fork.

PR #1222, opened 2025-01-20
Add permissive sync setting for less lenient validation of Release files

hstct:


pulp_deb/app/tasks/synchronizing.py
```diff
@@ -1318,6 +1306,80 @@ def _save_artifact_blocking(d_artifact):
...
+def _parse_release_file_attributes(d_content, main_artifact):
+    """
+    Parse the contents of main_artifact as a 'Release' file and update
+    d_content.content.* fields accordingly (e.g. codename, suite, etc.).
+    """
+    from debian import deb822
+
+    with main_artifact.file.open("rb") as fh:
+        release_file_data = fh.read()
+    release_dict = deb822.Release(release_file_data)
```
m-bucher, line 1340: I guess we do not use `release_file_data` anywhere else.
```suggestion
        release_dict = deb822.Release(fh.read())
```

pulp_deb/app/tasks/synchronizing.py
```diff
@@ -1318,6 +1306,80 @@ def _save_artifact_blocking(d_artifact):
...
+
+    if "codename" in release_dict:
+        d_content.content.codename = release_dict["Codename"]
+    if "suite" in release_dict:
+        d_content.content.suite = release_dict["Suite"]
```
m-bucher, line 1345: As long as we do not do anything else in the `if`s, this can be simplified to
```suggestion
    d_content.content.codename = release_dict.get("Codename")
    d_content.content.suite = release_dict.get("Suite")
```
Unless `d_content.content.codename = None` is not the same as not setting it at all :thinking: 

This would be an additional change to the original code, so I am OK, if this is ignored (for now).

m-bucher (changes requested):
Some small improvement suggestions. Feel free to convince me otherwise :smile:

pulp_deb/app/tasks/synchronizing.py
hstct: > Unless d_content.content.codename = None is not the same as not setting it at all

Yea, this is the problem and will require further changes.

pulp_deb/tests/unit/test_release_file_helpers.py
```diff
@@ -0,0 +1,220 @@
...
+
+@pytest.fixture
+def mock_main_artifact(tmp_path):
+    """
+    Create a fixture that has .file.open(...) returning a real file object.
+    We'll let the test function write the file content as needed.
+    """
+    artifact = MagicMock()
+    fake_file_path = tmp_path / "Release"
+    artifact._fake_file_path = fake_file_path
```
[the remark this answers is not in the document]
hstct: I test the actual file I/O behavior here. It will create an actual tmp file with the open method. I could mock this but it might hide issues that only occur with real file operations.

m-bucher (approved):

PR #1226, opened 2025-01-24
Drop custom fields with empty values

### LINKED ISSUE #1225: Repo Package files with blank custom fields with fail to parse
**Version**
main ([a193cc1](https://github.com/pulp/pulp_deb/commit/a193cc183c788f87fa74c261cd6368d1c0253c6d))

**Describe the bug**
If there's a blank custom field in the Packages index, sync fails.

This issue prevents some repos from syncing including for example Kali Rolling and other.

From [https://packages.microsoft.com/repos/aplinux-prod/dists/xenial/main/binary-amd64/Packages](https://packages.microsoft.com/repos/aplinux-prod/dists/xenial/main/binary-amd64/Packages):

```
Package: aplinux-geneva-mdsd-azsecpack
Version: 0.1.427-release-aplinux-rc3
Architecture: amd64
--- snip ---
Build-Depends:    
^^^^^^^^^^^^^^^^^  This will error
SHA256: 5b5f95a67507ae82f797937116d7ce022480fed916def156a5a32a3b487e33af
Size: 8230
```

**To Reproduce**
```bash
pulp deb remote create --name aplinux --url "https://packages.microsoft.com/repos/aplinux-prod/" --distribution xenial --architecture amd64
pulp deb repository create --name aplinux --remote aplinux
pulp deb repository sync --name aplinux
```

**Output**
```
Started background task /pulp/api/v3/tasks/1b8ffdc3-fbb1-43d2-bb75-902c2e4a3317/
.Error: Task /pulp/api/v3/tasks/1b8ffdc3-fbb1-43d2-bb75-902c2e4a3317/ failed: '{'custom_fields': {'Build-Depends': [ErrorDetail(string='This field may not be blank.', code='blank')]}}'
```

**Expected behavior**
The sync should ignore empty custom fields as it does with known ones and should be completed and not fail.

**Additional context**
The issue #694 is identical but [daviddavis](https://github.com/daviddavis)'s proposed fix #695 got closed.

This issue has a different implementation

tokfrans03:
fixes #1225

### DIFF
--- a/CHANGES/1225.bugfix
+++ b/CHANGES/1225.bugfix
@@ -1,0 +1,1 @@
+Fix bug where sync failed if a custom field in Package index was blank.
--- a/pulp_deb/app/serializers/content_serializers.py
+++ b/pulp_deb/app/serializers/content_serializers.py
@@ -413,6 +413,15 @@ class BasePackage822Serializer(SingleArtifactContentSerializer):
             log.warning(message)
             del package_fields[key]
 
+        # Also for Custom fields
+        empty_fields = [k for k, v in custom_fields.items() if not v]
+        for key in empty_fields:
+            message = _('Dropping empty "{}" custom field from "{}" package!').format(
+                key, unique_package_name
+            )
+            log.warning(message)
+            del custom_fields[key]
+
         # Delete package fields with values of incorrect type
         if "installed_size" in package_fields:
             try:


daviddavis:
I definitely think sync breaking on empty custom fields is not desirable and ought to be fixed. Does this change also allow packages with empty custom fields to be uploaded in Pulp? If so, is that desirable?

tokfrans03:
Hello. This change does allow packages with empty custom fields to be uploaded manually, and the empty custom fields (like standard ones) gets stripped. 

The issue, i think, is the underlying ambiguity of the deb822 standard if a empty field is empty or null and if it can therefore be omitted.

Per `man deb822`:
```
Empty field values are only permitted in source package control files (debian/control). Such fields are ignored
```

The question, i think, is weather pulp should keep them anyway or ignore/omit them.

I think, that any program or user wanting to consume a field that is not present should interpret it as empty/null.

In conclusion: Merge #695 if empty fields should be kept (I haven't tested this fix)

or

Merge #1226 if they should not

hstct (approved):
Tested the change and code lgtm! Thanks for the contribution!

hstct:
To clarify we also ran into an issue where sync fails due to an empty custom field, so this definitely fixes this.

PR #1249, opened 2025-03-04
publishing: publish with release.origin if defined

dguglielmi:
If the user provided the Origin value when creating the release, it is preferable to retain this value.

### DIFF
--- a/pulp_deb/app/tasks/publishing.py
+++ b/pulp_deb/app/tasks/publishing.py
@@ -221,7 +221,7 @@ def publish(
                             distribution=release.distribution,
                             codename=release.codename,
                             suite=release.suite,
-                            origin="Pulp 3",
+                            origin="Pulp 3" if release.origin == NULL_VALUE else release.origin,
                         )
                         if repository.description:
                             release.description = repository.description


quba42:
@dguglielmi By default, if the repo contains a Release, then the origin field of that Release will be used. Only if there is no Release, or if `publish_upstream` is set to `False`, do we revert to `origin="Pulp 3",`. This is a desired fallback to legacy pulp_deb behavior under specific circumstances.

Is there any chance you have the `publish_upstream_release_fields` flag set to `False` on your repository? If yes, just set it to `True` and you will get the behavior you want.

dguglielmi:
I understand. I created a new repository, and publish_upstream_release_fields is set to false (which is strange—I need to check why it's not true by default). So I thought interacting here would be a good approach. You can go ahead and close this MR. Thanks for the guidance!

quba42:
> I understand. I created a new repository, and publish_upstream_release_fields is set to false (which is strange—I need to check why it's not true by default). 

That is strange, the default should definitely be True, see:
https://github.com/pulp/pulp_deb/blob/main/pulp_deb/app/models/repository.py#L65

PR #1262, opened 2025-04-10
Add domain support

hstct:


pulp_deb/app/models/repository.py
```diff
@@ -99,9 +99,9 @@ def initialize_new_version(self, new_version):
         metadata (which may no longer be appropriate for the new RepositoryVersion is never
         retained.
         """
-        new_version.remove_content(ReleaseFile.objects.all())
-        new_version.remove_content(PackageIndex.objects.all())
-        new_version.remove_content(InstallerFileIndex.objects.all())
+        new_version.remove_content(ReleaseFile.objects.filter(pulp_domain=get_domain_pk()))
+        new_version.remove_content(PackageIndex.objects.filter(pulp_domain=get_domain_pk()))
+        new_version.remove_content(InstallerFileIndex.objects.filter(pulp_domain=get_domain_pk()))
```
mdellweg, line 104: Not sure this is necessary. You cannot remove what never was there...

mdellweg (commented):
I didn't find anything that looks wrong.

pulp_deb/app/models/repository.py
mdellweg: Fair enough. Also I'm not saying it's "wrong" to do it.

pulp_deb/app/tasks/copy.py
gerrod3, line 1: I didn't see any in this file, but all queries that don't use `pk` or `pk__in` (or have already been prefiltered through a domain object) must include `pulp_domain=get_domain()` to ensure you get the correct results.

I see in `tasks/publishing.py` in `_batch_fetch_artifacts` the Artifact & RemoteArtifact queries need pulp_domain. Also, in `tasks/synchronizing.py` in `_save_artifact_blocking` the Artifact get query needs pulp_domain.

pulp_deb/tests/functional/api/test_domains.py
```diff
@@ -0,0 +1,357 @@
...
+        deb_cleanup_domains([domain], cleanup_repositories=True)
+
+
+@pytest.mark.parallel
+def test_cross_domain_copy_all(
+    deb_cleanup_domains,
+    deb_copy_content_domain,
+    deb_setup_domain,
+):
+    """Test attempting to copy between different domains."""
```
gerrod3, line 264: Nice test!

gerrod3 (changes requested):
Really great work! Minor two changes in the task files and I think this will be good to merge!

gerrod3 (approved):
Awesome, thanks for doing this!

Issue #1276, opened 2025-05-14
Support syncing of DEP11 files


DEiselt:
**Is your feature request related to a problem? Please describe.**
Add the syncing of DEP11 metadata files to enable the use of app stores based on Appstream like KDE Discover. This  feature was mentioned / requested in #419 already. I am creating this separate issue for separate tracking of this request.

**Describe the solution you'd like**
When syncing a Repo containing DEP11 metadata files, those files should be synced as well.

**Describe alternatives you've considered**
There is no other alternative other than "not implementing" i can think of.

**Additional context**
As the files related to DEP11 aren't big and there is not many of them, i don't think a specific flag to toggle sync of those files is necessary. It should still be noted that there is a case where no DEP11 is available at all, this needs to be checked and the sync skipped accordingly.

PR #1277, opened 2025-05-14
Added support for syncing DEP11 files

DEiselt:
This change aims to add support for syncing of DEP11 metadata files, enabling the use of application stores like KDE Discover which are based on Appstream.

closes: #1276

### DIFF
--- a/CHANGES/1276.feature
+++ b/CHANGES/1276.feature
@@ -1,0 +1,1 @@
+Added support to sync DEP11 metadata files if available in the source repo.
--- a/pulp_deb/app/tasks/publishing.py
+++ b/pulp_deb/app/tasks/publishing.py
@@ -8,7 +8,7 @@ from pathlib import Path
 
 from datetime import datetime, timezone
 from debian import deb822
-from gzip import GzipFile
+import gzip
 import tempfile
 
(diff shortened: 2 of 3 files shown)


DEiselt:
Hi @hstct @quba42 could you take a look and give some feedback if this can be included / what needs to be improved?

quba42:
@DEiselt My apologies for the lack of response so far. As always things are super busy, but I have scheduled an internal meeting with @hstct for this Wednesday to have a look at this and your other PR.

hstct (commented):
It would be really great if we could make this feature optional.

Similar to the `sync_udebs` or `sync_sources` options. https://pulpproject.org/pulp_deb/restapi/#tag/Remotes:-Apt/operation/remotes_deb_apt_create

So an option you can set on the remote like `sync_dep11`.

---

Ideally we would want to have tests for this as well, however, this is trickier since we would need to create a new or expand one of the existing [test data repositories](https://github.com/pulp/pulp_deb/tree/main/pulp_deb/tests/functional/data). I'm not sure on the top of my head if these files need to be added via a `reprepo` command or not.

We'll need to look into this, but I wanted to give some immediate feedback first!

PR #1299, opened 2025-07-02
Update Changelog

pulpbot:


hstct:
what's the reasoning for doing the YANKED changelog?

mdellweg:
I'm seeing that in this Changelog there was a warning already.
My idea was that it is an interesting piece of information when you look at the changes and this is what kind of an automation I came up with.

PR #1311, opened 2025-08-12
Bump pulpcore upper bounds to 3.100 and python min to 3.11

hstct:


quba42 (approved):
Needed to fix main branch!

PR #1312, opened 2025-08-13
Add debian alternate content source

### LINKED ISSUE #1347: Add support for Alternate Content Sources (ACS)
As a user, I want to configure [pulpcore's Alternate Content Sources (ACS)](https://pulpproject.org/pulpcore/docs/user/guides/alternate-content-sources/) with pulp_deb content so that Pulp can serve `.deb` artifacts from approved upstream mirrors when available, and fall back to Pulp-hosted content otherwise.

hstct:
closes #1347


pulp_deb/app/tasks/synchronizing.py
```diff
@@ -606,7 +620,7 @@ def __init__(self, remote, optimize, mirror, previous_repo_version, *args, **kwa
             "optimize": optimize,
             "mirror": mirror,
         }
-        self.parsed_url = urlparse(remote.url)
+        self.parsed_url = urlparse(override_url or remote.url)
```
quba42: Since we don't use the paths parameter, I don't think we need the `override_url` either, since we will always be using the remote's URL anyway. This should allow us to reduce the diff in `synchronizing.py`

PR #1321, opened 2025-08-20
Update CI files for branch 3.5

pulpbot:
Automated changes by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action


mdellweg:
Actually, @pedro-psb can you take a look at this?
We decided that all the clever logic around the storage config would never cover all corner cases. And a more brute way is required now.

PR #1323, opened 2025-08-22
Update Azure and S3 storage settings for main

gerrod3:
Set storage settings in template_config.yml based on pulpcore requirement (>=3.75.0,<3.100).

Storage formats: Azure (new STORAGES), S3 (new STORAGES)

This addresses the recent change where plugin-template stopped automatically setting test scenario storage settings and now requires plugins to define them in their template_config.yml.


gerrod3:
Seems like the pulp-cli-deb content test is failing. Looks like it fails at the end where it does a `pulp deb content list` and checks that 5 units are there. Somehow my change has cause extras to be left over from the main tests (or changed the sync behavior to fail to sync all the units).

quba42:
@gerrod3 The pulp-cli-deb content tests are failing on all branches including `main` since we did a new `pulp-cli-deb` release this week. Have not yet had time to figure this out or fix it, but it is not related to your changes.

PR #1330, opened 2025-09-02
Terminate gpg-agent process after priv. key import

hstct:
Whenever a private key is imported via python-gnupg, this leads to a gpg-agent daemon process for the specified gpg home directory. This process will usually run until its socket file in the gpg home directory is removed, which we do once the temporary directory is cleaned up.

Unfortunately, a race condition can happen between the files the gpg-agent manipulates inside the gpg-home directory and the temporary directory cleanup, which can e.g. manifest in an error like: [Errno 39] Directory not empty: 'gpg-home'

There is a way to stop the gpg-agent belonging to a specific gpg home directory by calling gpgconf. By doing this right after we import the private key, we ensure that no gpg-agent process is running for our gpg home directory. All other methods we currently use from python-gnupg do not spawn a gpg-agent at the moment, so there is currently no need to call gpgconf there.

### DIFF
--- a/pulp_deb/app/tasks/synchronizing.py
+++ b/pulp_deb/app/tasks/synchronizing.py
@@ -10,6 +10,7 @@ import hashlib
 
 from asgiref.sync import sync_to_async
 from collections import defaultdict
+from functools import wraps
 from tempfile import NamedTemporaryFile
 from debian import deb822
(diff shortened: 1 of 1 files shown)


hstct:
messed up this PR. closing in favor of this one: https://github.com/pulp/pulp_deb/pull/1331

PR #1335, opened 2025-09-10
Add a postgresql debver type

mdellweg:
Alternative idea: Add a specific (autogenerated) version_sortkey field.


mdellweg:
Lot's of tinkering around here, and I still need to check some stuff. But feedback is welcome.

PR #1349, opened 2025-10-07
Fix an import/export bug when domains are enabled.

ggainey:
Also enabled import/export tests when domains are enabled, and made it possible for pulp-deb import/export to work when django-import-export==4.x is in use.


ggainey:
See https://github.com/pulp/pulp_rpm/pull/4149#discussion_r2377335909 for discussion on the rationale behind the render() changes.

quba42:
Have not yet had a detailed look at the changes, but can you add a change log for this. I feel like it would be helpful to track what releases have and don't have this change later.

PR #1357, opened 2025-10-27
Add support for Debian package signing

### LINKED ISSUE #1300: Add support for deb package signing
We've been looking at a strategy where we set keys on different repos. The difficulty with implementing this is requiring our users to sign their deb packages with different keys. If a user wants to publish their package to 10 repos then they would potentially have to juggle 10 keys. Ideally, we'd like pulp_deb to sign deb packages for the user.

We'd like [a solution similar to the one in RPM](https://pulpproject.org/pulp_rpm/docs/user/guides/sign-packages/) which would allow users to sign packages for repos when uploading them.

adrianabedon:
This change aims to add support for signing Debian packages upon direct upload to a repository. Matches current package signing support in pulp_rpm and also adds the ability to override the signing key fingerprint per release.

TODO

- [ ] Changelog
- [ ] Tests
- [ ] Docs

closes #1300


quba42:
Sounds interesting! We have seen the tech preview in pulp_rpm. I don't know the details of how exactly package signing in the APT ecosystem works. If you have some docs you can point us at, that would help us once we get around to start reviewing.

daviddavis:
I started [a new PR here](https://github.com/pulp/pulp_deb/pull/1369) since I'll be taking up this work. We can close this out.

PR #1369, opened 2025-12-03
Add support for package signing

### LINKED ISSUE #1300: Add support for deb package signing

daviddavis:
closes #1300


daviddavis:
Here are the Debian docs on signing packages:

https://www.debian.org/doc/manuals/securing-debian-manual/deb-pack-sign.en.html#check-pkg-sign

Here's our own docs about how to manually verify package signatures:

https://github.com/microsoft/linux-package-repositories?tab=readme-ov-file#verify-the-signature-of-an-individual-deb

This doc has a bit more info including how to configure dpkg to verify signatures:

https://blog.packagecloud.io/how-to-gpg-sign-and-verify-deb-packages-and-apt-repositories/

Here's an example package that is signed:

https://packages.microsoft.com/repos/code/pool/main/c/code/code_1.113.0-1774364744_amd64.deb

Note the signature (_gpgorigin) in the deb package:

```
$ wget https://packages.microsoft.com/repos/code/pool/main/c/code/code_1.113.0-1774364744_amd64.deb

$ ar t code_1.113.0-1774364744_amd64.deb
debian-binary
control.tar.xz
data.tar.xz
_gpgorigin
```

quba42:
I am afraid there are now merge conflicts due to the [new linting](https://discourse.pulpproject.org/t/preview-toolchain-upgrade-black-flake8-ruff/2174). It is mostly rearranging of import statements. I hope they are simple to resolve.

daviddavis:
No worries. I am at least [partly to blame](https://github.com/pulp/plugin_template/issues/1027) I think.

daviddavis:
@quba42 I updated this PR and it should be ready for review again.

pyproject.toml
```diff
@@ -28,7 +28,7 @@ requires-python = ">=3.11"
 dependencies = [
   # All things django and asyncio are deliberately left to pulpcore
   # Example transitive requirements: asgiref, asyncio, aiohttp
-  "pulpcore>=3.85.0,<3.115",
+  "pulpcore>=3.107.0,<3.115",
```
quba42, line 31: Just a reminder for myself: Katello has only just packaged for pulpcore 3.105. Once we merge this, it will take some time until Katello can start using this change, and thus the next Y-release.

quba42:
I will just leave this here as a comment: https://gist.github.com/daviddavis/5c6288813044fa0686e2aa936f41b653

docs/user/guides/signing_service.md
```diff
@@ -84,3 +86,76 @@ It assumes that both public and secret key for `GPG_KEY_ID="Pulp QE"` is present
...
+		- The script must use `PULP_SIGNING_KEY_FINGERPRINT` to select the signing key. The
+			`PULP_SIGNING_FINGERPRINT_TYPE` environment variable indicates the fingerprint type
+			(e.g. `v4`, `keyid`).
+		- The script should return JSON describing the signed file:
+			```json
+			{"deb_package": "/absolute/path/to/signed.deb"}
+			```
+2. Register the script with `pulpcore-manager add-signing-service`.
+		- Use `--class "deb:AptPackageSigningService"`.
+		- The public key fingerprint passed here is only used to validate the script registration.
```
quba42: Not sure if this is just an issue with my local environment, but in my local build these nested lists won't render as nested lists. Looking at a similar nested list in the pulp_rpm docs, they seem to use four spaces of indentation here instead of three tabs.

Edit: verified that using four spaces for indentation fixes this in my local docs build.

docs/user/guides/signing_service.md
```diff
@@ -84,3 +86,76 @@ It assumes that both public and secret key for `GPG_KEY_ID="Pulp QE"` is present
...
+FINGERPRINT="${PULP_SIGNING_KEY_FINGERPRINT:?PULP_SIGNING_KEY_FINGERPRINT is required}"
+# PULP_SIGNING_FINGERPRINT_TYPE contains the fingerprint type prefix (e.g. "v4", "keyid")
+FINGERPRINT_TYPE="${PULP_SIGNING_FINGERPRINT_TYPE:-v4}"
+WORKDIR="${PULP_TEMP_WORKING_DIR:-$(mktemp -d)}"
+SIGNED_PATH="${WORKDIR}/$(basename "${PACKAGE_PATH}")"
+
+cp "${PACKAGE_PATH}" "${SIGNED_PATH}"
+debsigs --sign=origin --default-key "${FINGERPRINT}" "${SIGNED_PATH}"
+
+echo {"deb_package": "${SIGNED_PATH}"}
```
quba42:
```suggestion
echo "{\"deb_package\": \"${SIGNED_PATH}\"}"
```
Without this I got the following error during singing service creation:
```
RuntimeError: The signing service script did not return valid JSON!
```

daviddavis:
@quba42 any other feedback?

quba42 (approved):
This is a big change, and I could probably spend another three months reviewing line by line and testing more use cases.

However, I feel I have done a fair amount of review and testing, and I want to create certainty around migration naming and order, and unblock everyone waiting on same.

As such, I propose merging as is. In many ways this will make it easier, not harder to create small follow up PRs without everyone needing to re-evaluate changes to this huge feature branch.

As such, I have only one last question for @daviddavis:

Do I understand you correctly, that there is a desire to keep the migrations on this PR as they currently are?:
```
0035_package_signing.py
0036_add_deb_package_signing_result.py
0037_DATA_fix_signing_fingerprint.py
0038_signing_fingerprint_prefix.py
0039_signing_keys.py
```
That is five separate migrations numbered 0035 to 0039? Or should we squish those together before merging?

Somewhat related: Are there any commits we want to squish together prior to merge?

I am open to leaving things separately as they are, to keep some of the internal history of how this change was created in steps prior to merging.

I am going to keep this PR open with this approving review, until I have gotten an answer vis-a-vis squashing (commits and/or migrations). This will also give @hstct a last chance to veto my approach of merging as is. 😉

quba42:
I am also considering if we should perform one more Y-release BEFORE we merge this, so that we have a latest greatest Y-version before the pulpcore floor is raised to `>=3.107`

daviddavis:
> Do I understand you correctly, that there is a desire to keep the migrations on this PR as they currently are?:

That's correct. We've already run these migrations in our production environment. Doing some research, it looks like I can maybe squash these migrations since we've already run them. I'm happy to take a stab at that.

> Somewhat related: Are there any commits we want to squish together prior to merge?

I kept separate commits that match the ones from pulp_rpm. I thought it would also be easier to review. That said, I am happy to squash them. Maybe one big commit? 

> I am also considering if we should perform one more Y-release BEFORE we merge this, so that we have a latest greatest Y-version before the pulpcore floor is raised to >=3.107

This makes sense to me.

daviddavis:
I took a look into squashing the db migrations. We'd have to create a new squashed migration with a `replaces` key, release it alongside the existing migrations, then cleanup the new migration (removing the `replaces` key) and old migrations. So sounds challenging but not impossible.

It looks like pulpcore and some of the other plugins (pulp_rpm, pulp_container) have squashed their migrations so maybe it's worth just squashing all the pulp_deb migrations at some point?

quba42:
> It looks like pulpcore and some of the other plugins (pulp_rpm, pulp_container) have squashed their migrations so maybe it's worth just squashing all the pulp_deb migrations at some point?

We already have a squash migration up to and including migration 0031: https://github.com/pulp/pulp_deb/blob/main/pulp_deb/app/migrations/0001_initial_squashed_0031_add_domains.py

We just have not yet dropped the individual migrations that this replaces (which we will do at some point).

Given everything you answered, my sense is that it is better to keep the commits and the migrations and merge everything as is. I will have one more internal discussion about it today, and assuming there is no dissent, I will then move ahead with one pre-merge Y-release, and then merge.

quba42:
@daviddavis Thanks for the feature, and thanks for your patience!

PR #1424, opened 2026-03-26
Allow publication of empty repositories

hstct:


pulp_deb/app/tasks/publishing.py
```diff
@@ -198,55 +199,23 @@ def publish(
...
-                    release_components_filtered = release_components.filter(
-                        distribution=distribution
-                    )
-                    components = list(
-                        release_components_filtered.distinct("component").values_list(
-                            "component", flat=True
-                        )
+                        codename=codename,
+                        suite=codename,
+                        origin="Pulp 3",
```
quba42, line 215: Sitting the origin here makes sense from a pulp_deb point of view (consistency with what we did for simple publishing).

However, we did not set the origin in the current Katello implementation, and I think a change in origin value causes hosts to complain and ask for user confirmation, so this would be a disruptive change for Katello users. I am leaning towards making our life easier by simply not setting an origin for now.

quba42: Never mind: Our current Katello implementation also results in `Origin: Pulp 3` in the Release file. I went and checked!

quba42 (approved):
Things I spent some time considering;

- Naming of the new settings.
- Can the new indentation level in the publish function be avoided?
- Can something be done against the spaghetti style taking over publish?

My conclusion: I could not come up with any improvements that are not out of scope for this change.

I did find and fixed one more edge case: If both simple and structured is enabled, we should just rely on the simple metadata (which is always published) and not also publish fallback structured empty metadata. This also prevents failure/collision for the case `STRUCTURED_EMPTY_REPO_DISTRIBUTION = "default"`. I then wen't and made that the new default so our new "default empty structured metadata" is more like our "default simple mode metadata" for consistency. It is a setting so users who don't like this can change it.

Issue #1430, opened 2026-04-01
Hoisting Distributed Publication to pulpcore


pedro-psb:
Hey! I've been working on hoisting the Distributed Publication feature to pulpcore (https://github.com/pulp/pulpcore/issues/7514). This was motivated by [RPM requiring](https://github.com/pulp/pulp_rpm/issues/4236) a solution to the same problem.

First of all I wanna hear some feedback or concerns about the initiative.
Then discuss superseeding plan, if that's ok.

## Superseeding plan

I have an implementation in place with basically the same approach as the one in place here, except it handles pass-through cases (e.g, pulp_file). In pass through publications, no published artifact is involved and ContentArtifacts are matched directly.

The feature adds a core model with the same name, but with some care taken to avoid name collisions. As it's internal, we don't need to worry with API collisions. Because the content handler calls pulp_deb's content_handler customization, deb will keep using it's own implementation if no action is taken.

Given that, here are some options I could think of for superseeding:

Let $v_0$ be the pulpcore version where the feature is released.

1. When deb bumps the bump pulpcore lowerbound to $>=v_0$, deb's implementation gets removed.
    - Until then, deb will use it's own implementation
    - The system will have doubled DistributedPublication objects. Pulpcore one will be ignored
    - Very low effort
2. Right now, add conditional to the code to use core's implementation if available. Then finalize with (2)
    - Deb uses core implementation right away, when possible
    - We might skip deb's DistributedPublication creation if core's implementation is detected (no duplicates).
    - Medium effort

pedro-psb:
Thinking in approach (1), there is one further action that may be taken: Add a data migration from deb's model to core's model to keep existing 'caches' working.

PR #1439, opened 2026-04-20
Create test-reminder for distributed publication phase out work

pedro-psb:
Distributed publication deb's feature was [merged into pulpcore](https://github.com/pulp/pulpcore/pull/7546) (should be released in 3.109). The phase our work is described here: https://github.com/pulp/pulp_deb/issues/1430

This adds a dummy tests that fails when upgrading pulpcore lowerbounds to >=3.115, which is when it's safe to remove the deb's implementation.

Feel free to close if you think that's this is a very low quality workaround for my bad memory :sweat_smile: (because it is)

### DIFF
--- a/pulp_deb/tests/unit/test_phase_out_reminders.py
+++ b/pulp_deb/tests/unit/test_phase_out_reminders.py
@@ -1,0 +1,38 @@
+"""Scheduled reminders disguised as tests — these exist to nag, not to verify behavior."""
+
+from importlib.metadata import requires as pkg_requires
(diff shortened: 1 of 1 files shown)


quba42 (approved):
Successfully tested this will indeed fail when the time comes by locally modifying it to `assert lower_bound < Version("3.84")` instead of `assert lower_bound < Version("3.115")` and running the test.

I am fine with merging this as a reminder.

pedro-psb:
@quba42 thanks!

PR #1440, opened 2026-04-21
Add more Pulp Exceptions.

aKlimau:
Assisted-by: Claude Sonnet 4.6 <noreply@anthropic.com>

### 📜 Checklist

- [ ] Commits are cleanly separated with meaningful messages (simple features and bug fixes should be [squashed](https://pulpproject.org/pulpcore/docs/dev/guides/git/#rebasing-and-squashing) to one commit)
- [ ] A [changelog entry](https://pulpproject.org/pulpcore/docs/dev/guides/git/#changelog-update) or entries has been added for any significant changes
- [ ] Follows the [Pulp policy on AI Usage](https://pulpproject.org/help/more/governance/ai_policy/)
- [ ] (For new features) - User documentation and test coverage has been added

See: [Pull Request Walkthrough](https://pulpproject.org/pulpcore/docs/dev/guides/pull-request-walkthrough/)

### DIFF
--- a/CHANGES/+add-pulp-exceptions.feature
+++ b/CHANGES/+add-pulp-exceptions.feature
@@ -1,0 +1,2 @@
+Convert existing Exceptions to Pulp Exceptions with error codes.
+Bump pulpcore lowerbounds to 3.113.
--- a/pulp_deb/app/exceptions.py
+++ b/pulp_deb/app/exceptions.py
@@ -1,0 +1,167 @@
+from gettext import gettext as _
+
+from pulpcore.plugin.exceptions import PulpException
+
+
+class NoReleaseFile(PulpException):
+    """
+    Raised when no Release file can be found at the expected URL.
+    """
+
+    error_code = "DEB0001"
+
+    def __init__(self, url):
+        super().__init__()
+        self.url = url
+
+    def __str__(self):
+        return f"[{self.error_code}] " + _(
+            "Could not find a Release file at '{url}', try checking the 'url' and "
+            "'distributions' option on your remote"
+        ).format(url=self.url)
+
+
+class NoValidSignatureForKey(PulpException):
+    """
+    Raised when verification of a Release file with the provided GPG key fails.
+    """
+
+    error_code = "DEB0002"
+
+    def __init__(self, url):
+        super().__init__()
+        self.url = url
+
+    def __str__(self):
+        return f"[{self.error_code}] " + _(
+            "Unable to verify any Release files from '{url}' using the GPG key provided."
+        ).format(url=self.url)
+
+
+class NoPackageIndexFile(PulpException):
+    """
+    Raised when no suitable package index file can be found.
+    """
+
+    error_code = "DEB0003"
+
+    def __init__(self, relative_dir):
+        super().__init__()
+        self.relative_dir = relative_dir
+
+    def __str__(self):
+        return f"[{self.error_code}] " + _(
+            "No suitable package index files found in '{relative_dir}'. If you are syncing from "
+            "a partial mirror, you can ignore this error for individual remotes "
+            "(ignore_missing_package_indices='True') or system wide "
+            "(FORCE_IGNORE_MISSING_PACKAGE_INDICES setting)."
+        ).format(relative_dir=self.relative_dir)
+
+
+class MissingReleaseFileField(PulpException):
+    """
+    Raised when an upstream Release file is missing a required field.
+    """
+
+    error_code = "DEB0004"
+
+    def __init__(self, distribution, field):
+        super().__init__()
+        self.distribution = distribution
+        self.field = field
+
+    def __str__(self):
+        return f"[{self.error_code}] " + _(
+            "The release file for distribution '{distribution}' is missing "
+            "the required field '{field}'."
+        ).format(distribution=self.distribution, field=self.field)
+
+
+class UnknownNoSupportForArchitectureAllValue(PulpException):
+    """
+    Exception Signifying that the Release file contains the 'No-Support-for-Architecture-all' field,
+    but with a value other than 'Packages'. We interpret this as an error since this would likely
+    signify some unknown repo format, that pulp_deb is more likely to get wrong than right!
+    """
+
+    error_code = "DEB0005"
+
+    def __init__(self, release_file_path, unknown_value):
+        super().__init__()
+        self.release_file_path = release_file_path
+        self.unknown_value = unknown_value
+
+    def __str__(self):
+        return f"[{self.error_code}] " + _(
+            "The Release file at '{release_file_path}' contains the "
+            "'No-Support-for-Architecture-all' field, with unknown value '{unknown_value}'! "
+            "pulp_deb currently only understands the value 'Packages' for this field, please "
+            "open an issue at https://github.com/pulp/pulp_deb/issues specifying the remote "
+            "you are attempting to sync, so that we can improve pulp_deb!"
+        ).format(
+            release_file_path=self.release_file_path,
+            unknown_value=self.unknown_value,
+        )
+
+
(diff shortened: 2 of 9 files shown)


pulp_deb/app/tasks/synchronizing.py
```diff
@@ -73,92 +83,6 @@
...
-        """
-        message = "The release file for distribution '{}' is missing the required field '{}'."
-        super().__init__(_(message).format(distribution, field), *args, **kwargs)
-
-
-class UnknownNoSupportForArchitectureAllValue(Exception):
-    """
-    Exception Signifying that the Release file contains the 'No-Support-for-Architecture-all' field,
-    but with a value other than 'Packages'. We interpret this as an error since this would likely
-    signify some unknown repo format, that pulp_deb is more likely to get wrong than right!
```
jobselko, line 148: Not a strong opinion, but I would keep the original class descriptions to ensure that no information was lost.

pulp_deb/app/tasks/synchronizing.py
jobselko: `SyncError`

pulp_deb/app/exceptions.py
```diff
@@ -0,0 +1,178 @@
...
+        )
+
+
+class DuplicatePackageChecksumError(PulpException):
+    """
+    Raised when newly added packages have the same name, version, and architecture
+    but different checksums.
+    """
+
+    error_code = "DEB0011"
```
jobselko: Should be `DEB0010`

pulp_deb/app/exceptions.py
```diff
@@ -0,0 +1,178 @@
...
+
+    def __str__(self):
+        return f"[{self.error_code}] " + _(
+            "Cannot create repository version since there are newly added packages with the "
+            "same name, version, and architecture, but a different checksum. If the log level "
+            "is DEBUG, you can find a list of affected packages in the Pulp log. You can often "
+            "work around this issue by restricting syncs to only those distirbution component "
```
jobselko: `distirbution` is a typo from the original code, can we fix it now?

pulp_deb/app/tasks/copy.py
mdellweg, line 75: Wow, these look plain wrong...

pulp_deb/app/tasks/copy.py
```diff
@@ -105,7 +106,7 @@ def process_entry(entry):
         )
 
     if dependency_solving:
```
mdellweg, line 108: Is this something that the api allows in the first place?

I would not change this Exception at all.
If the user can trigger an Unimplemented codepath, that's a server error. And should be prevented/fixed/implemented.

aKlimau: but does it hurt to let the user know what's caused the error?

> Is this something that the api allows in the first place?
> 
> I would not change this Exception at all. If the user can trigger an Unimplemented codepath, that's a server error. And should be prevented/fixed/implemented.

mdellweg: Is there any way the user can trigger this?

quba42: I think the API parameter already exists, and then you will quickly learn not to use it, because it is not implemented and can never work...

aKlimau: yes, if request for copying for some reason has "dependency_solving" set to true: 
```
    dependency_solving = serializers.BooleanField(
        help_text=_(
            "Also copy dependencies of any packages being copied. NOT YET"
            'IMPLEMENTED! You must keep this at "False"!'
        ),
        default=False,
    )
```

mdellweg: I'm sad now. Anyway, can we handle NotImplementedError generically in core? It looks like there would be a pattern.

quba42:
I will try to review this next/soon. It fixes our deprecation warnings, so thanks for that! Just need to find some time to do it...

CHANGES/+add-pulp-exceptions.feature
```diff
@@ -0,0 +1 @@
+Add more Pulp Exceptions.
```
quba42: I think it should be something like:
```suggestion
Convert existing Exceptions to Pulp Exceptions with error codes.
```
Or have I got this all wrong?

pyproject.toml
```diff
@@ -28,7 +28,7 @@ requires-python = ">=3.11"
 dependencies = [
   # All things django and asyncio are deliberately left to pulpcore
   # Example transitive requirements: asgiref, asyncio, aiohttp
-  "pulpcore>=3.85.0,<3.115",
+  "pulpcore>=3.105.0,<3.115",
```
quba42: This will disappear with rebase to latest `main` branch. I need to remember it is part of this change, so I don't get any ideas about trying to backport this. 😅

CHANGES/+add-pulp-exceptions.feature
aKlimau: Yes, your description is more accurate thank you

pyproject.toml
aKlimau: In my latest changes to this pr I bumped pulpcore lowerbounds to 3.113 to use FeatureNotImplementedError from pulpcore. Is that ok or too big of a jump for you?

quba42: My first thought was: Feel free to bump it to what ever version is needed. However, my second thought is:

Do we want to backport all or some of these changes to already released release branches? If yes, then bumping the pulpcore version a bunch suddenly becomes a factor. Right now we have:

- pulp_deb main requires `pulpcore>=3.107.0`
- pulp_deb 3.9 requires `pulpcore>=3.85.0`
- pulp_deb 3.8 requires `pulpcore>=3.85.0`

Looking at that, we probably just don't want to backport these changes to 3.8 or 3.9. So the question becomes: Is 3.113 as opposed to 3.107 for main branch an issue? My intuition is: not really.

=> I am back to "feel free to require 3.113".

pulp_deb/app/tasks/copy.py
quba42: They do look wrong. Don't look like they were copied from pulp_rpm either. Perhaps from some early development state and then we forgot to read them during review? 😰 I may dig through the history at some point to try and understand what happened here. Probably my fault. Certainly out of scope for this PR...

pyproject.toml
jobselko: What would be the reason for backporting features? It does not make sense in most cases :)

pulp_deb/app/tasks/synchronizing.py
jobselko: @aKlimau I meant everywhere in this PR, not just in this one case. But it is up to you.

pyproject.toml
quba42: > What would be the reason for backporting features? It does not make sense in most cases :)

In the case of this refactor of all our existing exceptions the reason to backport would be to avoid merge conflicts with bugfixes we need to backport. This might be less work overall if (a) it wasn't for the pulpcore dependency issues and (b) such merge conflicts are likely. But given (a) is not the case, and (b) seems unlikely, I don't think it makes sense here.

pulp_deb/app/tasks/synchronizing.py
aKlimau: I think in other cases descriptions stayed more or less the same (at least in terms of meaning) so I think new descriptions are a bit better for the consistency across plugins (text style and format)

quba42 (approved):
Everything looks good to me. All discussions resolved. I looked over it one more time and deliberately triggered some (not all) of the errors in oci-env.

Most importantly, this fixes our deprecation warnings. @aKlimau thanks for fixing this for us and thanks for always keeping it up to date so it can keep moving along!

PR #1450, opened 2026-04-28
Fix ruff linting formatter issue

hstct:


pulp_deb/app/serializers/repository_serializers.py
```diff
@@ -226,7 +226,7 @@ def check_cross_domain_config(cfg):
                 err.append(error.message)
             if err:
                 raise serializers.ValidationError(
-                    _("Provided copy criteria is invalid:'{}'".format(err))
+                    _("Provided copy criteria is invalid:'%s'") % err
```
jobselko: This should fix it: `_("Provided copy criteria is invalid:'{}'").format(err)`

hstct: Ah yea thanks, that would be a good solution as well. Both seem to work

hstct: Decided to use your suggestion because it fits with the other log messages.

Thank you again! :relieved:

mdellweg:
To be fair: Ruff pointed at an actual bug here. Translating strings after substituting parts just does not work.

hstct:
Yea, I noticed that so maybe calling it formatter issue is a bit wrong

jobselko (approved):

mdellweg:
Yes technically it resides in the "lint" side of ruff, triggered by `ruff check` not `ruff format`.

jobselko:
@hstct I made the changes directly in all supported branches to make the fix easier.

PR #1457, opened 2026-05-14
Filter serializer validated_data to model fields

daviddavis:
The serializer's validated_data may contain fields from pulpcore's ContentSerializer (e.g. 'overwrite') that don't exist on the plugin model. Filter validated_data against the model's actual fields before passing it to the constructor.

### DIFF
--- a/CHANGES/+filter_validated_data.bugfix
+++ b/CHANGES/+filter_validated_data.bugfix
@@ -1,0 +1,2 @@
+Filter serializer validated data to only include model fields, preventing errors from extra fields
+(e.g. ``overwrite``) added to pulpcore's ``ContentSerializer``.
--- a/pulp_deb/app/serializers/content_serializers.py
+++ b/pulp_deb/app/serializers/content_serializers.py
@@ -627,7 +627,13 @@ class BasePackageMixin(Serializer):
 
         from822_serializer = self.Meta.from822_serializer.from822(data=package_paragraph)
         from822_serializer.is_valid(raise_exception=True)
-        package_data = from822_serializer.validated_data
+        # Filter to model fields: validated_data may carry extra serializer-only fields from
+        # pulpcore's ContentSerializer (e.g. ``overwrite``) that would otherwise be forwarded
+        # to ``self.Meta.model(**package_data)`` and raise a TypeError.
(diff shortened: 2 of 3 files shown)


daviddavis:
I also have [a pulpcore PR to fix the issue](https://github.com/pulp/pulpcore/pull/7717) as well so this change isn't strictly necessary but it still may be worth merging anyway to prevent future problems.

quba42:
@daviddavis Do you recommend backporting this to already released versions? Taking a step back: I don't think I understand the context of where and how this issue was encountered, and by extension how critical it is.

If available, I would also like to request a review from @mdellweg here. My hope is that he has a better idea on whether adding this fix "in addition to any pulpcore side fix" is a good idea, or has any drawbacks?

Are we worried about any potential performance impacts of adding this?

daviddavis:
I don't think it's worth backporting and you may decide it's not even worth merging. It's a defensive change that checks the fields on content serializer before passing them to the `Package` model. 

The problem arose when I added a new field to content serializer in pulpcore (`overwrite`) that defaulted to True and because it had a default, it got passed in `validated_data` to the `Package` constructor, which threw an error because pulp_deb passes all validated_data to `Package()`. There are other fields on content serializer that don't map to `Package` fields (e.g. `repository`) but they don't defaults so they don't show up in `validated_data` so it hasn't been an issue before.

It's possible that another field with a default value could get added to the content serializer in pulpcore, which would again break pulp_deb. This would prevent that.

mdellweg:
I'm wondering if that is "a feature" that it tells us there are values serialized that no code used yet.

daviddavis:
Do you mean that the error is an indication that values aren't supported yet? Then yes, I think that is one way to view it.

By the way, I am planning on adding support for `overwrite` to pulp_deb but am waiting for the package signing feature to be merged so I can add something like [this](https://github.com/pulp/pulp_rpm/pull/4461).

quba42:
After the conversations on this PR, I am leaning towards not merging. Not a strong opinion though.

daviddavis:
@quba42 sounds good. I'll go ahead and close this PR out and we can revisit if/when it's ever needed.

PR #1458, opened 2026-05-20
Add support for overwrite when adding content

daviddavis:
Plumb the new pulpcore `overwrite` parameter to sign_and_create and signed_add_and_remove and override AptRepository.check_content_overwrite to exempt incoming packages already in the version that were produced by the signing workflow (tracked via DebPackageSigningResult).

### DIFF
--- a/CHANGES/+overwrite-signing-noop.feature
+++ b/CHANGES/+overwrite-signing-noop.feature
@@ -1,0 +1,4 @@
+Added support for the new `overwrite` parameter on the APT repository modify
+endpoint. Packages already produced by Pulp's signing workflow (tracked via
+`DebPackageSigningResult`) and present in the repository version are exempted
+from the overwrite check so the operation remains a NOOP.
--- a/pulp_deb/app/models/repository.py
+++ b/pulp_deb/app/models/repository.py
@@ -35,6 +35,7 @@ from pulp_deb.app.models import (
     SourcePackage,
     SourcePackageReleaseComponent,
 )
+from pulp_deb.app.models.signing_service import DebPackageSigningResult
 
 log = logging.getLogger(__name__)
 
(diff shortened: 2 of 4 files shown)


daviddavis:
This PR includes the changes from [the package signing PR](https://github.com/pulp/pulp_deb/pull/1369). Only the last commit is new.

quba42:
If you rebase to latest main branch, the deprecations test should turn green.

daviddavis:
Thanks, looks like it's green now.

quba42 (commented):
I had to look up what `overwrite` is and does. As I understand it, what it does is the following: Where in the past a upload content to repository might have failed on repo uniqueness constraints, with overwrite=True it instead succeeds by kicking out the content it collides with. Did I get this right?

Some more questions for my comprehension:

- It looks like this is making use of something in pulpcore. I don't see the `check_content_overwrite` method getting invoked anywhere in these changes, so I am guessing something in pulpcore will use it. Am I getting this right?
- It looks like this PR is disabling this overwrite mechanism only for cases where package signing is on and packages have already been signed in the caching mechanism? I am not sure I fully understand why this is needed. Are the overwrite mechanism and the "cached signed package" mechanisms clashing somehow for cases where they are both active?

daviddavis:
Mostly right. Before the overwrite feature, [duplicate content was always quietly replaced in Pulp](https://github.com/pulp/pulp_deb/blob/7f76175a106f6faa1bc94db57e023d0aaa5f1148/pulp_deb/app/models/repository.py#L284-L308). This is what happens now when `overwrite=True`. True is the default value for overwrite so it's backwards compatible.

Now with `overwrite`, you can set `overwrite=False`, which will call [`check_content_overwrite`](https://github.com/pulp/pulpcore/blob/4f054560aada77f55b9544b0d08cac7729e9f4f1/pulpcore/app/models/repository.py#L244) in a few different places like [when uploading content into a repo](https://github.com/pulp/pulpcore/blob/4f054560aada77f55b9544b0d08cac7729e9f4f1/pulpcore/app/serializers/content.py#L132) or [modifying a repo's content via `add_and_remove`](https://github.com/pulp/pulpcore/blob/4f054560aada77f55b9544b0d08cac7729e9f4f1/pulpcore/app/tasks/repository.py#L242). When `overwrite=False`, this method will raise a `ContentOverwriteError` if it detects that content would be overwritten/replaced.

This method `check_content_overwrite` can be overriden to add custom logic. Here we are overriding it to prevent `ContentOverwriteError` from being raised if the code detects that a signed version of the package being added to a repo already exists. This is to deal with the case that a user adds a package to a repo and it gets signed (producing a signed version of the package), and then they tried to add the package again with `overwrite=True`.

Without this custom logic, they would hit a `ContentOverwriteError`. Ideally this would just be [a noop](https://github.com/pulp/pulp_deb/blob/95b0c2afea641870c238f599fe6ad3ff11e1baf2/pulp_deb/app/tasks/signing.py#L164-L170) even if `overwrite=True` and that's what this code does--it uses `DebPackageSigningResult` to link the package being added to the repo to the signed version in the repo.

quba42 (approved):
Based on my code review and my cost-benefit-risk intuition, I think we should go ahead and merge this, so we can release it as part of 3.10.0!

PR #1474, opened 2026-07-09
Add a blocklist to exclude specific Package metadata from being publish

### LINKED ISSUE #1497: Filter custom package metadata during sync and publish
APT `Packages` files can contain custom fields that shouldn't always be synced or republished. For example Ubuntu's `Phase-Update-Percentage`.

The idea is to add `excluded_package_metadata_fields` parameters to remotes, repositories and publications as well as an global `EXCLUDED_PACKAGE_METADATA_FIELDS` setting.

Those fields are lists, so you can add multiple custom parameters here. The global and resource level values should be additive.

Filtering should:

- apply only to custom package fields
- match field names case-insensitively
- leave standard fields unchanged
- not affect verbatim publications

hstct:
closes #1497


quba42 (approved):
Everything I tested works as described and as documented as part of these changes.

Any code nitpicks I had were discussed and dropped as not important.

=> As far as I am concerned this is good to go.

PR #1479, opened 2026-07-21
Fix import oversight in test

mdellweg:
Import settings from django.conf instead of pulpcore.app.

### 📜 Checklist

- [ ] Commits are cleanly separated with meaningful messages (simple features and bug fixes should be [squashed](https://pulpproject.org/pulpcore/docs/dev/guides/git/#rebasing-and-squashing) to one commit)
- [ ] A [changelog entry](https://pulpproject.org/pulpcore/docs/dev/guides/git/#changelog-update) or entries has been added for any significant changes
- [ ] Follows the [Pulp policy on AI Usage](https://pulpproject.org/help/more/governance/ai_policy/)
- [ ] (For new features) - User documentation and test coverage has been added

See: [Pull Request Walkthrough](https://pulpproject.org/pulpcore/docs/dev/guides/pull-request-walkthrough/)

### DIFF
--- a/pulp_deb/tests/functional/api/test_sync.py
+++ b/pulp_deb/tests/functional/api/test_sync.py
@@ -1,8 +1,8 @@
 """Tests that sync deb repositories in optimized mode."""
 
 import pytest
+from django.conf import settings
 
-from pulpcore.app import settings  # noqa: TID251
 from pulpcore.tests.functional.utils import PulpTaskError
 
 from pulp_deb.tests.functional.constants import (


quba42:
What is the background? What effect does importing settings from `django.conf` instead of `pulpcore.app` during testing have?

mdellweg:
> What is the background? What effect does importing settings from `django.conf` instead of `pulpcore.app` during testing have?

For the settings used here, probably none. But it is the correct way (the first lines in pulpcore.app.settings said so like forever, and we ran into actual problems where not importing from django.conf collided with checks in management commands). And you can get rid if ignoring a ruff complaint and also prevent future copy & paste incidents.

dralley (approved):

PR #1481, opened 2026-07-21
Update CI image + purge python-gnupg + 3.115 compatibility

dralley:


jobselko (dismissed):

dralley:
python-gnupg is finicky against newer versions of GPG, and it was breaking with the CentOS Stream 10 image, so I'm just going to take this opportunity to replace it with pysequoia, which I needed to do anyway as part of the PQC work.

And to get it working with 3.115 there are some small changes required.

pulp_deb/tests/functional/conftest.py
```diff
@@ -541,72 +539,108 @@ def _deb_copy_content_domain(
     return _deb_copy_content_domain
 
 
+def import_signing_key(key_url, gpg_home):
```
dralley, line 542: This block copied straight from pulpcore.  We could import from pulpcore, but I figured it's easier to just keep them independent, and we're deleting enough code that I don't feel bad about it.

dralley:
@quba42 Matthias suggested I ask for your review also given the changes.  Overall it's a significant simplification since sequoia doesn't require all of the ceremony that `gnupg` does, setting up a one-off `gnupghome` and then needing to clean it up after, etc.

hstct (approved):
This looks good. I especially like the simplification of not using python-gnupg.

The only remaining question is if we should have one last release before we merge this, although I'm not sure if the release pipelines will work without it so this might be moot. @quba42

quba42 (approved):
Thinking about this again, since this does not raise the floor, we don't really gain anything from performing a Y-release before merging this (other than another release branch to maintain). So on second thought, I think we should just merge.

quba42:
There are very substantial merge conflicts for the backport, because 3.9 does not yet have https://github.com/pulp/pulp_deb/pull/1369 which has migrations.

3.10 release it is.

PR #1483, opened 2026-07-22
Make concurrent package signing race-safe

### LINKED ISSUE #1482: Concurrent package signing raises IntegrityError
When signing packages via `signed_add_and_remove`, `_sign_package` checks for an
existing `DebPackageSigningResult` (`filter().first()`) and later inserts a new one
with `DebPackageSigningResult.objects.create()`. These two steps are not atomic, so
when the same original artifact (same `sha256`) is signed with the same fingerprint
concurrently both operations pass the existence check, both sign, and the second
insert violates the `unique_together` constraint on
`(sha256, package_signing_fingerprint)`. This aborts the whole task with the error
below, even though the package was already validly signed. The check-then-create
pattern should be made race-safe (e.g. `get_or_create` / handling the
`IntegrityError` and reusing the existing result) so concurrent signing of the same
package reuses the prior result instead of failing.

​`IntegrityError: duplicate key value violates unique constraint "deb_debpackagesigningres_sha256_package_signing_f_<hash>_uniq". ​`

Related issue https://github.com/pulp/pulp_rpm/issues/4522

daviddavis:
_sign_package used a non-atomic check-then-create for DebPackageSigningResult, so concurrently signing the same original artifact with the same fingerprint could violate the (sha256, package_signing_fingerprint) unique constraint and abort signed_add_and_remove. Wrap the writes in a transaction and reuse the existing result on IntegrityError.

fixes #1482

Assisted-by: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/CHANGES/1482.bugfix
+++ b/CHANGES/1482.bugfix
@@ -1,0 +1,3 @@
+Fixed an `IntegrityError` that could abort `signed_add_and_remove` when the same
+package was signed concurrently by making the `DebPackageSigningResult` creation
+race-safe and reusing the existing result.
--- a/pulp_deb/app/tasks/signing.py
+++ b/pulp_deb/app/tasks/signing.py
@@ -184,11 +184,18 @@ def _sign_package(package, signing_service, signing_fingerprint, package_release
             content=signed_package,
             relative_path=content_artifact.relative_path,
         )
-        DebPackageSigningResult.objects.create(
+        # get_or_create guards against concurrent signing of the same package, which
+        # would otherwise violate the unique constraint and fail the task.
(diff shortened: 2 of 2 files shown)


quba42 (approved):
LGTM. Are there any concerns about merging this before releasing 3.10?

Should we backport this bugfix anywhere? I am not sure if any released versions have had the feature.

daviddavis:
> Are there any concerns about merging this before releasing 3.10?

No concerns.

> Should we backport this bugfix anywhere? I am not sure if any released versions have had the feature.

Yea, I don't think any released versions had this feature.

PR #1488, opened 2026-07-29
Raise the pulpcore upper bound to <3.130

quba42:
Note: We have not yet decided if we want to do this, but we want to see if tests will turn green.


m-bucher (approved):
Can't say much about the implications, but code-wise this looks good 😄

PR #1492, opened 2026-07-29
Allow users to specify distribution and component when modifying content

### LINKED ISSUE #1491: Allow users to supply distribution and component when modifying repo contents
Today, when you upload a package into a repo, you can specify what distribution or component you want it to land in and pulp_deb will create the necessary components for you like PackageReleaseComponent. No such feature exists though for the repository modify endpoint. I imagine could add these parameters though (distribution and component) to the endpoint and have similar functionality.

daviddavis:
Allow `repositories/deb/apt/{pulp_id}/modify/` requests to add/remove packages using optional `distribution` and `component` parameters. The task will create or remove matching release structure content while preserving package-only behavior when both parameters are omitted.

fixes #1491

### DIFF
--- a/CHANGES/1491.feature
+++ b/CHANGES/1491.feature
@@ -1,0 +1,4 @@
+Allow `repositories/deb/apt/{pulp_id}/modify/` requests to assign added packages using the optional
+`distribution` and `component` parameters. Supplying either parameter creates the corresponding
+release component, release architecture, and package-release-component content; omitting both
+preserves the existing package-only behavior.
--- a/pulp_deb/app/serializers/repository_serializers.py
+++ b/pulp_deb/app/serializers/repository_serializers.py
@@ -10,6 +10,7 @@ from pulpcore.plugin.models import SigningService
 from pulpcore.plugin.serializers import (
     PgpKeyFingerprintField,
     RelatedField,
+    RepositoryAddRemoveContentSerializer,
     RepositorySerializer,
     RepositorySyncURLSerializer,
     ValidateFieldsMixin,
@@ -24,6 +25,28 @@ from pulp_deb.app.models import (
     AptRepositoryReleaseServiceOverride,
 )
 from pulp_deb.app.schema import COPY_CONFIG_SCHEMA
+
+
+class AptRepositoryAddRemoveContentSerializer(RepositoryAddRemoveContentSerializer):
+    distribution = serializers.CharField(
+        help_text=_(
+            "Name of the distribution any packages from add_content_units or remove_content_units "
+            "should be added to or removed from. Defaults to DEFAULT_DISTRIBUTION if only a "
+            "component is provided."
+        ),
+        required=False,
+    )
+    component = serializers.CharField(
+        help_text=_(
+            "Name of the component any packages from add_content_units or remove_content_units "
+            "should be added to or removed from. Defaults to DEFAULT_COMPONENT if only a "
+            "distribution is provided.."
+        ),
+        required=False,
+    )
+
+    class Meta(RepositoryAddRemoveContentSerializer.Meta):
+        fields = RepositoryAddRemoveContentSerializer.Meta.fields + ["distribution", "component"]
 
 
 class ServiceOverrideField(serializers.DictField):
--- a/pulp_deb/app/serializers/repository_serializers.py
+++ b/pulp_deb/app/serializers/repository_serializers.py
@@ -10,20 +10,51 @@ from pulpcore.plugin.models import SigningService
 from pulpcore.plugin.serializers import (
     PgpKeyFingerprintField,
     RelatedField,
+    RepositoryAddRemoveContentSerializer,
     RepositorySerializer,
     RepositorySyncURLSerializer,
     ValidateFieldsMixin,
 )
 from pulpcore.plugin.util import get_domain, get_url
 
+from pulp_deb.app.constants import PACKAGE_UPLOAD_DEFAULT_DISTRIBUTION
 from pulp_deb.app.models import (
     AptPackageSigningService,
     AptReleaseSigningService,
     AptRepository,
     AptRepositoryReleasePackageSigningFingerprintOverride,
     AptRepositoryReleaseServiceOverride,
+    Release,
 )
 from pulp_deb.app.schema import COPY_CONFIG_SCHEMA
+
+
+class AptRepositoryAddRemoveContentSerializer(RepositoryAddRemoveContentSerializer):
+    distribution = serializers.CharField(help_text=_("Name of the distribution."), required=False)
+    component = serializers.CharField(help_text=_("Name of the component."), required=False)
+
+    def validate(self, data):
+        data = super().validate(data)
+        if not (data.get("distribution") or data.get("component")):
+            return data
+
+        # A request that only names a component is scoped to the default distribution.
+        distribution = data.get("distribution") or PACKAGE_UPLOAD_DEFAULT_DISTRIBUTION
+        releases = Release.objects.filter(distribution=distribution)
+        repository = self.context.get("repository")
+        repository_version = data.get("base_version") or (
+            repository.latest_version() if repository else None
+        )
+        added = releases.filter(pk__in=data.get("add_content_units", [])).exists()
+        present = repository_version and releases.filter(pk__in=repository_version.content).exists()
+        if not (added or present):
+            raise DRFValidationError(
+                {"distribution": _("This distribution has no Release in the repository.")}
+            )
+        return data
+
+    class Meta(RepositoryAddRemoveContentSerializer.Meta):
+        fields = RepositoryAddRemoveContentSerializer.Meta.fields + ["distribution", "component"]
 
 
 class ServiceOverrideField(serializers.DictField):
--- a/pulp_deb/app/tasks/signing.py
+++ b/pulp_deb/app/tasks/signing.py
@@ -14,19 +14,129 @@ from pulpcore.plugin.models import (
     ContentArtifact,
     CreatedResource,
     PulpTemporaryFile,
+    RepositoryVersion,
     Upload,
     UploadChunk,
 )
 from pulpcore.plugin.tasking import add_and_remove, general_create
 from pulpcore.plugin.util import get_url
 
-from pulp_deb.app.models import AptRepository, Package, PackageReleaseComponent
+from pulp_deb.app.constants import (
+    DEFAULT_COMPONENT,
+    DEFAULT_DISTRIBUTION,
+)
+from pulp_deb.app.models import (
+    AptRepository,
+    Package,
+    PackageReleaseComponent,
+    ReleaseArchitecture,
+    ReleaseComponent,
+    SourcePackage,
+    SourcePackageReleaseComponent,
+)
 from pulp_deb.app.models.signing_service import (
     AptPackageSigningService,
     DebPackageSigningResult,
 )
 
 log = logging.getLogger(__name__)
+
+
+def _prepare_package_removals(repo, remove_content_units, base_version_pk, distribution, component):
+    """Expand the removal list to include the release component relationships of each package.
+
+    Removing a (source) package also requires removing its PackageReleaseComponent /
+    SourcePackageReleaseComponent links. When a distribution/component is given, the removal is
+    scoped to that component: a package is only removed from the repository if the scope held its
+    last relationship, so packages linked elsewhere or not linked at all are kept.
+    """
+    # "*" removes all content, so there is nothing to resolve here.
+    if not remove_content_units or "*" in remove_content_units:
+        return
+
+    repository_version = (
+        RepositoryVersion.objects.get(pk=base_version_pk)
+        if base_version_pk
+        else repo.latest_version()
+    )
+    # A distribution/component narrows the removal to a single release component.
+    scoped = distribution is not None or component is not None
+    if scoped:
+        distribution = distribution or DEFAULT_DISTRIBUTION
+        component = component or DEFAULT_COMPONENT
+
+    for model, relationship_model, relationship_field in (
+        (Package, PackageReleaseComponent, "package"),
+        (SourcePackage, SourcePackageReleaseComponent, "source_package"),
+    ):
+        units = model.objects.filter(pk__in=remove_content_units)
+        relationships = relationship_model.objects.filter(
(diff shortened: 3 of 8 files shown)


quba42:
First of all: I really appreciate this change! I started by testing some workflows using this, against my own assumptions and expectations. Overall the current state makes very reasonable design choices. It appears to work well (though I have not yet tested the content removal workflows :wink:)

That being said, on the details, there are several design choices I would like to have some (open ended) discussion on. To avoid this turning to chaotic, I am going to avoid GitHub's review feature for now, and will instead open a thread for each independent thing I would like to discuss. I may need a bit of time to add each thread below.

CHANGES/1491.feature
```diff
@@ -0,0 +1,4 @@
+Allow `repositories/deb/apt/{pulp_id}/modify/` requests to assign added packages using the optional
+`distribution` and `component` parameters. Supplying either parameter creates the corresponding
+release component, release architecture, and package-release-component content; omitting both
+preserves the existing package-only behavior.
```
quba42, line 4: > omitting both preserves the existing package-only behavior.

As I mentioned on the issue, I would very much like to follow up with a breaking change at some point that makes the "existing package-only behavior" impossible. It never really made a lot of sense to allow adding packages that are then ignored by the default publish. It is not at all intuitive for new users. However, in the interest of making progress, I think it is best we go ahead as stated here, and postpone the whole discussion about a breaking change for a follow up PR.

pulp_deb/app/tasks/signing.py
```diff
@@ -29,6 +41,100 @@
...
+    ReleaseComponent, ReleaseArchitecture and *ReleaseComponent relationships exist and add them
+    to the content being added.
+    """
+    packages = list(Package.objects.filter(pk__in=add_content_units))
+    source_packages = list(SourcePackage.objects.filter(pk__in=add_content_units))
+    if not (packages or source_packages) or (distribution is None and component is None):
+        return
+
+    distribution = distribution or PACKAGE_UPLOAD_DEFAULT_DISTRIBUTION
+    component = component or PACKAGE_UPLOAD_DEFAULT_COMPONENT
```
quba42: This is a bit of a minor technicality, but I do want to discuss it regardless: Defaulting to the "upload defaults", feels a little weird, since modify is not, an "upload" action. I suppose the alternative would be to add two new settings PACKAGE_MODIFY_DEFAULT_DISTRIBUTION, and PACKAGE_MODIFY_DEFAULT_COMPONENT. Which raises the question of: does it make sense to start collecting a separate distribution and component default name for each workflow that needs a default value for distributions or components. If I had simply named the variables PACKAGE_DEFAULT_DISTRIBUTION and PACKAGE_DEFAULT_COMPONENT when I introduced them, it would have been simpler. I may be over thinking this. But it just feels a little weird that using modify results in a default component of 'upload'...

pulp_deb/app/tasks/signing.py
```diff
@@ -29,6 +41,100 @@
...
+    release_component, _ = ReleaseComponent.objects.get_or_create(
+        distribution=distribution, component=component
+    )
+    add_content_units.append(str(release_component.pk))
+    # Each binary package needs its architecture and a link to the release component.
+    for package in packages:
+        architecture, _ = ReleaseArchitecture.objects.get_or_create(
+            distribution=distribution, architecture=package.architecture
+        )
```
quba42: Whatever the wisdom of this existing design, at least for syncs, we never explicitly create a ReleaseArchitecture for `architecture=all`. Publish will automatically append "all" to the list of architectures in accordance with the spec for the publication format we use. Just for consistency I am leaning towards not creating ReleaseArchitecture objects with architecture=all here either. I am not sure if having them could ever be harmful, but my bias is to err on the side of consistency, perhaps even on the side of "sub-optimal but consistent".

pulp_deb/app/serializers/repository_serializers.py
```diff
@@ -10,22 +10,53 @@
...
+        repository = self.context.get("repository")
+        repository_version = data.get("base_version") or (
+            repository.latest_version() if repository else None
+        )
+        added = releases.filter(pk__in=data.get("add_content_units", [])).exists()
+        present = repository_version and releases.filter(pk__in=repository_version.content).exists()
+        if not (added or present):
+            raise DRFValidationError(
+                {"distribution": _("This distribution has no Release in the repository.")}
+            )
```
quba42: I consider this to be the most important design choice of this new feature.
It is the thing users who want to use the new fields are most likely to trip over.

From the point of view of the publish code, we don't need to demand there be a Release object to go with the distribution we are targeting. If there is a ReleaseComponent with distribution="my-distribution", but no Releae with distribution="my-distribution", then the publish code will simply invent a default Release object: https://github.com/pulp/pulp_deb/blob/main/pulp_deb/app/tasks/publishing.py#L219

So, we could drop this requirement entirely. If users provide a distribution and/or component value to the modify endpoint, we would simply create the repo structure that these values imply. If they don't like the default values in the Release file, they could still add a custom Release object later.

However, I wonder if it makes sense to have some barrier against users accidentally messing up their repositories, for example because they have a typo in the distribution they supplied. (I suppose if you do mess it up, then the weird result would at least be contained within its own repo version which you could simply delete again).

With the current design there is a barrier against providing a weird distribution, but there is no barrier against providing a weird component. So another version of this validation might be to demand you either add or already have present a ReleaseComponent that matches both the distribution and the release you are targeting.

So I currently see three options here:

1. Validation based on presence or addition of Release object (current state of the PR)
2. No validation at all (you simply get what you ask for, even if what you asked for contains obvious errors).
3. Validation based on presence or addition of a ReleaseComponent (I am not going to let you add the packages to a repo structure that is not already part of the repo, and that you have not explicitly asked me to add).

Opinions?

quba42:
I am going to leave it at that for now. Note that while I may have produced a lot of text, all of the above are posed as open questions. Disagreeing with any implied changes is perfectly fine. If, after some discussion, we conclude we want to stay with the current design that is a perfectly reasonable outcome.

pulp_deb/app/tasks/signing.py
quba42: I kind of wish I could go back in time, and make it so I introduced these variables as:
```
DEFAULT_DISTRIBUTION = 'pulp'
DEFAULT_COMPONENT = 'default'
```
Instead of:
```
PACKAGE_UPLOAD_DEFAULT_DISTRIBUTION = 'pulp'
PACKAGE_UPLOAD_DEFAULT_COMPONENT = 'upload'
```

CHANGES/1491.feature
daviddavis: 100% agree. This issue has bitten us before. And I agree that a follow up discussion and PR makes sense.

pulp_deb/app/tasks/signing.py
daviddavis: I think `DEFAULT_DISTRIBUTION` and `DEFAULT_COMPONENT` make sense. What if we rename them but also continue supporting them (PACKAGE_UPLOAD_*) for a few releases?

pulp_deb/app/serializers/repository_serializers.py
daviddavis: Typos would be my biggest concern. If someone misspells something, it is not an easy action to undo. In our system, creating a release or component is an explicit step the user must perform so I think it helps to minimize this risk.

That said, we are validating that the release and the component exist before we hand the modify request off to Pulp so we don't have a strong opinion which option we decide here. In fact, I see good arguments for each.

pulp_deb/app/tasks/signing.py
quba42: I could see us needing this in more places in the future, so a plugin wide `DEFAULT_DISTRIBUTION` and `DEFAULT_COMPONENT` for whenever we need a distribution/component but the user did not provide one makes the most sense to me. Let's introduce these new variables as part of this PR, leaving the existing `PACKAGE_UPLOAD_DEFAULT_*` variables untouched for now. We can always drop them in favor of the new variables in a follow up.

pulp_deb/app/tasks/signing.py
quba42: Having thought about it some more I think the existing design is correct. A ReleaseArchitecture content indicates that a release supports a particular architecture. `Architecture: all` is not an architecture, but rather a special value with special meaning within APT repo metadata. I am seriously considering a follow up task to make this even more explicit by putting some validation on ReleaseArchitecture that generally prevents creation of a ReleaseArchitecture "all" going forward.

pulp_deb/app/serializers/repository_serializers.py
quba42: We discussed this one internally and we came out leaning towards "no validation", that is you get what you asked for whether that is what you wanted or not.

For number 3. we decided this would again ask users to interact with the plumbing we are trying to insulate them from.

For number 1. (validation based on Release, current state) we decided we should not demand the presence of Release objects given they are not technically required.

Overall my thinking is: If we want to move towards making use of this feature mandatory, then it must be as easy to use as possible. And throwing validation errors demanding certain things should already be in the repo (or must be provided with the modify) makes it harder to use/discourages usage.

I do still worry about typos being a very bad experience here. (Also providing one of distribution/component and not understanding about the implication that this causes the other to default to our default value). The thing that mitigates my concern is that the resulting chaos would be contained within its own repository version. Of course, if users start building further versions on top of the broken one before they realize their error, it can quickly become hard to recover from. I would hope that for most users this would only be an issue when they first experiment with the plugin, and will then quickly learn...

On the whole I still lean towards making the feature easy to use even if this increases the risk of typos resulting in bad outcomes.

daviddavis:
@quba42 I think I've addressed everything. Please let me know what you think.

balasankarc:
cc @stanhu FYI as you worked on https://github.com/pulp/pulp-cli-deb/pull/260

daviddavis:
@balasankarc @stanhu reading https://github.com/pulp/pulp-cli-deb/issues/245, it seems like maybe the field here should be `distributions` instead of `distribution`? So that way you can add a package to multiple distributions when you add it to the repo. I don't think that would be too difficult to do.

pulp_deb/app/serializers/repository_serializers.py
```diff
@@ -26,6 +27,14 @@
 from pulp_deb.app.schema import COPY_CONFIG_SCHEMA
 
 
+class AptRepositoryAddRemoveContentSerializer(RepositoryAddRemoveContentSerializer):
+    distribution = serializers.CharField(help_text=_("Name of the distribution."), required=False)
+    component = serializers.CharField(help_text=_("Name of the component."), required=False)
```
quba42: I think it is worth fleshing out these descriptions a bit more:
```suggestion
    distribution = serializers.CharField(help_text=_("Name of the distribution any packages from add_content_units or remove_content_units should be added to or removed from. Defaults to DEFAULT_DISTRIBUTION if only a component is provided."), required=False)
    component = serializers.CharField(help_text=_("Name of the component any packages from add_content_units or remove_content_units should be added to or removed from. Defaults to DEFAULT_COMPONENT if only a distribution is provided.."), required=False)
```

pulp_deb/app/tasks/signing.py
```diff
@@ -29,6 +42,102 @@
...
+    component = component or DEFAULT_COMPONENT
+    release_component, _ = ReleaseComponent.objects.get_or_create(
+        distribution=distribution, component=component
+    )
+    add_content_units.append(str(release_component.pk))
+    # Each binary package needs its architecture and a link to the release component.
+    for package in packages:
+        architecture, _ = ReleaseArchitecture.objects.get_or_create(
+            distribution=distribution, architecture=package.architecture
+        )
```
quba42:
```suggestion
        if package.architecture != 'all':
            architecture, _ = ReleaseArchitecture.objects.get_or_create(
                distribution=distribution, architecture=package.architecture
            )
```

pulp_deb/app/tasks/signing.py
```diff
@@ -29,6 +42,102 @@
 log = logging.getLogger(__name__)
 
 
+def _prepare_package_removals(repo, remove_content_units, base_version_pk, distribution, component):
+    """Expand the removal list to include the release component relationships of each package.
+
+    Removing a (source) package also requires removing its PackageReleaseComponent /
+    SourcePackageReleaseComponent links. When a distribution/component is given, the removal is
+    scoped to that component: a package is only removed from the repository if the scope held its
+    last relationship, so packages linked elsewhere or not linked at all are kept.
```
quba42, line 51: This describes my preferred behavior, and I tested this is in fact what happens. Also tested that removing a package without providing distribution and/or component, removes it and all PRCs referencing it from the repo as in the past.

So all things removal look good to me. 👍

quba42 (changes requested):
I found two more minor changes I would like to see.

With those final changes I am happy to merge. I have manually tested many workflows I wanted to see (like adding a package already in the repo to another ReleaseComponent, removing packages from just one ReleaseComponent. Removing a package from the repo and all ReleaseComponents (by not providing the new params). Adding multiple packages. Re-running a command that results in a noop because everything is already present, etc.

I also looked over the new automatic tests which provide good coverage for those same workflows I most wanted to test manually. 👍

quba42:
One thing I did not test is what happens when you provide both add_content_units and remove_content units and a component or distribution in a single call. From the way the individual calls work and how the code is structured I am pretty confident it will simply be the conjunction of the two individual calls. I am not sure how important this corner case is. I would expect the vast majority of users to split that request up into two calls, just because the request "add these packages to this ReleaseComponent and also remove these other packages from the same ReleaseComponent at the same time" is conceptually pretty weird. :smile: 

If you end up requesting the same package be added and also removed in the same call, I don't particularly care what happens since such a clearly contradictory request arguably _should result in undefined behavior_.

daviddavis:
I agree that most users will probably add/remove in a single call. It should handle requests to both add and remove content in a single call though by both adding and removing the packages that the user requests to add and remove. 

The one edge case you alluded to where there is an overlap of content is not obvious for users though without looking at the code (I believe that the answer is that the content gets added). I agree though that we shouldn't worry about it since it's obviously unusual and contradictory.

daviddavis:
I went ahead and added a new functional test (test_add_and_remove_packages_in_same_request) to test adding/removing packages a single request.

quba42 (approved):

PR #1496, opened 2026-07-30
Defer server side imports in functional tests

daviddavis:
The package signing tests imported AptPackageSigningService at module scope, which requires django.setup() to have run. That only happened as a side effect of the pytest-django plugin, which is declared in unittest_requirements.txt, so collection failed for anyone installing only functest_requirements.txt.

Route the signature check through a session fixture depending on `django.setup()`, which is how pulpcore initializes Django for functional tests. Adding pytest-django to functest_requirements.txt would also work, but it drags database fixtures into an HTTP only suite and makes Django mandatory for every functional test run.

Assisted-by: GitHub Copilot

### DIFF
--- a/pulp_deb/tests/functional/api/test_package_signing.py
+++ b/pulp_deb/tests/functional/api/test_package_signing.py
@@ -7,7 +7,6 @@ import pytest
 
 from pulpcore.client.pulp_deb.exceptions import ApiException
 
-from pulp_deb.app.models import AptPackageSigningService
 from pulp_deb.tests.functional.utils import get_local_package_absolute_path
 
 
@@ -62,6 +61,7 @@ def add_package_to_repo(
 def test_sign_package_on_upload(
     tmp_path,
     download_content_unit,
+    deb_check_signature,
     deb_signing_key_primary,
     deb_signing_key_secondary,
     deb_package_signing_service,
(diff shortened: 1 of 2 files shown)


quba42:
Note: Adding pulp_rpm to my oci-env had the side effect of fixing my local test runs.

daviddavis:
In pulp_rpm there's a pytest entry point [here](https://github.com/pulp/pulp_rpm/blob/d0c4919fe1f672eb12068039851d7d7729b3fe10/pyproject.toml#L52) that [loads the constants file](https://github.com/pulp/pulp_rpm/blob/d0c4919fe1f672eb12068039851d7d7729b3fe10/pulp_rpm/pytest_plugin.py#L22) and in that constants file, [they are calling django.setup()](https://github.com/pulp/pulp_rpm/blob/d0c4919fe1f672eb12068039851d7d7729b3fe10/pulp_rpm/tests/functional/constants.py#L9).

daviddavis:
A couple alternatives to this change: call `django.setup()` at the top of the functional conftest (which is what pulp_rpm does), giving us a one-line fix that keeps the natural module-scope import, at the cost of making Django setup mandatory for every functional test run rather than just the signing tests. Or I could drop the server-side import entirely and reimplement the signature check in the functional test utils using python-gnupg, which is already in functest_requirements — that's roughly 30 lines of duplicated logic, but it removes the Django dependency from these tests completely and avoids the circularity of verifying the server's signing with the server's own verifier.

PR #1499, opened 2026-08-13
Fix debian structured copy with large content queries

hstct:
for fixing issues with 65k parameter limit on copy tasks


hstct:
closed in favor of #1506

PR #1522, opened 2026-09-09
Handle architecture all in repository modify

daviddavis:
## Summary
- Repository modify previously accessed an unassigned/stale ReleaseArchitecture for `Architecture: all`.
- It now only adds ReleaseArchitecture content for concrete architectures while retaining PackageReleaseComponent.

## Testing
- `oci-env test -p pulp_deb functional -k test_repository_modify -vv`
- Result: 12 passed, 1 skipped

## Context
- Follows #1492 and complements #1513.

Assisted By: GitHub Copilot

### DIFF
--- a/CHANGES/+repository-modify-architecture-all.bugfix
+++ b/CHANGES/+repository-modify-architecture-all.bugfix
@@ -1,0 +1,1 @@
+Fixed modifying a repository with an architecture-independent package when distribution or component fields were provided.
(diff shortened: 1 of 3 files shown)


quba42 (approved):
Code review all looks good. (Only add the architecture to the add_content_units list if we actually entered the if block that picks out an architecture, duh!).

The test looks right code wise and I checked it fails without the fix.

PR #1525, opened 2026-09-10
Preserve apt automatic upgrade Release fields

hstct:
Preserving `NotAutomatic` and `ButAutomaticUpgrades` fields. These fields are used by official Debian backport repositories to indicate to APT that backports should use a default priority.


pulp_deb/app/serializers/content_serializers.py
quba42, line 812: We could add validation for the following rule:

> Specifying "yes" for ButAutomaticUpgrades without specifying "yes" for NotAutomatic is invalid. 

Source: https://wiki.debian.org/DebianRepository/Format#NotAutomatic_and_ButAutomaticUpgrades

Example for serializer based model verification: https://github.com/pulp/pulp_deb/blob/main/pulp_deb/app/serializers/content_serializers.py#L916

Downside is, if there are repos in the wild that violate the rule, then we can no longer sync them at all. A more defensive approach might be to log a warning and drop the value for ButAutomaticUpgrades if ButAutomaticUpgrades is "yes", and NotAutomatic is not "yes". That way the sync would still succeed, but we do not save explicitly forbidden value combinations to the DB. Thoughts?

pulp_deb/app/models/content/metadata.py
```diff
@@ -29,6 +29,8 @@ class Release(Content):
     origin = models.TextField(default=NULL_VALUE)
     label = models.TextField(default=NULL_VALUE)
     description = models.TextField(default=NULL_VALUE)
+    not_automatic = models.BooleanField(default=False)
+    but_automatic_upgrades = models.BooleanField(default=False)
```
quba42, line 33: With this, we have no way to represent the case "The release file does not have the fields `NotAutomatic` or `ButAutomaticUpgrades`". Instead in those cases, we would include the fields but set them to False by default.

I am not sure if "Does not have the field set" vs. "Has the field set, but the value is False" ever makes a difference, but normally I would prefer to just do what the original upstream repo did. However, this means we would need the default to be null. However, if https://github.com/pulp/pulp_deb/blob/main/pulp_deb/app/constants.py#L23 is true, then we can't use the combination of BooleanField, have the field be part of unique_together and also allow null. We could use `models.TextField(default=NULL_VALUE)`, but then we can't use the `YesNoField` for the serializer. We could duplicate the YesNoField logic for a TextField, but that is not great.

=> I am not sure what the best compromise is here. Let's discuss it in a call.

CHANGES/+preserve-release-metadata-fields.bugfix
```diff
@@ -0,0 +1 @@
+Preserved the `NotAutomatic` and `ButAutomaticUpgrades` fields from Debian Release metadata during sync and structured publish
```
quba42, line 1: Not that important, but I would consider this a feature:
> Added support to sync and publish the `NotAutomatic` and `ButAutomaticUpgrades` Release file fields.

quba42 (changes requested):
See the comments below. I am not sure if we should implement the suggestions. Let's discuss it and then upgrade the threads accordingly.

CHANGES/+preserve-release-metadata-fields.feature
```diff
@@ -0,0 +1 @@
+Added support to sync and publish the `NotAutomatic` and eButAutomaticUpgrades`` Release file fields.
```
quba42: Nitpick:
```suggestion
Added support to sync and publish the `NotAutomatic` and `ButAutomaticUpgrades` Release file fields.
```

quba42 (dismissed):
With the exception of the typo in the changelog message, this all looks good to me.

Tested against Debian bookworm-backports and experimental, as well as a repo that does not sport either flag. => Everything works as expected.

quba42 (approved):

quba42:
Note: Here are some real world test cases illustrating this feature:
```
alias pulp='pulp -p oci-env'
pulp status

NAME='quickstart-nginx-bookworm-amd64'
REMOTE_OPTIONS=(
  --url=http://nginx.org/packages/debian/
  --distribution=bookworm
  --component=nginx
  --architecture=amd64
  --policy=on_demand
)

NAME='debian-experimental-contrib-amd64'
REMOTE_OPTIONS=(
  --url=http://ftp.de.debian.org/debian/
  --distribution=experimental
  --component=contrib
  --architecture=amd64
  --policy=on_demand
)


NAME='debian-bookworm-backports-contrib-amd64'
REMOTE_OPTIONS=(
  --url=http://ftp.de.debian.org/debian/
  --distribution=bookworm-backports
  --component=contrib
  --architecture=amd64
  --policy=on_demand
)

# For each of the above cases use the following workflow, and then check the published Release file.
pulp deb remote create --name=${NAME} ${REMOTE_OPTIONS[@]}
pulp deb repository create --name=${NAME} --remote=${NAME}
pulp deb repository sync --name=${NAME}
pulp deb publication create --repository=${NAME}
pulp deb distribution create --name=${NAME} --base-path=${NAME} --repository=${NAME}
```

PR #1530, opened 2026-09-15
Enable pulpcore's reclaim disk space feature

quba42:
Entirely untested.

Regarding test coverage:

pulp_rpm does not have any tests for this, relying entirely on pulpcore to do the right thing.

If we do decide to add a test, I would like to see the following:
- Sync repo A with (at least) packages a and b.
- Sync repo B with (at least) packages b and c. (b must be checksum identical for both repos)
- Call reclaim space for repo A.
- Assert that the artifact for package a is gone.
- Assert that the artifacts for package b and c are not gone.
- (optional) Download package a and check that the artifact is back (I believe it should have been turned into a remote/on_demand artifact).


quba42:
Tested this manually. Reclaim of both on_demand and immediate synced repos works as expected. Reclaimed content can still be consumed on_demand as expected.

We discussed we are happy with the tests in pulpcore just like pulp_rpm is.

hstct (approved):
The pipeline failures for the cloud providers are likely unrelated. So change itself looks good

